PR #28, opened 2017-06-21
dash.ly -> dash (right now pip users get both libraries)

jackwluo:


chriddyp:
ah thanks @jackwluo ! Can you bump the version too? In the `package.json` and in the `version.py` file. plotly/dash-html-components might need this update too, perhaps plotly/dash-renderer too

dash_core_components/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.5.0'
+__version__ = '0.6.1'
```
chriddyp: it should just be `0.5.1` actually

jackwluo:
Fixed

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dash-core-components",
-  "version": "0.5.0",
+  "version": "0.5.2",
```
chriddyp: @jackwluo  - this needs to match the python version. `0.5.1`

PR #33, opened 2017-07-03
Include dimensions of selected region in `selectedData`

chriddyp:


chriddyp:
lasso

unstephenk:
This is a great feature!! How do we use it?

PR #34, opened 2017-07-10
Added mouse event interaction for pie chart and sankey diagram

alishobeiri:


src/components/Graph.react.js
```diff
@@ -1,5 +1,6 @@
 import React, {Component, PropTypes} from 'react';
 import {contains, filter, has, type} from 'ramda';
+import {omit} from 'underscore'
```
chriddyp: We use `ramda` instead of underscore. Check out their docs here: http://ramdajs.com

src/components/Graph.react.js
```diff
@@ -18,11 +19,14 @@ const filterEventData = (gd, eventData, event) => {
         const data = gd.data;
         for(let i=0; i < eventData.points.length; i++) {
             const fullPoint = eventData.points[i];
-            const pointData = filter(function(o) {
+            var pointData = filter(function(o) {
```
chriddyp: we aren't mutating `pointData` anywhere after this, so it can remain `const` right?

src/components/Graph.react.js
```diff
@@ -18,11 +19,14 @@ const filterEventData = (gd, eventData, event) => {
...
                 return !contains(type(o), ['Object', 'Array'])
             }, fullPoint);
-
-            if (has('customdata', data[pointData.curveNumber]) &&
+            if(data[0]['type'].toLowerCase() === 'pie'){
+              return omit(pointData, ['originalEvent', 'cxFinal', 'cyFinal', 'cy', 'cx', 'vTotal']);
+            } else if (data[0]['type'].toLowerCase() === 'sankey'){
+              return pointData
+            }else if (has('customdata', data[pointData.curveNumber]) &&
```
chriddyp: what about if users set the `customdata` key in `sankey` or `pie`?

src/components/Graph.react.js
```diff
@@ -18,11 +19,14 @@ const filterEventData = (gd, eventData, event) => {
...
         for(let i=0; i < eventData.points.length; i++) {
             const fullPoint = eventData.points[i];
-            const pointData = filter(function(o) {
+            var pointData = filter(function(o) {
                 return !contains(type(o), ['Object', 'Array'])
             }, fullPoint);
-
-            if (has('customdata', data[pointData.curveNumber]) &&
+            if(data[0]['type'].toLowerCase() === 'pie'){
+              return omit(pointData, ['originalEvent', 'cxFinal', 'cyFinal', 'cy', 'cx', 'vTotal']);
```
chriddyp: instead of "blacklisting" (excluding the things we don't want), can we "whitelist" (only include the keys that we want)? That way, it'll be a little bit more explicit

src/components/Graph.react.js
```diff
@@ -18,11 +19,14 @@ const filterEventData = (gd, eventData, event) => {
...
+            } else if (data[0]['type'].toLowerCase() === 'sankey'){
+              return pointData
```
chriddyp: So instead of returning an array of points, we're just returning the single object. Is that because you can never click or hover on multiple points in a `pie` or a `sankey`? That's true for `clicking` and `hovering`, but might not necessarily be true in the future for `selecting`.

I think that we should return everything in an array for consistency for now, even if that chart type doesn't have a user interaction that supports returning multiple points. 

If we were to change the structure to only display a single object instead of an array for chart types that don't support returning multiple points, then we should add this logic to the rest of those chart types (like `scatter3d` and all of the maps). I don't think that that is something that I want to maintain.

alishobeiri:
@chriddyp Sorry, my last commit I didn't quite understand what was going on with the customdata and how everything was working together. Turns out, for the sankey and pie graphs there was an error being thrown that didn't allow mouse event data to be passed back to dash. I moved around the conditions on the if statement to prevent this. So the hoverdata should work on all types of plots now.

Right now I don't think it's possible to allow for customdata hover events for sankey and pie graphs with more than one trace per plot. Sankey and pie plots don't return any identifiable information about their traces so I was unable to match a certain trace to a curveNumber. I tried to use all the following variables in conjunction but was unable to determine a curveNumber: `gd.data`, `gd.calcdata`, `fullPoint`, `eventData`, `pointData` and `data`. 

I think what needs to happen if possible is to change hover.js to allow for pie and sankey plots to return a curveNumber like all the other types of plots. I would've tried to do that myself but I wasn't sure how to test any changes I would make locally.

src/components/Graph.react.js
```diff
@@ -21,20 +21,44 @@ const filterEventData = (gd, eventData, event) => {
...
                 has('pointNumber', fullPoint) &&
-                has('curveNumber', fullPoint)
-            ) {
-                pointData['customdata'] = data[
-                    pointData.curveNumber
-                ].customdata[fullPoint.pointNumber];
+                has('customdata', data[pointData.curveNumber])) {
+               pointData['customdata'] = data[
+                                            pointData.curveNumber
+                                        ].customdata[fullPoint.pointNumber];
```
chriddyp: this makes sense to me. can you format this to match the style of the rest of the code? In particular, the indenting needs to be updated

src/components/Graph.react.js
```diff
@@ -21,20 +21,44 @@ const filterEventData = (gd, eventData, event) => {
...
+                }
+
+                pointData['customdata'] = gd.data[
+                                                traceNumber
+                                              ].customdata[pointNumber];
+                pointData['index'] = pointNumber;
+              }
+            }
+              points[i] = pointData;
+          }
```
chriddyp: it seems to me that this is just a bug in plotly.js: `pointNumber` and `curveNumber` should be included in the `pointData` for all the chart types, including `pie` and `sankey`. Let's remove this block for now and can you open an issue in `plotly.js` with a reproducable example? https://github.com/plotly/plotly.js/issues/new

chriddyp:
💃 , thanks!

PR #36, opened 2017-07-14
input name attr and hidden type

mjschock:
hi there - any objection to adding a hidden type and name attribute to the input component?

i'm able to add the hidden type anyway as it doesn't seem to be validated but i'm unable to add the name attribute :(

use case:

```
dcc.Input(
    name='_csrf_token',
    type='hidden',
    value=app.server.jinja_env.globals['csrf_token'](),
)
```

### DIFF
--- a/src/components/Input.react.js
+++ b/src/components/Input.react.js
@@ -1,7 +1,7 @@
 import React, {Component, PropTypes} from 'react';
 
 /**
- * A basic HTML input control for entering text, numbers, or passwords.
+ * A basic HTML input control for entering text, numbers, passwords, or hidden fields.
  *
  * Note that checkbox and radio types are supported through
  * the Checklist and RadioItems component. Dates, times, and file uploads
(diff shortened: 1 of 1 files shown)


mjschock:
p.s. loving the libraries btw - also i was able to use `type='email'` too...  is it really necessary to whitelist the types?

mjschock:
in the mean time i created a custom component with the code in this PR - i'm fine with that for now but curious why the plain Input component doesn't allow the 'name' attribute?

chriddyp:
Thanks for posting @mjschock and sorry for the delay! There is no particular reason to exclude all of the attributes - I was just being lazy. I'll add the rest of the properties now: I have a full list of them in the `dash_html_components` library.

chriddyp:
I white-listed the `type`s since not all types are supported across all web browsers. Here's a table of compatibility:

image
from https://developer.mozilla.org/en/docs/Web/HTML/Element/input

PR #40, opened 2017-07-24
Clear the selectedData if the selection box was cleared

chriddyp:
This partially fixes https://github.com/plotly/dash/issues/97 but depends on https://github.com/plotly/plotly.js/issues/1911 getting fixed


chriddyp:
clear-unselect-null

Note how this fix doesn't distinguish between single clicks outside of the box and double clicks that clear the box. Once https://github.com/plotly/plotly.js/issues/1911 is fixed (or some alternative solution is proposed) we can update this solution to only pass `null` if the selection box was indeed cleared.

chriddyp:
New solution using the `plotly_deselect` event:

clear-unselect

PR #41, opened 2017-07-24
[MRG] Add basic textarea component

betatim:
This adds a basic `<textarea>` component. Mostly a straight copy of `<input>`.

Seems to work but the dev instructions in the README don't quite work for me (in terms of what I need to rerun when etc). However a small dash app that uses a textarea instead of an input works.

Should we add some more of the `wrap`, `rows`, `cols` properties? Anything else?

### DIFF
--- a/dash_core_components/metadata.json
+++ b/dash_core_components/metadata.json
@@ -1294,5 +1294,84 @@
         "description": "inline style to be passed to the span wrapping each line if wrapLines is true. Can be either an object or a function that recieves current line number as argument and returns style object."
       }
     }
+  },
+  "src/components/Textarea.react.js": {
+    "description": "A basic HTML textarea for entering multiline text.",
+    "methods": [],
+    "props": {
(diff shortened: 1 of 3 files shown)


chriddyp:
very nice! this looks good to me. I think we should add _all_ of the HTML attributes to `Input` and to `TextArea`. I do this programatically in `dash-html-components` and we could just copy and paste the output like from: https://github.com/plotly/dash-html-components/blob/master/src/components/Textarea.react.js#L30-L177

chriddyp:
@betatim - I'm happy to wrap this up and add the rest of the properties if you'd like. Let me know!

betatim:
I was going to add them either later tonight or tomorrow. If I don't feel free to merge.

chriddyp:
looks good to me. thanks @betatim !!

betatim:
:clinking_glasses: good to help out with a cool project :)

PR #49, opened 2017-08-08
Datepicker Component Addition

alishobeiri:


src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
+import React, {Component, PropTypes} from 'react';
+import { DateRangePicker } from 'react-dates';
+import moment from 'moment';
+
+/**
+ * DatePickerRange is a tailor made component designed for selecting
+ * timespan across multiple days off of a calendar.
```
chriddyp: we might want to mention that it is based off the airbnb date component

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+        initialEndDate: this.props.end_date,
+        firstInitialVisibleMonth: this.props.initial_visible_month,
+        initialMinDateRange: this.props.min_date_range,
+        initialMaxDateRange: this.props.max_date_range
+      };
+
+    }
+
+    convertPropsToMoment(props) {
+      let startDate, endDate=null; let initialVisibleMonth = moment(props.startDate);
```
chriddyp: break each of these out into their own lines

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+      if(props.startDate != undefined) { startDate = moment(props.startDate); }
```
chriddyp: I'm not sure if `undefined` is the right check here. Should this be a check against `null`? If it is `undefined`, then it should be `typeof props.startDate !== 'undefined'` i believe

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+      if(props.endDate != undefined) { endDate = moment(props.endDate); }
+      if(props.initialVisibleMonth != undefined) { initialVisibleMonth = moment(props.initialVisibleMonth); }
+      let min, max;
+      if(props.minDateRange != undefined && props.maxDateRange != undefined) {
+        min = moment(props.minDateRange);
+        max = moment(props.maxDateRange);
+      }
+      try{
```
chriddyp: missing space here

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+        }
+      } catch (TypeError) {
+        // Continue regardless of error
+      }
+
+      return {startDate, endDate, initialVisibleMonth, min, max}
+    }
+
+    componentWillReceiveProps(newProps) {
+      newProps.startDate = newProps.start_date;
```
chriddyp: we shouldn't mutate `newProps` - this will modify the props for the parent component which is unpredicable. Instead, you should make a new object and transfer the properties over to it.

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+      newProps.endDate = newProps.end_date;
+      newProps.initialVisibleMonth = newProps.initial_visible_month;
+      newProps.minDateRange = newProps.min_date_range;
+      newProps.maxDateRange = newProps.max_date_range;
+      const momentProps = this.convertPropsToMoment(newProps);
+      if(this.state.initialStartDate != newProps.start_date) {
+        this.setState({
+          initialStartDate: newProps.start_date,
```
chriddyp: this is a little confusing ot me at first glance - why are we modifying the `initialStartDate`? Shouldn't it be constant?

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+            }}
+            focusedInput={ this.state.focusedInput }
+            onFocusChange={focusedInput => this.setState({ focusedInput })}
+            isOutsideRange={date =>
+              date < this.state.minDateRange || date >= this.state.maxDateRange
+            }
+            showClearDates={ this.props.clearable }
+            disabled={ this.props.disabled }
+            keepOpenOnDateSelect={ this.props.stay_open_on_select }
+            reopenPickerOnClearDates={ this.props.open_calendar_on_clear }
```
chriddyp: should this be `reopen_calendar_on_clear`?

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+            initialVisibleMonth={() => {
+              if(this.state.startDate != null) {
```
chriddyp: missing space after the `if` here

chriddyp: in general, we should use `===` or `!==` wherever possible. Using just `==` or `!=` can lead to some confusing behaviour

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+    }
+}
+
+DatePickerRange.propTypes = {
+    id: PropTypes.string,
+
+    /**
+     * Additional CSS class for the root DOM node
+     */
+    className: PropTypes.string,
```
chriddyp: it doesn't look like this is used

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+
+    /**
+     * Specifies the date that is initially presented when the user
+     * opens the calendar, best practice is to pass value via datetime
+     * object
+     */
+    initial_visible_month: PropTypes.string,
+
+    /**
+     * Text that will be displayed in place of 'Start Date'
```
chriddyp: '`start_date`'?

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+     */
+    start_date_placeholder_text: PropTypes.string,
+
+    /**
+     * Dash-assigned callback that gets fired when the value changes.
```
chriddyp: this comment doesn't look right

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,298 @@
...
+     */
+    end_date_placeholder_text: PropTypes.string,
+
+    /**
+     * If true, the calendar will automatically open when cleared
```
chriddyp: we might want ot pythonify these, so "If True"

chriddyp:
looks good! just a few small suggestions 👍

src/components/DatePickerRange.react.js
alishobeiri: The `startDate` prop might change with user input, so like if the user selects a day on the calendar that won't update the prop `start_date`. So we need to keep track of what `start_date` was previously in `initialStartDate` to determine whether we should change the `startDate` when the value of `start_date` has changed due to a callback. 

Sorry just to clarify it a bit, `start_date` might change from callback input and to determine whether it has changed from callback input we must keep track what the original start date that `start_date` set was because `startDate` might change due to user input.

src/components/DatePickerRange.react.js
alishobeiri: The default palceholder text for the picker is 'Start Date' and 'End Date' for the end date, so thats what I was getting at, I can try to clarify it a bit.

alishobeiri:
@chriddyp Let me know what you think

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+ * which can be found here: https://github.com/airbnb/react-dates
+ */
+export default class DatePickerRange extends Component {
+    constructor(props) {
+      super(props);
+      this.props.startDate = this.props.start_date;
+      this.props.endDate = this.props.end_date;
+      this.props.initialVisibleMonth = this.props.initial_visible_month;
+      this.props.minDateRange = this.props.min_date_range;
+      this.props.maxDateRange = this.props.max_date_range;
```
chriddyp: we should treat `props` as immutable - never modifying elements or creating new ones. Otherwise, the parent component (who is passing in `props` to this component) might end up with `props` in a different structure and that can cause bugs.

So we should remove all of these `this.props.` assignements

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+        }
+      } catch (TypeError) {
+        // Continue regardless of error
+      }
+
+      return {startDate, endDate, initialVisibleMonth, min, max}
+    }
+
+    componentWillReceiveProps(newProps) {
+      let propObj = {
```
chriddyp: this can be `const` I beleive (it's not getting re-assigned)

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+        startDate: newProps.start_date,
+        endDate: newProps.end_date,
+        initialVisibleMonth: newProps.initial_visible_month,
+        minDateRange: newProps.min_date_range,
+        maxDateRange: newProps.max_date_range
+      }
+      const momentProps = this.convertPropsToMoment(propObj);
+      if (this.state.initialStartDate != newProps.start_date) {
```
chriddyp: this should be `!==`

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+        this.setState({
+          initialStartDate: newProps.start_date,
+          startDate: momentProps.startDate
+        });
+      }
+      if (this.state.initialEndDate != newProps.end_date) {
+        this.setState({
+          initialEndDate: newProps.end_date,
```
chriddyp: should this name be changed? `initial` to me means that it's only set once, when the component is rendered. But in this case, we're re-setting the value everytime the component gets new props.

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+                if (setProps) {
+                  setProps({
+                    startDate: startDateStr
+                  });
+                }
+                if (fireEvent) {
+                  fireEvent('change');
+                }
+              }
+              this.props.start_date = startDate;
```
chriddyp: similarly here - we should not mutate `this.props`

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+          maxDateRange: momentProps.maxDateRange
+        })
+      }
+    }
+
+    render() {
+        const {setProps, fireEvent} = this.props
+        return (
+          <DateRangePicker
+            startDate={ this.state.startDate }
```
chriddyp: I believe that the convention is no spaces around `{`. Can you run eslint on this? It'll print out all the errors. 
```
./node_modules/.bin/builder run lint
```

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+          startDate: momentProps.startDate
+        });
+      }
+      if (this.state.initialEndDate != newProps.end_date) {
+        this.setState({
+          initialEndDate: newProps.end_date,
+          endDate: momentProps.endDate
+        });
+      }
+      if (this.state.firstInitialVisibleMonth != newProps.initial_visible_month){
```
chriddyp: similarly here: `!==` not `!=`

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+      }
+      if (this.state.initialMinDateRange != newProps.min_date_range){
+        this.setState({
+          initialMinDateRange: newProps.min_date_range,
+          minDateRange: momentProps.minDateRange
+        })
+      }
+      if (this.state.initialMaxDateRange != newProps.max_date_range){
+        this.setState({
+          initialMaxDateRange: newProps.max_date_range,
```
chriddyp: it looks like `initialMaxDateRange` isn't being used anywhere (it's not in the `render` method). Should it be? If not, can we remove it?

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+      }
+      if (this.state.initialEndDate != newProps.end_date) {
+        this.setState({
+          initialEndDate: newProps.end_date,
+          endDate: momentProps.endDate
+        });
+      }
+      if (this.state.firstInitialVisibleMonth != newProps.initial_visible_month){
+        this.setState({
+          firstInitialVisibleMonth: newProps.initial_visible_month,
```
chriddyp: Similarly here - it looks like `firstInitialVisibleMonth` isn't being used anywhere (it's not in the render method). Should it be? If not, can we remove it?

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+            onFocusChange={focusedInput => this.setState({ focusedInput })}
+            isOutsideRange={date =>
+              date < this.state.minDateRange || date >= this.state.maxDateRange
+            }
+            showClearDates={ this.props.clearable }
+            disabled={ this.props.disabled }
+            keepOpenOnDateSelect={ this.props.stay_open_on_select }
+            reopenPickerOnClearDates={ this.props.reopen_calendar_on_clear }
+            initialVisibleMonth={() => {
+              if (this.state.startDate != null) {
```
chriddyp: `!==` instead of `!=`

chriddyp: see https://stackoverflow.com/questions/38977829/what-is-the-difference-between-null-and-null

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+        );
+    }
+}
+
+DatePickerRange.propTypes = {
+    id: PropTypes.string,
+
+    /**
+     * Specifies the starting date for the component, best practice is to pass
+     * value via datetime object
```
chriddyp: I'd change this to:
```
Specifies the starting date of the component. Accepts `datetime.datetime` objects or strings in the format `YYYY-MM-DD`
```
(or w/e the accepted format is)

chriddyp: and similarly edit the `end_date` below

chriddyp: and the rest of the components. i.e. replace "best practice is to pass value via datetime object" with "Accepts `datetime.datetime` objects or strings in the format 'YYYY-MM-DD'"

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+     * Specifies the ending date for the component, best practice is to pass
+     * value via datetime object
+     */
+    end_date: PropTypes.string,
+
+    /**
+     * Specifies the lowest selectable date for the component,
+     * best practice is to pass value via datetime object
+     */
+    min_date_range: PropTypes.string,
```
chriddyp: for me, `range` feels like a set of values. "Range of values" being something like "from 0 to 5". Would a better name for this be something like `min_date_allowed`?

chriddyp: similarly for `max_date_allowed` below

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+     * best practice is to pass value via datetime object
+     */
+    max_date_range: PropTypes.string,
+
+    /**
+     * Specifies the date that is initially presented when the user
+     * opens the calendar, best practice is to pass value via datetime
+     * object
+     */
+    initial_visible_month: PropTypes.string,
```
chriddyp: this is a little bit confusing to me. If this just specifies the month, then why does it take an entire date string? should this be `initial_visible_date`?

chriddyp: Oh, is it because the datepicker only shows a single month at a time? But instead of passing in a single month integer, you pass in an entire date string. Is that how it works? If so, I guess that makes sense. I would sort of expect this attribute to take an integer from 1 to 12, but I guess a datestring will be just as easy.

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * Text that will be displayed in the first input
+     * box of the date picker when no date is selected - replaces 'Start Date'
```
chriddyp: Instead of `replaces 'Start Date'`, should it be `. Default value is 'Start Date'.`

chriddyp: similarly below with `End Date`

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * If True, calendar will open in a full screen overlay portal, will
+     * take precedent over 'withPortal' if both are set to true
+     */
+    with_full_screen_portal: PropTypes.bool,
+
+    /**
+     * Specifies what day is the first day of the week, values must be
+     * from [0, ..., 6]
```
chriddyp: would it be helpful to say that `0` corresponds to Monday / Sunday? Or is that configurable? The day that starts the week is different around the world I believe

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * If True the calendar will display days that rollover into
+     * the next month
+     */
+    show_outside_days: PropTypes.bool,
+
+    /**
+     * Specifies the format that the month will be displayed in the calendar,
+     * valid formats are variations of "MM YY"
```
chriddyp: This is a little confusing to me. Why is `YY` in the `month_format` section? It would be helpful to have a couple of examples here, like:
```
...valid formats are variations of "MM YY". For example:
- "MM" will render as "12"
- "MMMM" will render as "1212"
```

or however it works

src/components/DatePickerRange.react.js
```diff
@@ -0,0 +1,306 @@
...
+     */
+    month_format: PropTypes.string,
+
+    /**
+     * Specifies the format that the selected dates will be displayed
+     * valid formats are variations of "MM YY DD"
+     */
```
chriddyp: similarly here - a few examples would be helpful

src/components/DatePickerSingle.react.js
```diff
@@ -0,0 +1,250 @@
+import React, {Component, PropTypes} from 'react';
```
chriddyp: similar comments above apply here

src/components/DatePickerRange.react.js
alishobeiri: The values `initialMaxDateRange` and the rest are being used in the if statements to determine whether the new callback value has changed them. So when a callback changes the value of max_date_range we check to see whether it has changed with regard to its previous value `initialMaxDateRange` and if so we update the state.

I agree it is a pretty confusing way to set it up though, I've changed the variable names to `prevMaxDateAllowed` to make it less confusing

src/components/DatePickerRange.react.js
alishobeiri: Yeah, the date picker can show more than one month at time but you're essentially selecting which month it should show. 

You need to pass a year and a month to determine which month the calendar would open to, the day isn't actually required but I thought it'd just be simpler to keep everything using datetime object.

alishobeiri:
@chriddyp Let me know what you think when you get a chance, I ran eslint but it seemed to only run on the bundle so it didn't really help to highlight the syntax as much. 

I tried to adhere to what I saw by hand and I kept the javascript in line with jscs styleguide using an atom linter.

app.py
```diff
@@ -29,6 +30,34 @@
 
     '''),
 
+    dcc.DatePickerSingle(
+        id='date-picker-single',
+        initial_visible_month=dt.datetime(1997, 5, 5),
+        min_date_range=dt.datetime(1997, 4, 29),
```
chriddyp: Does this example need to get updated? I get 
```
Exception: Unexpected keyword argument `min_date_range`
Allowed arguments: id, date, min_date_allowed, max_date_allowed, initial_visible_month, placeholder, reopen_calendar_on_clear, number_of_months_shown, with_portal, with_full_screen_portal, first_day_of_week, stay_open_on_select, show_outside_days, month_format, display_format, disabled, clearable
```

chriddyp: although we can actually just remove this example all together. I think that I commited it accidently.

PR #51, opened 2017-08-09
Date picker verticalfix

alishobeiri:
disabled portals on vertical calendar

### DIFF
--- a/src/components/DatePickerRange.react.js
+++ b/src/components/DatePickerRange.react.js
@@ -132,6 +132,11 @@ export default class DatePickerRange extends Component {
 
   render() {
     const { setProps, fireEvent } = this.props;
+    let verticalFlag = true;
+    if (this.props.calendar_orientation === 'vertical') {
+      verticalFlag = false;
+    }
+
     return (
       <DateRangePicker
         startDate={this.state.startDate}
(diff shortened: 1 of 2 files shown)


src/components/DatePickerRange.react.js
```diff
@@ -132,6 +132,11 @@ export default class DatePickerRange extends Component {
 
   render() {
     const { setProps, fireEvent } = this.props;
+    let verticalFlag = true;
+    if (this.props.calendar_orientation === 'vertical') {
+      verticalFlag = false;
+    }
```
chriddyp: this can be simplified to `const isVertical = (calendar_orientation === 'vertical')`

src/components/DatePickerRange.react.js
```diff
@@ -132,6 +132,11 @@ export default class DatePickerRange extends Component {
 
   render() {
     const { setProps, fireEvent } = this.props;
+    let verticalFlag = true;
+    if (this.props.calendar_orientation === 'vertical') {
```
chriddyp: put `calendar_orientation` in the `this.props` assignment above: `const { calendar_orientation, setProps, fireEvent } = this.props;`

chriddyp:
looks good to me. can you make bump the version's to `0.12.1` and add an entry in the changelog too?

PR #65, opened 2017-08-25
Changed marks PropType to match RangeSlider

alishobeiri:
Switch allows `marks` to be better documented in the dash-doc tables


chriddyp:
💃

PR #66, opened 2017-08-29
Add the ability to disable the dcc.Interval property

alishobeiri:
As per community suggestion:
https://community.plot.ly/t/how-to-turn-off-interval-event/5565


chriddyp:
this can effectively be done currently by setting `interval` to some huge number, but this is a better solution. 💃

PR #75, opened 2017-09-18
Dropdown search support for non-ascii characters

chriddyp:
fixes issue reported in https://community.plot.ly/t/autocomplete-with-dropdown-is-invalid-for-chinese/5818


chriddyp:
Once merged, this will be available on `v0.12.6`. Upgrade with `pip install dash-core-components==0.12.6`

PR #100, opened 2017-10-19
Input with type number should return number after value change

Madhu94:
Attempting to close https://github.com/plotly/dash/issues/124. @chriddyp, please let me know if I should do something else here.

### DIFF
--- a/src/components/Input.react.js
+++ b/src/components/Input.react.js
@@ -21,15 +21,22 @@ export default class Input extends Component {
     render() {
         const {
             fireEvent,
-            setProps
+            setProps,
+            type
         } = this.props;
         const {value} = this.state;
-
         return (
             <input
                 onChange={e => {
                     this.setState({value: e.target.value});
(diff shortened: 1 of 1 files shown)


src/components/Input.react.js
```diff
@@ -21,15 +21,22 @@ export default class Input extends Component {
     render() {
         const {
             fireEvent,
-            setProps
+            setProps,
+            inputType
```
chriddyp: I believe that this should be `type` to match https://github.com/plotly/dash-core-components/pull/100/files#diff-368d71637f90189b9dab46fb80b2c1f1R78

chriddyp:
This looks great @Madhu94, thank you! Just one small comment about the `prop` name.

Madhu94:
@chriddyp Is this good to go now ?

chriddyp:
Looks great @Madhu94 , thank you!

PR #105, opened 2017-10-31
Allow children in Checklist options

Akronix:
I'm developing a webapp which needs to have checkboxes that contain another (Dash's) React 
component, so I've modified the Checklist component to allow to include options with children components, as well as a button to show/hide those children.

Here is a short "demo":
anim

This is a first approach, so take it as a prototype. We can work on it and improve it until it reaches a production-ready code and then we can squash all commits and merge with master.

### DIFF
--- a/demo/Demo.react.js
+++ b/demo/Demo.react.js
@@ -299,6 +299,48 @@ class Controller extends Component {
 ReactDOM.render(<Controller/>, mountNode);`
 
 
+const ChecklistContainerExample = `
+const applesDescription = (<div style={{marginLeft: '20px', marginBottom: '20px'}}>
+                                <p><img src="https://upload.wikimedia.org/wikipedia/commons/thumb/e/ee/Apples.jpg/97px-Apples.jpg" alt="Apples.jpg" /></p>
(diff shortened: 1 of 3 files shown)


chriddyp:
Thanks @Akronix ! I think that this is out of the scope for this core checkbox component - I'd like to keep the options as simple possible and this feels like a more specialized UI that would be better suited in a fork (something like `CollapsableChecklist`).

Another note - once something like https://github.com/plotly/dash-renderer/pull/26 is added, this behaviour will be able to be specified with something like:
```
dcc.CheckList(options=[
    {
        'value': 'melons',
        'label': html.Details([
            html.Summary('Label of the item'),
            html.Div(html.Img(src='...'), html.Div('More about the mango option')
        ])
    }
])
```

In other words, we'll be able to allow the user to supply any component as the `label`, not just strings. In this case, the `html.Details` component hides and shows content but users could supply any type of label: `html.H1`, `html.Img`, etc

Akronix:
Wow,  plotly/dash-renderer#26 looks very promising!
And ambitious too. So, until it is done, I'll keep CollapsableChecklist in our "dash components suit: https://github.com/Grasia/grasia-dash-components

PR #117, opened 2017-11-24
v0.15.2 - Fix remote build

chriddyp:


chriddyp:
New `dcc.Interval` usage:

```
import dash

app = dash.Dash()

app.layout = html.Div([
    html.Div(id='my-output-interval'),
    dcc.Interval(id='my-interval', interval=500),
])


@app.callback(
    Output('my-output-interval', 'children'),
    [Input('my-interval', 'n_intervals')])
def display_output(n):
    return '{} have passed'.format(n)


if __name__ == '__main__':
    app.run_server(debug=True)
```

PR #120, opened 2017-11-30
Remove unused imports. Standard imports should be placed before other…

jbampton:
… imports https://www.python.org/dev/peps/pep-0008/#imports


chriddyp:
Thanks @jbampton !

PR #121, opened 2017-12-02
Use single quotes consistently. The python code in this project uses …

jbampton:
…single quotes except for this file which had both single and double quotes.


chriddyp:
Thanks @jbampton !

PR #122, opened 2017-12-02
Remove duplicate CSS white-space key. The white-space CSS property is…

jbampton:
… specified as a single keyword chosen from the list of values as shown here: https://developer.mozilla.org/en-US/docs/Web/CSS/white-space


jbampton:
Requirement that the keys are unique (within one dictionary)  
 https://docs.python.org/3/tutorial/datastructures.html#dictionaries

chriddyp:
:woman_facepalming: thanks @jbampton !

PR #128, opened 2017-12-15
Support Dash-Renderer 2.0 - *Breaking Changes*

chriddyp:
This PR updates the components to support the features introduced in https://github.com/plotly/dash-renderer/pull/32.

Namely, it will enable various properties across various components to accept arbitrary dash components as values. 

For example, the `label` property in `dcc.RadioItems` will be able to take a arbitrary Dash component instead of just a string.

i.e., instead of this:
```python
dcc.RadioItems(
    options=[
        {‘label’: ‘Item 1’, ‘value’: 1}, 
        {‘label’: ‘Item 2’, ‘value’: 1}
    ]
)
```

the developer could do this:
```python
dcc.RadioItems(
    options=[
        {‘label’: html.Div(‘Item 1’, style={‘color’: ‘cyan’}), ‘value’: 1}, 
        {‘label’: ‘Item 2’, ‘value’: 1}
    ]
)
```

This would require a dependency upgrade of `dash-renderer` to a 2.0 series, which would cause other component libraries that are not compatible with that version to break. There should be no breaking changes within this library.

This PR is a work-in-progress.

### DIFF
--- a/dash_core_components/metadata.json
+++ b/dash_core_components/metadata.json
@@ -1911,6 +1911,13 @@
         },
         "required": false,
         "description": ""
+      },
+      "render": {
+        "type": {
+          "name": "func"
+        },
+        "required": false,
+        "description": "Dash-assigned callback for rendering"
       }
     }
   },
--- a/src/components/RadioItems.react.js
+++ b/src/components/RadioItems.react.js
@@ -28,7 +28,8 @@ export default class RadioItems extends Component {
             labelClassName,
             labelStyle,
             options,
-            setProps
+            setProps,
+            render
         } = this.props;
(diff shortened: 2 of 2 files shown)


chriddyp:
Prerelease versions are available with this code
```
pip install dash-core-components==0.16.0rc1
pip install dash-html-components==0.9.0rc1
pip install dash-renderer==0.13.0rc2
pip install dash-table-experiments==0.6.0rc1
```

valentijnnieman:
@chriddyp Do we still want to keep this PR open, with regards to the new plan for 1.0 https://github.com/plotly/dash/issues/469, and seeing as this is almost a year old?

Issue #113, opened 2017-11-23
dcc.Link or dcc.Location isn't firing callbacks for IE 11


chriddyp:
![image](https://user-images.githubusercontent.com/1280389/33154852-596c0f18-cfb9-11e7-900c-65e7e6ad2117.png)

originally reported in https://community.plot.ly/t/multi-page-dash-app-internet-explorer-issue/6206/3

ned2:
I believe the issue is that IE does not support the `Event` constructor, which `dcc.Link` uses to trigger a `pushstate` event. This in turn means that the `dcc.Location` listener will never pick it up.

This [Stack Overflow post](https://stackoverflow.com/questions/26596123/internet-explorer-9-10-11-event-constructor-doesnt-work) links to a MSDN page which includes a polyfill for `CustomEvent` as a workaround.

PR #130, opened 2017-12-18
sketch of solution to fix #129

### LINKED ISSUE #129: Difficulty emulating vanilla anchor tags with dcc.Link
The `dcc.Link` component renders as `<a>` tags without an `href` attribute, since the routing is all taken care of in the React side of things. This means that out of the box these links don't receive the default linkified browser styles of underline and pointer cursor on hover etc as well as the browser providing a preview of the link address in the bottom left corner. It would be nice to have this behaviour.

The styles can be emulated through custom CSS but not the href preview on hover. I've tried working around this by using an anchor tag within the `dcc.Link` like this:

```python
href = '/some-path'
dcc.Link(html.A(text, href=href), href=href)
```
All the styles and hover behaviour works as expected, but the problem is that when clicking on the link, in addition to Dash handling the `dcc.Link`'s callback, the nested `<a>` also gets a click event which then causes a whole page redirect, reloading the whole Dash app, which is obviously no good. I played around with using the CSS pointer-event attribute to disable the internal `<a>`, however this then disables all the hover events wanted in the first place.

Would it be feasible to add the `href` to the `<a>` produced by `dcc.Link`, modifying its `onclick` to prevent the page being redirected?

chriddyp:


ned2:
Well that's a pretty good start! I've verified that the solution works in all major browser with the exception of two. For Safari, I still need to track down an Apple machine, but can do shortly. 

IE, on the other hand, is interesting. It turns out that even before this modification, `dcc.Link` did not actually work. I believe that the `dcc.Link` has _never_ worked in IE, as it [does not support the Event constructor](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event). Just noticed the existing issue #113 about this and have added a comment there. 

I'm not really sure what's needed for an integration test. Would you be able to provide some guidance as what would be required? Would this be a separate test independent to `test_gallery`?

chriddyp:
> IE, on the other hand, is interesting. It turns out that even before this modification, dcc.Link did not actually work. I believe that the dcc.Link has never worked in IE, as it does not support the Event constructor. Just noticed the existing issue #113 about this and have added a comment there.

Thanks for reporting! Fix in https://github.com/plotly/dash-core-components/pull/140

> Would you be able to provide some guidance as what would be required? Would this be a separate test independent to test_gallery?

Gladly! It would be a separate test. The main thing that I'd want to test is that clicking on `dcc.Link` won't refresh the page - that is, it will act as `dcc.Link` currently works, without refreshing the page - that is, that `dcc.Link` caused the update to happen, not the `html.A`.
One way to test this might be to set `app.layout` to be a function and assert that it is only fired once, on page start. If it were to fire multiple times, then that would mean that the page was being refreshed and that the click event is firing the `html.A` instead of the `dcc.Link`.

Most of the integration tests are in https://github.com/plotly/dash-renderer/blob/master/tests/test_render.py - that's a good place to reference. In particular, we'll probably need to use the "call_count" logic - a multi-processing-safe value that gets incremented when a function gets called:
https://github.com/plotly/dash-renderer/blob/cef2313478ee2c37b91b9ef773bf6991bc96240e/tests/test_render.py#L433-L440

Here's a sketch of what this particular test might look like (I haven't tested this). I'm using some new functions (`wait_for_element_by_css_selector`, `wait_for_text_to_equal`) that were introduced in #140 .

```python
    def test_update_link_click(self):
        app = dash.Dash(__name__)
        def serve_layout():
            return html.Div([
                html.Div(id='waitfor'),
                dcc.Location(id='location')
                dcc.Link('Link', href='/test', id='link')
                html.Div(id='output')
            ])
        app.layout = serve_layout
        call_count = Value('i', 0)
        
        @app.callback(Output('output', 'children'), [Input('link', 'href')])
        def update_callback(href):
            call_count.value = call_count.value + 1
            return href

        self.startServer(app=app)
        self.wait_for_element_by_css_selector('#waitfor')
        self.wait_for_text_to_equal('#output', '/')

        self.assertEqual(call_count.value, 1)
        self.wait_for_element_by_css_selector('#link').click()
        self.wait_for_text_to_equal('#output', '/test')
        self.assertEqual(call_count.value, 1)
```

To run this test, you can do 
```
$ python -m unittest test.test_integration.Tests.test_update_link_click
```

To debug this test, I recommend putting in a 
```
import ipdb; ipdb.set_trace()
```
inside your test. That'll open up the python debugger allowing you to execute commands in your python console. As you execute selenium commands (like `self.driver.find_element_by_id('wait_for')`), they'll get run in the browser window that you have open. It's like magic 🐰 🎩 :slightly_smiling_face:

Issue #81, opened 2017-09-25
url parameters are not available in `dcc.Location`


chriddyp:
we're only sending the `window.location.pathname` (https://github.com/plotly/dash-core-components/blob/master/src/components/Location.react.js#L36) but we could send an object of `window.location.search` (https://stackoverflow.com/questions/5448545/how-to-retrieve-get-parameters-from-javascript)

as reported in https://community.plot.ly/t/get-url-with-get-parameters/5878/1

mjclawar:
@chriddyp is it OK if I take this one at some point in the next week? We've run into multiple situations where it would be very helpful. 

Do you have any thoughts on changing the `Link` to support `window.location.search` as well?

As a general question: Are there contribution guidelines for this repo (e.g., PRs, building, testing) hiding anywhere? Is it OK to just change the `src` directory, add relevant tests, and then submit a PR when the issue is addressed?

Thanks!

chriddyp:
@mjclawar  - that would be great!

There are not any contributing guidelines yet, so here is a summary:
### Testing
- We have integration tests that start a dash app and run a selenium, see https://github.com/plotly/dash-core-components/blob/c898d75a1a189943bcf92bc73b95bd08432ddd7d/test/test_integration.py#L147
- We have image tests - these tests take a snapshot of the screen _during_ an integration test and can be used instead of tedious testing of each expected out (so long as you display your outputs on the screen). Here's an example: https://github.com/plotly/dash-core-components/blob/c898d75a1a189943bcf92bc73b95bd08432ddd7d/test/test_integration.py#L198
- I recommend using `import ipdb; ipdb.set_trace()` to set breakpoints while running these selenium tests, especially when trying to do complex selenium things. selenium isn't super well documented. in fact, you can use this workflow to write your test entirely in the console, and then paste the working commands back into the actual testing code
- These integration tests are awesome because you can test the actual behaviour in the Python dash code
- We don't have any tests for `Link` and how `Link` works with `Location`. It's definitely time to add them.

### Developing
- `npm run prepublish` will generate the JS bundle and the `metadata.json` file that's required for generating the python components
- I usually run it with a file watcher, like `$ ag -l --js | entr npm run prepublish`
- Make sure to set `app.scripts.config.serve_locally` and `app.css.config.serve_locally=True` if you are working on an app in development

### Releasing
- It'd be helpful, at the end of the PR, for you to bump the version number appropriately and update the `CHANGELOG.md` file
[...]

PR #131, opened 2017-12-21
81 location url parameters

mjclawar:
### Description
This PR adds `search`, `hash` and `href` props to the `dcc.Location` component (in addition to the already-existing `pathname`) to better mimic the `window.location` API. 
This addresses #81, with the caveat that `dcc.Link` does not have any changes to props that would allow a user to directly update `search` independently of `hash` (e.g., using `dcc.Link(href="?testQuery=testValue")` will wipe `hash` from `dcc.Location`. Note that this is the same behavior as `window.location`, so should be considered "correct", but it seemed like updating `dcc.Link` should be a different issue.

### Implementation
- The changes to `dcc.Location` pretty much extends the previous `pathname`-specific `updateLocation` approach. The only major difference is that if `href` changes in props compared to previous props and compared to `window.location.href`, `href` takes precedent over the combination of `${pathname}${searchVal}${hashVal}` ([see here](https://github.com/mjclawar/dash-core-components/blob/81-location-url-parameters/src/components/Location.react.js#L63)).
- gitignore was updated for IntelliJ `.idea/`, but can remove that if needed. 
- Added a `package-lock.json`

### Tests added
This adds tests for `html.A` tags, `dcc.Link`, and `html.Button` clicks interacting with the updated `dcc.Location` component.
peek 2017-12-21 09-31


src/components/Location.react.js
```diff
@@ -1,39 +1,85 @@
...
-                // Refresh the page
-                window.location.pathname = pathname;
-            } else {
-                // Just push the new location into the URL
-                if (setProps) setProps({pathname});
-                window.history.pushState({}, '', pathname);
+        const checkExistsUpdateWindowLocation = (fieldName) => {
+            const propVal = props[fieldName];
+
+            if (!propVal && window.location[fieldName] && setProps) { // Prop is undefined?
```
chriddyp: we use stronger checks for `undefined` here like `R.type(propVal) == 'Undefined'` or even `!R.has(props, propVal)` (http://ramdajs.com/docs/#type, http://ramdajs.com/docs/#has)

src/components/Location.react.js
```diff
@@ -1,39 +1,85 @@
...
+                const update = {};
+                update[fieldName] = window.location[fieldName];
+                setProps(update);
```
chriddyp: Instead of updating the props 1 by 1, I think we should collect all of the properties and update them with a single call to `setProps`. Otherwise, I believe that Dash will rerender the app on every call to `setProps`

src/components/Location.react.js
```diff
@@ -1,39 +1,85 @@
...
+            } else if (propVal && (propVal !== window.location[fieldName])) { // Prop has changed?
+                if (refresh) { // Refresh the page?
+                    window.location[fieldName] = propVal;
+                } else if ((this.props[fieldName] !== propVal)) { // If this prop has changed, need to setProps
```
chriddyp: ✏️ - Extra set of parentheses here: `((` > `(`

src/components/Location.react.js
```diff
@@ -1,39 +1,85 @@
...
+                    const update = {};
+                    update[fieldName] = propVal;
+
+                    if (setProps)
+                        setProps(update);
```
chriddyp: just for consistency, could you add `{` and `}` around this  block?

src/components/Location.react.js
```diff
@@ -1,39 +1,85 @@
...
+        // Check if the prop value needs to be updated (note that this mutates propsToSet)
+        const pathnameUpdated = checkExistsUpdateWindowLocation('pathname');
+        const hrefUpdated = checkExistsUpdateWindowLocation('href');
+        const hashUpdated = checkExistsUpdateWindowLocation('hash');
+        const searchUpdated = checkExistsUpdateWindowLocation('search');
+
+        if (hrefUpdated) // Special case -- overrides everything!
+            window.history.pushState({}, '', href);
+        else if (pathnameUpdated || hashUpdated || searchUpdated) { // Otherwise, we can mash everything together
+            const searchVal = typeof search !== 'undefined' ? search : '';
```
chriddyp: for consistency, I use Ramda in the lib, so instead of `typeof search !== 'undefined'`, `R.Type(search) !== 'Undefined'`

src/components/Location.react.js
```diff
@@ -55,7 +101,17 @@ export default class Location extends Component {
...
+    /** pathname in window.location - e.g., "/my/full/pathname" */
     pathname: PropTypes.string,
+    /** search in window.location - e.g., "?myargument=1" */
+    search: PropTypes.string,
+    /** hash in window.location - e.g., "#myhash" */
+    hash: PropTypes.string,
+    /** href in window.location - e.g., "/my/full/pathname?myargument=1#myhash" */
+    href: PropTypes.string,
+
+    /** Refresh the page when the location is updated? */
```
chriddyp: 👍 very nice, thanks for adding these

test/test_integration.py
```diff
@@ -324,3 +327,140 @@ def test_gallery(self):
...
+            html.Div(id='test-pathname', children=[]),
+            html.Div(id='test-hash', children=[]),
+            html.Div(id='test-search', children=[]),
+        ])
+
+        @app.callback(
+            output=Output(component_id='test-pathname', component_property='children'),
+            inputs=[Input(component_id='test-location', component_property='pathname')])
+        def update_location_on_page(pathname):
+            assert isinstance(pathname, str)
```
chriddyp: I believe that these asserts won't actually fail the test as the dash app is run in a subprocess in these tests. They'll still cause the callback to fail (which might implicitly break other assertions in the main process) but in either case we might want to put this assertion out of this context

test/test_integration.py
```diff
@@ -324,3 +327,140 @@ def test_gallery(self):
...
+            state=[State(component_id='test-location', component_property='pathname')])
+        def update_pathname(n_clicks, current_pathname):
+            if n_clicks is not None:
+                return '/new/pathname'
+
+            return current_pathname
+
+        self.startServer(app=app)
+
+        def _waiter():
```
chriddyp: 👍  - this is great. can we move this to `test/utils.py` so that we can reuse it elsewhere?

test/test_integration.py
```diff
@@ -324,3 +327,140 @@ def test_gallery(self):
...
+        # Check that search is updated in the Location -- note that this goes through href and therefore wipes the hash
+        self.driver.find_element_by_id('test-link-search').click()
+
+        self.snapshot('link -- /test/pathname?testQuery=testValue')
+        self.assertEqual(self.driver.find_element_by_id('test-search').text, '?testQuery=testValue')
+        self.assertEqual(self.driver.find_element_by_id('test-hash').text, '')
+
+        # Check that pathname is updated through a Button click via props
+        self.driver.find_element_by_id('test-button').click()
+        time.sleep(1)  # Need to wait for the callback to fire TODO is there a better way to wait?
```
chriddyp: There isn't really, besides using a `wait_for` with whatever you are waiting for

test/test_integration.py
```diff
@@ -324,3 +327,140 @@ def test_gallery(self):
...
+
+        self.snapshot('link -- /new/pathname?testQuery=testValue')
+        self.assertEqual(self.driver.find_element_by_id('test-pathname').text, '/new/pathname')
```
chriddyp: for example, you _should_ be able to remove the `time.sleep` and replace it with a `wait_for` here and place the snapshot below the assertions

chriddyp:
> This addresses #81, with the caveat that dcc.Link does not have any changes to props that would allow a user to directly update search independently of hash (e.g., using dcc.Link(href="?testQuery=testValue") will wipe hash from dcc.Location. Note that this is the same behavior as window.location, so should be considered "correct", but it seemed like updating dcc.Link should be a different issue.

Could you open up a new issue about this? My guess is that we are OK here. I believe that the main use for `dcc.Link` will be just to set the `href`.

src/components/Location.react.js
mjclawar: OK. That's what I had originally done, but was running into some issues so broke them up to make the logic clearer in testing. I'll change it back since the code is working now with tests :+1:

src/components/Location.react.js
mjclawar: Will do.

src/components/Location.react.js
mjclawar: Great. Haven't used that before so appreciate the example here and the links to the documentation in another comment.

src/components/Location.react.js
mjclawar: It's actually a null in some cases (pathname when the page first loads, e.g.), so will update to `(R.type(propVal) === 'Undefined' || propVal === null)` if you prefer. I think that's why the original code had `!propVal`.

src/components/Location.react.js
mjclawar: :see_no_evil:

chriddyp:
@mjclawar  - Let me know when you're ready for another review!

mjclawar:
All set for another review.

mjclawar:
@chriddyp had to push a change to uncomment the rest of the test suite. I think all the changes you requested should be in there now!

chriddyp:
Thanks! It looks like tests are failing because of some dependency mismatched. I'll make a separate PR to lock down these versions in tests

chriddyp:
This looks good to me @mjclawar . Once we get the tests to pass (hopefully through #136 ), I'm happy to merge this.

mjclawar:
Thanks -- wasn't sure why the tests were failing on CircleCI. Happy new year! :tada:

chriddyp:
@mjclawar - #136 fixed the tests. Would you mind rebasing your branch off of master?

mjclawar:
Rebased.

chriddyp:
Tests passed! 🎉  Thank you @mjclawar , this was a great PR 🍻

mjclawar:
@chriddyp do you have an expected timeline for getting this published as the next version/how do you typically approach publishing for the dash packages? Thanks!

chriddyp:
Hey @mjclawar - Thanks for the reminder! I just published `v0.15.4` now

Amoners:
I want to use this feature, but I don't know how to run the test. Is there any document about this?

PR #140, opened 2018-01-08
Fix Link and Location for IE 

chriddyp:
Hopefully fixes https://github.com/plotly/dash-core-components/issues/113


chriddyp:
Works on Edge and IE11 🎉 
image
image

test/test_integration.py
```diff
@@ -34,21 +34,28 @@
...
+    def wait_for_text_to_equal(self, selector, assertion_text):
+        start_time = time.time()
+        while time.time() < start_time + 20:
+            el = self.wait_for_element_by_css_selector(selector)
+            try:
+                return self.assertEqual(el.text, assertion_text)
+            except Exception as e:
+                pass
+            time.sleep(0.25)
+        raise e
```
chriddyp: FYI @mjclawar ^^ I rewrote some of the `wait_for` logic so that we don't have to do any more fragile `time.sleep`. Basically just waiting (up to 20 seconds) for our test assertions to come true, otherwise throwing an exception. See https://github.com/plotly/dash-core-components/pull/140/files#diff-67ec2bf00b0a54c0fa8d06ee654230d5R435 for usage.

PR #142, opened 2018-01-11
Fix undefined reference to __version__ in setup.py.  Pylint showed an…

jbampton:
… undefined variable error. You can test this fix by changing the 'version' in version.py or setup.py and then running from the command line: python3 setup.py --version


chriddyp:
Thanks @jbampton !

jbampton:
👍

PR #160, opened 2018-02-11
Move PropTypes from React package to prop-types

mjclawar:
### Description
This moves the `PropTypes` import from `react` to `prop-types`.

Example:
```js
import React, { PropTypes } from 'react';
```

is now
```js
import React from 'react';
```

This will allow `dash-core-components` to support React 16+ by patching in different versions to the `dash-renderer` by the Dash developer (for now).

For example, in the application definition file:
```python
import dash_renderer
dash_renderer._js_dist_dependencies = [
                {
                    'external_url': [
                        'https://unpkg.com/react@16/umd/react.production.min.js',
                        'https://unpkg.com/react-dom@16/umd/react-dom.production.min.js'
                    ],
                    'relative_package_path': [
                        'react.production.min.js',  # where the user has added this file to the installed dash-renderer for now
                        'react-dom.production.min.js'  # where the user has added this file to installed dash-renderer for now
                    ],
                    'namespace': 'dash_renderer'
                }
            ]
```

### Changes
- Added `package-lock.json`
- Ran the https://github.com/reactjs/react-codemod#react-proptypes-to-prop-types PropTypes codemod from React organization
- Updated the version to 0.19.0


src/components/Upload.react.js
```diff
@@ -1,4 +1,5 @@
-import React, {Component, PropTypes} from 'react';
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
```
chriddyp: a few nit-picky things. I have the personal convention of using `{Component}` rather than `{ Component }`

src/components/Upload.react.js
```diff
@@ -252,4 +253,4 @@ Upload.defaultProps = {
         borderColor: '#c66',
         backgroundColor: '#eee'
     }
-};
+};
```
chriddyp: Also, my editor will automatically add new lines to the ends of files whereas it looks like yours removes them. Could we add these newlines back?

chriddyp: Otherwise the next time I make a PR, it'll have these superfluous diffs

mjclawar: Mind if I add a .editorconfig with a newline at the end of all files then?

chriddyp:
Looks good to me here too. Once we change `{ Component }` to `{Component}` and add the new lines back, I'll merge this

src/components/Upload.react.js
mjclawar: I think this was the codemod that made all the style changes at the end of files.

mjclawar:
@chriddyp fixed the style mismatches. I added a `.editorconfig` file that makes sure the newline gets added to future files, but I can remove that if you prefer to avoid adding that file.

chriddyp:
`.editorconfig` - very nice! didn't know about this

PR #170, opened 2018-02-28
swap out `Plotly.newPlot` with `Plotly.react`

chriddyp:
Solves issues from https://github.com/plotly/plotly.js/issues/2423


chriddyp:
also upgrading plotly.js to 1.34.0: https://github.com/plotly/plotly.js/releases

PR #177, opened 2018-03-26
add check that dash was correctly imported and exit with helpful message if failed

ned2:
A common gotcha that people new to Dash encounter is naming their script 'dash.py' which then shadows the Dash package itself, causing unhappy times. This adds an explicit check that the `dash.development` attribute exists, which if it doesn't, it's likely this situation has happened, so a more helpful error message is printed to standard error, and the script exits. 

This catches at least some of the times that this error occurs. If we also change the import order for example code from the Dash User guide to make `from dash.dependencies import Input, State, Output` come after after the import of `dash_core_components` this will then catch a lot more.

### DIFF
--- a/dash_core_components/__init__.py
+++ b/dash_core_components/__init__.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+from __future__ import print_function as _
 
 import os as _os
 import sys as _sys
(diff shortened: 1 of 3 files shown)


ned2:
This mirrors the changes in both plotly/dash-html-components#39 and plotly/dash-components-archetype#42

chriddyp:
perfect 👌

Issue #171, opened 2018-03-04
Duplicated event handlers in Graph


mjkramer:
In attempting to create an offshoot of Graph with custom click handling, I found that the JS event handlers are being re-appended every time the Graph is updated by a Dash callback. Here's a minimal example:

```python
import dash
from dash.dependencies import Input, Output
import dash_core_components as dcc
import plotly.graph_objs as go

app = dash.Dash('')

app.layout = html.Div([
    dcc.Input(id='input'),
    dcc.Graph(id='graph')
])

@app.callback(Output('graph', 'figure'),
              [Input('input', 'value')])
def update(value):
    return go.Figure(data=[go.Scatter(x=[1,2,3], y=[2, 1, 3])])

if __name__ == '__main__':
    app.run_server()
```

If you log the ```plotly_click``` handler in Graph.react.js, you'll see what I mean. Without the Dash callback, the handler runs once every time you click on a point. Add the callback, and now the handler runs twice, initially. Then type away in the input, and twice becomes thrice. And so on.

Maybe I'm abusing the framework by playing with the plotly_click handler directly, but even so, this behavior seems mildly wasteful at best. And for adding customized click handling, it's a nightmare. I don't know enough React to prescribe a solution, but my guess is the that the problem stems from the that ```plot``` always queues up ```bindEvents``` to run after ```Plotly.react``` completes .

mjkramer:
Maybe I'm missing some subtlety, but this naive solution seems to work:

```diff
diff --git a/src/components/Graph.react.js b/src/components/Graph.react.js
index cc1800a..5deb8de 100644
--- a/src/components/Graph.react.js
+++ b/src/components/Graph.react.js
@@ -78,8 +78,10 @@ export default class PlotlyGraph extends Component {
             return Plotly.animate(id, figure, animation_options);
         } else {
             return  Plotly.react(id, figure.data, figure.layout, config).then(() => {
-                this.bindEvents(props);
-                this.setState({hasPlotted: true});
+                if (! this.state.hasPlotted) {
+                    this.bindEvents(props);
+                    this.setState({hasPlotted: true});
+                }
             });
         }
     }
```

mjkramer:
OK, previous "solution" is no bueno because it prevents the handlers from seeing the updated ```props```. Now I see why ```bindEvents``` was being called repeatedly. This seems to work better:

```diff
diff --git a/src/components/Graph.react.js b/src/components/Graph.react.js
[...]

PR #178, opened 2018-03-27
Fix graph issues

chriddyp:
- should fix https://github.com/plotly/dash-docs/issues/81
- also adds a resize after first plot in case graphs aren’t sized up in their container

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,4 +6,4 @@ venv/
 *.pyc
 *.log
 .DS_Store
-
+dist/
--- a/src/components/Graph.react.js
+++ b/src/components/Graph.react.js
@@ -78,7 +78,7 @@ export default class PlotlyGraph extends Component {
         } else {
             return Plotly.react(id, figure.data, figure.layout, config).then(() => {
                 if (!this._hasPlotted) {
-                    this.bindEvents(props);
+                    this.bindEvents();
                     Plotly.Plots.resize(document.getElementById(id));
                     this._hasPlotted = true;
                 }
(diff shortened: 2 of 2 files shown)


chriddyp:
@mjkramer - If you're still around, would you mind elaborating on this comment? https://github.com/plotly/dash-core-components/issues/171#issuecomment-370213691 I'm essentially patching this with that suggestion but I'm forgetting what you meant by
> OK, previous "solution" is no bueno because it prevents the handlers from seeing the updated props. Now I see why bindEvents was being called repeatedly.

chriddyp:
@plotly/dash or @plotly/frontend - could someone review?

chriddyp:
@charleyferrari - could you check if this fixes the resizing issue that you saw in your latest dashboard project? try out the prerelease with
```
pip install dash-core-components==0.21.1rc2
```

mjkramer:
Hi Chris. When `bindEvents` is only called once, don't those callbacks end up always seeing the `props` as they were at the time that `bindEvents` was called? Or is it irrelevant (i.e. do `clear_on_hover`, `setProps`, and `fireEvent` never change)? Why not, in addition to the `if (!this._hasPlotted)` fix, get rid of the `props` argument to `bindEvents`, and just let the callbacks access `this.props` directly? I do this in my components, and while I can't claim to fully understand what's going on under the hood, I haven't seen any issues.

chriddyp:
thanks @mjkramer ! agreed, `this.props` should suffice here.

chriddyp:
Fixes the issues in the docs https://github.com/plotly/dash-docs/issues/81
image

nicolaskruchten:
The resize thing looks good to me!

chriddyp:
Ping @charleyferrari re https://github.com/plotly/dash-core-components/pull/178#issuecomment-376387427

charleyferrari:
@chriddyp working for me with Halliburton app:

hburt3

PR #184, opened 2018-04-09
fixes for finance charts

chriddyp:
rebased from https://github.com/plotly/dash-core-components/pull/183


chriddyp:
Fixes plotly/dash#241, stopgap until plotly/plotly.js#2510 is fixed

PR #185, opened 2018-04-12
Add missing input disabled attr

bpostlethwaite:
@chriddyp 

fixes https://github.com/plotly/dash/issues/51

Also fixes source maps by using inline source maps. This means Dash will include the source maps in the main bundle which is already going to be served in the html. I can exclude this commit if you want (it worked for me here and where I use this in dash-renderer).

Now we have:
gifit_2018-04-12_180504

being generated from this code:

```python
import dash

app = dash.Dash()
app.layout = html.Div(
    [
        dcc.Input(
            type='int',
            disabled=True,
            placeholder='I am disabled'
        )
    ]
)
if __name__ == '__main__':
    app.run_server()
```


chriddyp:
💃 

Might be good to add an integration test on the `disabled` behaviour too. Something like `add_keys` shouldn't add any characters or something. Selenium integration tests are here: https://github.com/plotly/dash-core-components/blob/master/test/test_integration.py

bpostlethwaite:
I'll add a test thanks!

bpostlethwaite:
@chriddyp I added some test server configuration (documented) and a test. Also a quick fix for some broken error handling behaviour in the TestIntegration class. 

Can you take a quick look - if you are happy with these types of changes I can make similar changes while I am in other Dash repos.

bpostlethwaite:
I recommend stepping through commit by commit (there are only a few) because there is an autopep8 at the end

CONTRIBUTING.md
```diff
@@ -17,3 +17,16 @@ like this:
...
 export TOX_PYTHON_27=~/.pyenv/versions/2.7.14/bin/python
 ```
+
+## Local configuration
+You can configure the test server with the following variables:
+### DASH_TEST_CHROMEPATH
+If you run a special chrome set the path to your chrome binary with this environment variable.
+
+### DASH_TEST_PROCESSES
+If you encounter errors about Multi-server + Multi-processing when running under Python 3 try running the tests with the number of server processes set to 1.
```
chriddyp: A couple of tests in the `dash-renderer` repo depend on there being multiple processes, in particular: https://github.com/plotly/dash-renderer/blob/c0f886d2a6d2507c22527a86be614f3c2838b667/tests/test_render.py#L1360-L1366

In those tests, one callback is supposed to hang while another process serves the other callback. I don't think that any tests in this repo absolutely require multiple processes.

So, we'll just have to be cautious about adding it to `dash-renderer` and we'll need to make sure that CI still runs things with multiple processes

test/test_integration.py
```diff
@@ -387,7 +387,8 @@ def test_location_link(self):
                 id='test-link-search',
                 href='?testQuery=testValue',
                 refresh=False),
-            html.Button('I am a magic button that updates pathname', id='test-button'),
```
chriddyp: We should add `flake8` and `pylint` to the tests, here's how we do it in the `dash` repo:
https://github.com/plotly/dash/blob/25c97d1b3efd99bb7e3f0aa807abb7ebc2712a6b/tox.ini#L17-L18

chriddyp:
Looks good! Let's do a release while we're at it and make sure that you have the appropriate rights. You can make a separate PR or do it in this one. The process:
- Bump `version.py` and `version` in `package.json`
- Edit `CHANGELOG.md`
- `npm run prepublish`
- `git add .`, `git commit -m vx.y.z`, `git push`
- `npm run publish-all`
- `git tag -a 'vx.y.z` -m 'vx.y.z'`
- `git push origin master --follow-tags`
- Bump version in `dash-docs` and check image tests: https://github.com/plotly/dash-docs/blob/9acdd179fb1f51c8000bff45bf2f31c268705523/requirements.txt#L8

Issue #133, opened 2017-12-22
Change `values` to `value` in Checklist?


ned2:
Currently, not providing an initial value for the `values` attribute of dcc.Checklist results in the front end displaying "Error loading layout" and an obscure Exception on the backend which does not shed much light on the source of the problem (included at end). In order to have no checkboxes selected (and avoid this issue), you must provide an initial value of `values=[]`, but I think it's a reasonable assumption that the default value would have this behavior and so could be omitted. 

Relevant post on the forum: https://community.plot.ly/t/filling-a-checklist-through-a-callback/7435

The current exception:
```
Traceback (most recent call last):
  File "/home/ned/.pyenv/versions/3.6.3/envs/app/lib/python3.6/site-packages/flask/app.py", l
ine 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/ned/.pyenv/versions/3.6.3/envs/app/lib/python3.6/site-packages/flask/app.py", l
ine 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/home/ned/.pyenv/versions/3.6.3/envs/app/lib/python3.6/site-packages/flask/app.py", l
ine 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/ned/.pyenv/versions/3.6.3/envs/app/lib/python3.6/site-packages/flask/_compat.py
", line 33, in reraise
    raise value
  File "/home/ned/.pyenv/versions/3.6.3/envs/app/lib/python3.6/site-packages/flask/app.py", l
ine 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/ned/.pyenv/versions/3.6.3/envs/app/lib/python3.6/site-packages/flask/app.py", l
ine 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/ned/.pyenv/versions/3.6.3/envs/app/lib/python3.6/site-packages/flask/app.py", l
ine 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/ned/.pyenv/versions/3.6.3/envs/app/lib/python3.6/site-packages/flask/_compat.py
", line 33, in reraise
    raise value
  File "/home/ned/.pyenv/versions/3.6.3/envs/app/lib/python3.6/site-packages/flask/app.py", l
ine 1612, in full_dispatch_request
    rv = self.dispatch_request()
[...]

PR #191, opened 2018-04-25
Added values as defaultProps in checklist

shreyasbapat:
Solves plotly/dash#239


chriddyp:
This looks good! A few small things to take this across the finish line 🏁 
1. Create a screenshot test. 
- Create a new function in `test_integration`: `test_checklist_empty_values`
- Add a `self.snapshot('initial rendering of checklist')` - This'll take a screenshot of the app using the percy.io screenshot service
- Click on the checklist item programatically with a CSS selector. This'll be something like:
```
self.wait_for_element_by_css_selector('#checklist-container input')[0].click()
```
- Verify that the appropriate callback was fired with the appropriate value. To check that a callback was fired.
- Altogether, this test will look _something_ like:
```
from multiprocessing import Value

class Tests(IntegrationTests):
    # ... the other functions in the class

    def test_checklist(self):
        app = dash.Dash()

        app.layout = html.Div([
            dcc.Checklist(
                id='checklist',
                options=[
                    {'label': 'option 1', 'value': 'option-1'},
                    {'label': 'option 2', 'value': 'option-2'},
                    {'label': 'option 3', 'value': 'option-3'},
                ]
            ),
            html.Div(id='output')
        ])
        
        call_count = Value('i', 0)

                      [Input('checklist', 'values')])
        def update_output(value):
            call_count.value = call_count.value + 1
            return str(value)

        self.startServer(app)

        # check initial text
        wait_for_text_to_equal('output', 'None')

        self.snapshot('initial rendering of checklist')

        # click on the value
        self.wait_for_element_by_css_selector(
            '#checklist-container input')[0].click()

        # check the output text
        wait_for_text_to_equal('output', '[option-1]')

        # check that the callback was fired twice
        self.assertEqual(call_count.value, 2)

        self.snapshot('second rendering of checklist')
```
3. Bump the version in `version.py` and the `version` field in `package.json` and add an entry in the `CHANGELOG.md`. This is technically a bug fix, so we'll go from `0.22.1` to `0.22.2` (see https://semver.org/)
4. Tag me for another review!

Let us know if you have the time to make this changes. If not, perhaps another community member can take it up. And thanks again for your contribution and your time!

shreyasbapat:
Yeah, I will make these changes tomorrow only! Thanks!

shreyasbapat:
@chriddyp  Can you guide me a bit why this is happening? https://circleci.com/gh/plotly/dash-core-components/595?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

bpostlethwaite:
@shreyasbapat I can take a look at this PR and see if I can get those tests running for you. I'll update this thread once Percy is generating snapshots we can look at.

shreyasbapat:
@bpostlethwaite  Actually, I am grateful that you said it, but I want to know the cause so that I fix it. As I am not very much aware with the codebase, If you just guide me a bit, I think I would be able to do it.

bpostlethwaite:
Ok great, here are a couple clues from my investigation yesterday

- dcc.Checkbox works fine when you supply a values field
- dcc.Checkbox breaks when you add the values field which seems to break JSON serialization. Trying to investigate it in the ipdb console leads to recursive errors.
- My next step was to inspect the generated code for Checkbox and step through the case where it no longer has a values field

Happy hunting!

chriddyp:
Ugh, this sounds familiar: https://github.com/plotly/dash-core-components/issues/133#issuecomment-353641641

bpostlethwaite:
Yep that is almost definitely the issue.

shreyasbapat:
Thanks a lot @bpostlethwaite  ! :smile:  If I am unable to solve the problem in a day or two, I will ask your help!

shreyasbapat:
Couldn't figure it out @bpostlethwaite ! Sorry for wasting time.

bpostlethwaite:
Well given https://github.com/plotly/dash-core-components/issues/133#issuecomment-353641641 it seems like these are our choices:
A) rename "values" to some other prop name 
B) start messing around with the Base implementation of the Component
C) Disband PR and continue with a mandatory "values" prop

@chriddyp have you investigated any hijinx like whitelisting values in the Collections.MutableMapping implementation? We could sit another class on top that would do keyname delegation or monkey patch MutableMapping. 

`values` (and for that matter `keys`) are not exactly obscure prop names. Maybe I'll spend a little time on this today.

bpostlethwaite:
It seems like `items`, `pop`, `popitem`, `clear`, `update`, and `setdefault` may be other reserved method/attribute names.

shreyasbapat:
What can be the way to solve this?

valentijnnieman:
I'm going to close this PR for now as it doesn't seem to be moving forward - feel free to reopen if you are still working on this!

PR #192, opened 2018-04-26
rename dashFireEvent to fireEvent

cpsievert:
Based on this code, it looks like the `dashFireEvent` props should be named `fireEvent`

https://github.com/plotly/dash-core-components/blob/9cafeee28f2d1b02bbb55a78c8e25030695c7793/src/components/Graph.react.js#L109-L146

Although, I vaguely recall @chriddyp mentioning at some point that [events](https://github.com/plotly/dash/blob/3dfa941fbba79efafc397360520f5a8e964236f6/dash/dependencies.py#L22-L26) may be deprecated at some point...are you still thinking that will be the case?

### DIFF
--- a/dash_core_components/metadata.json
+++ b/dash_core_components/metadata.json
@@ -1253,7 +1253,7 @@
         "required": false,
         "description": "Function that updates the state tree."
       },
-      "dashFireEvent": {
+      "fireEvent": {
         "type": {
           "name": "func"
         },
--- a/src/components/Graph.react.js
+++ b/src/components/Graph.react.js
@@ -462,7 +462,7 @@ PlotlyGraph.propTypes = {
     /**
      * Function that fires events
      */
-    dashFireEvent: PropTypes.func
+    fireEvent: PropTypes.func
 }
 
 PlotlyGraph.defaultProps = {


cpsievert:
Somewhat unrelated, but it seems like this code doesn't do anything (what is `this.eventEmitter`)?

https://github.com/plotly/dash-core-components/blob/9cafeee28f2d1b02bbb55a78c8e25030695c7793/src/components/Graph.react.js#L157-L161

chriddyp:
> Although, I vaguely recall @chriddyp mentioning at some point that events may be deprecated at some point...are you still thinking that will be the case?

Yes, I think so. I've removed all mentions of events in the docs and all events have an associated property that can be used as a `Input` in a callback

chriddyp:
> Somewhat unrelated, but it seems like this code doesn't do anything (what is this.eventEmitter)?

Looking at `git blame`, I see 
https://github.com/plotly/dash-core-components/commit/7145cb863e1a37cf2a4109c0dae40b9074f3bbbe#diff-300f42217511a927d7aba9885b4a9c3cL35

The purpose of that line was to clear event handlers on unmount. That should probably still happen and that code isn't doing anything right now.

PR #193, opened 2018-04-26
add clear_on_doubleclick prop for clearing clickData

cpsievert:


src/components/Graph.react.js
```diff
@@ -212,6 +219,12 @@ PlotlyGraph.propTypes = {
      */
     clickData: PropTypes.object,
 
+    /**
+     * If True, `clear_on_doubleclick` will clear the `clickData` property
+     * when the user double clicks (when dragmode='zoom').
```
chriddyp: Only on `dragmode='zoom'`? Also, could we be more specific about `dragmode`, like `figure['layout']['dragmode'] = 'zoom'` or `config['dragmode'] = 'zoom'`

chriddyp:
Seems reasonable and a natural extension of the existing API. I would imagine that `clear_on_click` would be more common / easier to use, why the `doubleclick`?

Ultimately I really want the plotly.js to make this natural behaviour by adding a `click` mode to `select` so that we don't need to program in hacks like this and so that the behaviour is consistent across all contexts (cc @alexcjohnson ), this is being tracked in https://github.com/plotly/plotly.js/issues/1852

cpsievert:
> I would imagine that `clear_on_click` would be more common / easier to use, why the `doubleclick`?

By `clear_on_click`, I assume you mean *any* click event on the plot? It seems as though that would make it too easy to clear a selection, which would be especially problematic if you want selections to persist. In plotly's shiny integration, I've actually been ['clearing click data' on any double-click event (regardless of dragmode)](https://github.com/ropensci/plotly/blob/38f447f6e4e20ae376420305f743f4997d2a9a17/inst/htmlwidgets/plotly.js#L293-L300) for quite some time now and haven't really run into issues with it. Note that d00621f clears `clickData` on `"plotly_deselect"` events as well now which is consistent with that shiny behavior (and should've been included in the initial PR)

alexcjohnson:
@chriddyp @cpsievert can we drop this PR now that plotly.js has built-in click-to-select https://github.com/plotly/plotly.js/pull/2944?

chriddyp:
definitely

PR #194, opened 2018-04-29
Added MathJax CDN to enable Latex

kaushiksk:
Solves plotly/dash#242

### DIFF
--- a/dash_core_components/__init__.py
+++ b/dash_core_components/__init__.py
@@ -22,6 +22,11 @@ _components = _dash.development.component_loader.load_components(
 _this_module = _sys.modules[__name__]
 
 _js_dist = [
+	{
+        'external_url': 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-MML-AM_CHTML',
+        'relative_package_path': 'mathjax-2.7.4.js',
+        'namespace': 'dash_core_components'
+    },
     {
         'external_url': 'https://cdn.plot.ly/plotly-1.35.2.min.js',
         'relative_package_path': 'plotly-1.35.2.min.js',


chriddyp:
Thanks @kaushiksk ! We'll need a couple of extra things to get this across the finish line 🏁 
1. We'll need a copy of this library for users that are using `serve_locally=True`:
- Download the contents of this file into the `dash_core_components` folder. Include the version number in the title, e.g. `mathjax-2.7.4.js` (note that `relative_package_path` is referring to this filename)
- Add the file to the `MANIFEST.in` file so that it is included in the distribution: https://github.com/plotly/dash-core-components/blob/master/MANIFEST.in
2. Create a screenshot test. 
- Create a new function in `test_integration`: `test_latex`
- Include a sample graph with LaTeX in it
- `self.wait_for_element_by_css_selector` on that graph and then add a `time.sleep(3)` after since MathJax rendering is async
- Add a `self.snapshot('latex rendering in graph')` - This'll take a screenshot of the app using the percy.io screenshot service
3. Bump the version in `version.py` and the `version` field in `package.json` and add an entry in the `CHANGELOG.md`. This is technically a bug fix, so we'll go from `0.22.1` to `0.22.2` (see https://semver.org/)
4. Tag me for another review!

Let me know if you need help with this or if you don't have the time to do those steps.

And thanks again for your contribution to Dash! 🙌

chriddyp:
Thank you @kaushiksk ! Looking at the screenshot test, it looks like the LaTeX isn't getting rendered :(

Now, this could be an issue with percy.io not rendering the MathJax correctly or perhaps we aren't waiting long enough for the MathJax to render.

Could you share a screenshot of this test running locally and confirm that it is rendering appropriately?

dash_core_components/__init__.py
```diff
@@ -22,6 +22,11 @@
 _this_module = _sys.modules[__name__]
 
 _js_dist = [
+	{
+        'external_url': 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-MML-AM_CHTML',
+        'relative_package_path': 'mathjax-2.7.4.js',
```
chriddyp: I believe this should be `mathjax-2.7.4.min.js`

test/test_integration.py
```diff
@@ -496,3 +496,27 @@ def update_graph(n_clicks):
...
+                    'layout': {
+                        'title': 'Graph of \(y = \frac{x^2 - 2}{4}\)'
+                    }
+                }
+            ),
+        ])
+        self.startServer(app=app)
+
+        graph = self.wait_for_element_by_css_selector('#graph')
+        time.sleep(2)
```
chriddyp: In case this was a timing issue, let's try bumping this up to 4

test/test_integration.py
```diff
@@ -496,3 +496,27 @@ def update_graph(n_clicks):
...
+                    }],
+                    'layout': {
+                        'title': 'Graph of \(y = \frac{x^2 - 2}{4}\)'
+                    }
+                }
+            ),
+        ])
+        self.startServer(app=app)
+
+        graph = self.wait_for_element_by_css_selector('#graph')
```
chriddyp: Here's an idea - let's "wait for" the actual graph to be rendered rather than just the container. Let's update this CSS selector to be:
```
'#graph .svg-container'
```

dash_core_components/__init__.py
chriddyp: Or vice-versa, everything could be labeled `mathjax-2.7.4.js` rather than `mathjax-2.7.4.min.js`. I don't really have a preference, but they do need to all have the same name 😄

chriddyp:
Here's the screenshot on percy.io
image

test/test_integration.py
```diff
@@ -496,3 +496,27 @@ def update_graph(n_clicks):
...
+            html.Label('Graph'),
+            dcc.Graph(
+                id='graph',
+                figure={
+                    'data': [{
+                        'x': [1, 2, 3, 4, 5],
+                        'y': [1, 3, 5, 9, 13]
+                    }],
+                    'layout': {
+                        'title': 'Graph of \(y = \frac{x^2 - 2}{4}\)'
```
chriddyp: Ah I see, this isn't actually valid in plotly.js. Text needs to be _all_ latex or _none_ latex. 

Let's use this example from the docs: https://plot.ly/python/LaTeX/

```
'title': '$\\sqrt{(n_\\text{c}(t|{T_\\text{early}}))}$'
```

dash_core_components/__init__.py
kaushiksk: Uh my bad. I'll change it in MANIFEST.in

kaushiksk:
@chriddyp Here's how the [LateX example from the docs](https://plot.ly/python/LaTeX/)  looks like in my Firefox 59.0.2 on Ubuntu 16.04.

screenshot from 2018-05-02 00-15-19
Notice how none of the Latex is rendered.

I ran the following snippet locally and noticed similar behavior. 
```python
import dash

app = dash.Dash(__name__)
app.layout = html.Div([

    html.Label('Graph'),
    dcc.Graph(
        id='graph',
        figure={
            'data': [{
                'x': [1, 2, 3, 4, 5],
                'y': [1, 3, 5, 9, 13]
            }],
            'layout': {
                'title': '$\\sqrt{(n_\\text{c}(t|{T_\\text{early}}))}$'
            }
        }
    ),
])

if __name__ == "__main__":
    app.run_server(debug=True)
```


Here's the ouput for the above code:
screenshot from 2018-05-02 00-21-52

Now if I change the `title` to plaintext like `"My Title"`, it renders the plaintext title.

screenshot from 2018-05-02 00-26-23
 
I could see a script tag in the html pointing to the MathJax CDN, so MathJax is getting loaded.

bpostlethwaite:
This is what Percy.io sees 
image

I'll see if I can take this PR out for a spin later today. I have some experience with Mathjax

chriddyp:
> @chriddyp Here's how the LateX example from the docs looks like in my Firefox 59.0.2 on Ubuntu 16.04.

I think this is just a firefox issue

test/test_integration.py
```diff
@@ -496,3 +496,27 @@ def update_graph(n_clicks):
...
+            html.Label('Graph'),
+            dcc.Graph(
+                id='graph',
+                figure={
+                    'data': [{
+                        'x': [1, 2, 3, 4, 5],
+                        'y': [1, 3, 5, 9, 13]
+                    }],
+                    'layout': {
+                        'title': 'sqrt{(n_\\text{c}(t|{T_\\text{early}}))}$'
```
chriddyp: we're missing a leading `$` here, I believe it should be 
```
$sqrt{(n_\\text{c}(t|{T_\\text{early}}))}$
```

mkhorton:
Is this intended just for graph labels, or would equations render inside a dcc.Markdown component too?

chriddyp:
> Is this intended just for graph labels, or would equations render inside a dcc.Markdown component too?

As far as I know, just graph labels. I'm haven't looked into what it would take for MathJAX to just automatically work _everywhere_. If anyone knows or would like to do that research, please update the thread :)

mkhorton:
As I understand it, MathJax by default will render any equation strings it finds anywhere in the page--but I wouldn't be surprised if this has been disabled somehow just to be polite? I'm really note sure.

There are React components out there which will do MathJax rendering for you, but it seems a shame to write another Dash component using these and possibly be loading the MathJax library twice, if MathJax is already a dependency for the core components.

chriddyp:
> As I understand it, MathJax by default will render any equation strings it finds anywhere in the page--but I wouldn't be surprised if this has been disabled somehow just to be polite? I'm really not sure.

Yeah, I think that one issue is that Dash renders everything dynamically, so I'm assuming that we'd to inform MathJax whenever new changes were made.

chriddyp:
> There are React components out there which will do MathJax rendering for you, but it seems a shame to write another Dash component using these and possibly be loading the MathJax library twice, if MathJax is already a dependency for the core components.

It might be nice if there was some kind of "`MathJaxProvider` component that could be placed at the top of the app and used:
```
app.layout = MathJaxProvider([
    html.Div(...)
])
```

test/test_integration.py
kaushiksk: I must've missed that while testing with different inputs. Fixed this. Check perci.io output now, hopefully, this should work.

chriddyp:
image

Well, it's looking better but we still don't see the actual mathtype. This could just be an artifact of percy. Could you share a screenshot from your own environment to confirm that this works?

kaushiksk:
>Could you share a screenshot from your own environment to confirm that this works?

@chriddyp I've shared screenshots from my environment in the comments above.

chriddyp:
> I've shared screenshots from my environment in the comments above.

It looks like the screenshots that you shared in https://github.com/plotly/dash-core-components/pull/194#issuecomment-385757627 were just in Firefox, and that MathJax has issues rendering in Firefox. Could you share screenshots from another browser that doesn't have those issues, like Chrome?

Issue #115, opened 2017-11-23
datepicker doesn't open if date values are `None`


chriddyp:
Originally reported in https://community.plot.ly/t/datepickerrange-clearable-in-tabs-not-working-as-expected/6850/4

```python
import dash
from dash.dependencies import Input, Output
import dash_core_components as dcc
import dash_html_components as html
from datetime import datetime as dtime

app = dash.Dash()


app.config['suppress_callback_exceptions'] = True
app.layout = html.Div([
    dcc.DatePickerRange(
        id='Dr1',
        clearable=True,
        reopen_calendar_on_clear=True,
        start_date_placeholder_text='Select a date',
        end_date=dtime.today().strftime("%Y-%m-%d")
    ),
    html.Div(id='output')
])


@app.callback(
    Output('output', 'children'),
    [Input('Dr1', 'start_date'),
     Input('Dr1', 'end_date')])
def dates_selected(start_date, end_date):
    value = "From- %s  To-   %s" % (start_date, end_date)
    return value


if __name__ == '__main__':
    app.run_server(debug=True)
```

Quite a few error messages in the console
![image](https://user-images.githubusercontent.com/1280389/33187985-ec8ca686-d063-11e7-9ce2-e4cfae76a676.png)

xblad:
@chriddyp
Hi Chris, I think the issue above is related to the fact, that `DatePickerRange`'s attribute `initial_visible_month` isn't defined. Maybe the default value could be set (e.g. `datetime.today()`) and it would solve the issue (AirBnB's date-picker shows current month by default as well).

90debasish:
Hi Chris,

in the above example what i am facing issue is :
1. In the beginning when the date picker starts it prints as None- This is completley ok
2. However when you select some date and then let say clear those dates then it still shows the previous value but it should show as None since the dates are clear.

The problem with the initial approach is:
1. Let say we selected some date from and end; we get the data from those period
2. However let say if we want only data which is less than the end period after previous selection.-------> here we are not able to do so since we still have from value which is not cleared.

byronz:
issue has been fixed in 
```
dash                 1.0.2
dash-core-components 1.0.0
dash-html-components 1.0.0
dash-renderer        1.0.0
dash-table           4.0.2
```

PR #201, opened 2018-05-16
Add current date's initial visible month

xblad:
Added current date as the last resort for the initialVisibleMonth property of the component in DatePickerRange.react.js and in DatePickerSingle.react.js (issue #115)


chriddyp:
Thanks! This looks good to me. A few other things would be helpful:
- Would you mind adding a test? We can just render the component and then take a screenshot of the new behaviour. 
- Would you mind updating the `CHANGELOG.md` and the version numbers in `version.py` and `package.json`?

xblad:
@chriddyp, I have added the tests and updated version number (_v0.22.3_ + _tag_).
Nevertheless, it is my first contribution on Github, so quick revision of the changes is highly appreciated 😃 
Btw, I see, that **percy/dash-core-components** test was failed, but I don't have an access to the resource, so I'm not able to check why...

valentijnnieman:
@chriddyp This seems solid - can we update & merge this? 💃

mikesmith1611:
It looks like change has been implemented:
https://dash.plot.ly/dash-core-components/datepickersingle
can confirm it works

PR #211, opened 2018-06-15
Confirmation modal

T4rk1n:
> Update July 9, 2018
> This component is almost ready! We're still interested in hearing from the community about this component. You can try out this component by installing the "prerelease" version:
> `pip install dash-core-components==0.25.0rc2`
> Feedback welcome 🙌 

Simple example:
```python
import dash

app = dash.Dash()

app.layout = html.Div([
    dcc.ConfirmDialogProvider(
        children=html.Button(
            'Click Me',
        ),
        id='danger-danger',
        message='Danger danger! Are you sure you want to continue?'
    ),
    html.Div(id='output')
])


              [Input('danger-danger', 'submit_n_clicks')])
def update_output(submit_n_clicks):
    if not submit_n_clicks:
        return ''
    return """
        It was dangerous but we did it!
        Submitted {} times
    """.format(submit_n_clicks)

if __name__ == '__main__':
    app.run_server(debug=True)

```

confirm-dialog

Another example with callback controlled confirm dialog:
```python
import dash

app = dash.Dash()

app.layout = html.Div([
    dcc.ConfirmDialog(
        id='confirm',
        message='Danger danger! Are you sure you want to continue?',
    ),

    dcc.Dropdown(
        options=[
            {'label': i, 'value': i}
            for i in ['Safe', 'Danger!!']
        ],
        id='dropdown'
    ),
    html.Div(id='output')
])


@app.callback(Output('confirm', 'displayed'),
              [Input('dropdown', 'value')])
def display_confirm(value):
    if value == 'Danger!!':
        return True
    return False


              [Input('confirm', 'submit_n_clicks')])
def update_output(submit_n_clicks):
    if submit_n_clicks:
        return 'It wasnt easy but we did it {}'.format(submit_n_clicks)


if __name__ == '__main__':
    app.run_server(debug=True)
```
confirm-dialog-dropdown


T4rk1n:
Here's a screenshot of the Confirm modal, percy couldn't take a snapshot of it.
comfirmsnap

chriddyp:
Looking good! I'd like to submit this to the community for review. Could you post a couple of representative code examples in this PR?

T4rk1n:
Here's an example of dash app where you can add cities to the graph. If you push the reset button, it will ask to confirm before removing the user cities from the graph.

```python
import json
import dash

app = dash.Dash()

data = [
    {
        'x': [32],
        'y': [48],
        'name': 'Washington',
        'mode': 'markers',
        'marker': {'size': 12}
    },
    {
        'x': [12],
        'y': [9],
        'name': 'New York',
        'mode': 'markers',
        'marker': {'size': 12}
    }
]

container_style = {'display':'flex', 'flex-direction': 'column', 'flex-grow': 1}

app.layout = html.Div([
    html.Div(id='controls', style=container_style,  children=[
        dcc.Input(id='city-name', placeholder='City name'),
        dcc.Input(id='x', placeholder='x'),
        dcc.Input(id='y', placeholder='y'),
        html.Button(id='add-button', children='add'),
        html.Button(id='reset-button', children='Reset', n_clicks=0),
    ]),
    html.Div([
        dcc.Confirm(id='confirm', message='Warning, you will lose all custom input.'),
        dcc.Graph(id='graph', style={'flex-grow': 3}),
        html.Div(id='values', style={'display': 'None'})
    ]),
], style={'display': 'flex', 'justify-items': 'stretch', 'align-items': 'stretch'})


@app.callback(Output('values', 'children'),
              [Input('add-button', 'n_clicks'), Input('confirm', 'result')],
              [State('city-name', 'value'), State('x', 'value'), State('y', 'value'), State('values', 'children')])
def on_add_city(n_clicks, confirm_result, city_name, x, y, values):
    if confirm_result:
        return ''
    if n_clicks > 0:
        graph_values = json.loads(values) if values else list(data)
        graph_values.append({
            'x': [int(x)],
            'y': [int(y)],
            'name': city_name,
            'mode': 'markers',
            'marker': {'size': 12}
        })
        return json.dumps(graph_values)


@app.callback(Output('confirm', 'init'), [Input('reset-button', 'n_clicks')])
def on_click_confirm(n_clicks):
    if n_clicks > 0:
        return {'ask': True, 'value': 'confirmed'}
    return


@app.callback(Output('graph', 'figure'), [Input('values', 'children')])
def on_change_data(values):
    if values:
        v = json.loads(values)
        return {'data': v}
    return {'data': data}


if __name__ == '__main__':
    app.run_server(debug=True)

```

nicolaskruchten:
I'm concerned about these `n_clicks` checks against `0`... Seems like the confirmation would be one-time-use-only in this case, no? How would we support confirmation every time?

chriddyp:
> I'm concerned about these `n_clicks` checks against `0`

These checks against are `0` are just for Dash's warm-up initialization, when it fires all of the callbacks on page load.

test/test_integration.py
```diff
@@ -528,3 +528,48 @@ def update_graph(n_clicks):
...
+                return 'canceled'
+
+        self.startServer(app)
+        button = self.wait_for_element_by_css_selector('#button')
+        self.snapshot('confirmation -> initial')
+        time.sleep(1)
+        button.click()
+        time.sleep(1)
+
+        self.driver.switch_to.alert.accept()
```
chriddyp: 🎉  very nice!

src/components/ConfirmDialog.react.js
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Message to show in the popup.
+     */
+    message: PropTypes.string,
+
+    /**
+     * Number of times the modal was submited or canceled.
+     */
+    n_clicks: PropTypes.number,
```
chriddyp: Given that we have `submit_n_clicks` and `cancel_n_clicks`, is there a use case for the general `n_clicks`?

src/components/ConfirmDialog.react.js
```diff
@@ -0,0 +1,77 @@
...
+     */
+    submit_n_clicks: PropTypes.number,
+    /**
+     * Number of times the popup was canceled.
+     */
+    cancel_n_clicks: PropTypes.number,
+    /**
+     * Set to true to send the popup.
+     */
+    send_confirm: PropTypes.bool,
```
chriddyp: For some reason, `"send"` doesn't feel right to me. What about `display_confirm`? Or maybe just `display`? Or `display_confirm_dialog`? Any other thoughts @plotly/dash ?

src/components/ConfirmDialog.react.js
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Is the modal currently displayed.
+     */
+    displayed: PropTypes.bool,
```
chriddyp: I think that I thought that `displayed` would take the place of `send_confirm`:
```
@app.callback(Output('confirm', 'displayed'), [Input('my-button', 'n_clicks'])
def display_modal(n_clicks):
    return boolean(n_clicks)
```

Does that work? Or am I missing a functional difference between `displayed` and `send_confirm`?

src/components/ConfirmDialog.react.js
```diff
@@ -0,0 +1,77 @@
+import PropTypes from 'prop-types';
+import {Component} from 'react';
+
+/**
+ * ConfirmDialog wraps window.confirm
```
chriddyp: This is the message that will appear in the component's `help(dcc.ConfirmDialog)`, so let's expand this a bit. The Dash users probably don't know what `window.confirm` is referring to. Perhaps something like:
> ConfirmDialog is used to display the browser's native "confirm" modal, with an optional message and two buttons ("OK" and "Cancel"). This `ConfirmDialog` can be used in conjunction with buttons when the user is performing an action that should require an extra step of verification. 

And then similarly for the `ConfirmDialogProvider`. For the `ConfirmDialogProvider`, we should mention that you can pass in a button directly as `children`

src/components/ConfirmDialogProvider.react.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Wrap children onClick to send a confirmation dialog.
+ */
+export default class ConfirmDialogProvider extends React.Component {
+    render() {
+        const { id, setProps, children } = this.props;
+
+        // Will lose the previous onClick of the child
+        const wrapClick = (child) => React.cloneElement(child, {onClick: () =>
```
chriddyp: Nice, this seems like a good solution.

test/test_integration.py
```diff
@@ -528,3 +528,48 @@ def update_graph(n_clicks):
...
+        self.snapshot('confirmation -> confirmed')
+
+        time.sleep(0.2)
+        button.click()
+        time.sleep(1)
+        self.driver.switch_to.alert.dismiss()
+        time.sleep(0.5)
+        self.wait_for_text_to_equal('#confirmed', 'canceled')
+
+        self.snapshot('confirmation -> canceled')
```
chriddyp: Very nice tests! I'd like to see a couple of other things:
- A similar test for the `dcc.ConfirmDialogProvider`
- Let's test that the callback is called an appropriate number of times. In other tests, I use a thread/multiprocess-safe variables to do that, see e.g. the "`call_count`" variables here: https://github.com/plotly/dash-renderer/blob/master/tests/test_render.py#L478-L484

chriddyp:
Looking good @T4rk1n ! I just made a few other comments. Let's also make a prerelease of this package so that some community members can easily take it for a spin. I wrote some instructions on how to publish prereleases in this comment here: https://github.com/plotly/dash-core-components/pull/215#issuecomment-399472460

src/components/ConfirmDialog.react.js
T4rk1n: Send_confirm is for activating the modal, it get sets to false after fire, displayed was for telling if the confirmation was currently showing. I just changed for just displayed and it works the same.

nicolaskruchten: I think a good rule to keep in mind here and with most functional/React style coding is that prop names should not be verbs, especially not imperative ones like `send` but rather descriptions of state like `displayed` :)

test/test_integration.py
T4rk1n: I added a test for the provider and put a call count variable. The call_count get increased normally in my local tests but in circleci it get increased by 6 instead of two, I put the value to be the n_clicks instead for the tests to work on circleci but I think there's an error.

src/components/ConfirmDialogProvider.react.js
```diff
@@ -0,0 +1,84 @@
...
+import PropTypes from 'prop-types';
+
+import ConfirmDialog from './ConfirmDialog.react'
+
+
+
+/**
+ * Wrap children onClick to send a confirmation dialog.
+ * You can add a button directly as a children:
+ * `dcc.ConfirmDialogProvider(html.Button('click me', id='btn'), id='confirm')`
```
chriddyp: Our users probably won't know what `onClick` means, so let's change this to:
> A wrapper component that will display a confirmation dialog when its child component has been clicked on.
> For example:
```
dcc.ConfirmDialogProvider(
    children=html.Button('Click Me'),
    message='Danger - Are you sure you want to continue?',
    id='confirm'
)
```

src/components/ConfirmDialogProvider.react.js
```diff
@@ -0,0 +1,84 @@
...
+     */
+    n_clicks_timestamp: PropTypes.number,
+    /**
+     * Number of times the submit was clicked
+     */
+    submit_n_clicks: PropTypes.number,
+    /**
+     * Number of times the popup was canceled.
+     */
+    cancel_n_clicks: PropTypes.number,
```
chriddyp: Can we add `submit_n_clicks_timestamp` and `cancel_n_clicks_timestamp` here and in the `ConfirmDialog` component?

src/components/ConfirmDialogProvider.react.js
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Message to show in the popup.
+     */
+    message: PropTypes.string,
+
+    /**
+     * Number of times the modal was submited or canceled.
+     */
+    n_clicks: PropTypes.number,
```
chriddyp: I'm still wondering if there is a use case for this property given that we have `submit_n_clicks` and `cancel_n_clicks`. Thoughts @plotly/dash ?

src/components/ConfirmDialogProvider.react.js
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Last timestamp the popup was clicked.
+     */
+    n_clicks_timestamp: PropTypes.number,
+    /**
+     * Number of times the submit was clicked
```
chriddyp: > Number of times the submit button was clicked

src/components/ConfirmDialogProvider.react.js
T4rk1n: If I add the timestamps for submit and cancel, I don't see one as I was only using it to know which one was clicked in the test and it wasn't a real use case. I will remove it.

bpostlethwaite:
@T4rk1n you have the :dancer: to merge this :heavy_check_mark: 

We have interested parties waiting for release. Thanks!

PR #74, opened 2017-09-14
Dash Tabs component

chriddyp:

PR #213, opened 2018-06-21
Improved Tabs component

valentijnnieman:
> Update July 9, 2018
> We are looking to merge and release this component this week. Please give it a try! You can install with:
> `dash_core_components==0.24.0rc2`
> Feedback welcome 🙌 

***

Hello there! 
Here is the improved Tabs component I've been working on, branched off the current master branch. It supports className's and inline style's for the multiple containers (parent, tabs, tab, tab-content, tab--selected) which appends classes to the base classes (let me know if it should overwrite the default classes instead of appending) and has a `vertical` bool for rendering vertically. 
screen shot 2018-06-21 at 11 37 40
screen shot 2018-06-21 at 11 36 40

It's also more responsive:
screen shot 2018-06-21 at 11 38 31

It's basically 2 components: a `dcc.Tabs` component which is the container, where you can specify `vertical` and set some classes and styles, and a `dcc.Tab` component, which renders a tab with a `label`, and renders it's `children` (which can be anything in Dash) when that tab is `selected`! Thanks to @T4rk1n for his work and ideas!

Here's an example of how to use Tabs:

```
import dash

app = dash.Dash()

app.layout = html.Div([
    html.H1('Dash Tabs component demo', style={
            'textAlign': 'center', 'margin': '48px 0', 'fontFamily': 'system-ui'}),
    dcc.Tabs(id="tabs", children=[
        dcc.Tab(label='Tab one', children=[
            html.Div([
                dcc.Graph(
                    id='example-graph',
                    figure={
                        'data': [
                            {'x': [1, 2, 3], 'y': [4, 1, 2],
                                'type': 'bar', 'name': 'SF'},
                            {'x': [1, 2, 3], 'y': [2, 4, 5],
                             'type': 'bar', 'name': u'Montréal'},
                        ],
                        'layout': {
                            'title': 'Dash Data Visualization'
                        }
                    }
                )
            ])
        ]),
        dcc.Tab(label='Tab two', children=[
            html.Div([
                html.H1("This is the content in tab 2"),
                html.P("A graph here would be nice!")
            ])
        ]),
        dcc.Tab(label='Tab three', children=[
            html.Div([
                html.H1("This is the content in tab 3"),
            ])
        ]),
    ],
        style={
        'fontFamily': 'system-ui'
    },
        content_style={
        'borderLeft': '1px solid #d6d6d6',
        'borderRight': '1px solid #d6d6d6',
        'borderBottom': '1px solid #d6d6d6',
        'padding': '44px'
    },
        parent_style={
        'maxWidth': '1000px',
        'margin': '0 auto'
    }
    )
])


if __name__ == '__main__':
    app.run_server(debug=True)
```

~~I haven't included any builds yet, because I'm not sure if I should - CircleCI etc will handle that stuff, right?~~

edit: 
Please be aware that the component API has changed a bit - it now uses 2 components:
- dcc.Tabs() is the container component, which renders it's `children` as tabs, complete with selection logic - you no longer have to set this up yourself in Dash, it's handled internally, which I hope is more easy to use. Would love to hear some feedback on this, especially if this is giving you trouble or prevents a certain use case for you.
- dcc.Tab() is the actual Tab component - it has a `label` prop that holds the label of the tab's button, and renders it's `children` as the content that will be visible when this tab is selected


kyleabeauchamp:
I noticed that the changes to `test_integration.py` from #74 (e.g., https://github.com/plotly/dash-core-components/blob/7b6b5ff9c648793aeb7dd1eea0444dae83e688ce/test/test_integration.py#L219) are not in the current PR.  Do you recall if the old integration tests are still applicable for the new Tab classes?  If so, should we add back in those tests to this PR so that the features are tested?

valentijnnieman:
@kyleabeauchamp Good point - although the old integration tests aren't applicable here because I've changed the component API a bit, some integration testing with the new components would be nice!

kyleabeauchamp:
I have an open PR as well that attempts this, FWIW.

freshwuzhere:
This looks AWESOME!  Thanks

valentijnnieman:
@kyleabeauchamp Whoops, sorry, didn't see your PR in time :)

kyleabeauchamp:
One more naive question: is rebuilding and committing bundle.js part of the PR process?  I suspect that would facilitate testing this branch (for those of us too lazy to get a fully working dev environment).

valentijnnieman:
@kyleabeauchamp Good question - I wasn't sure about this (couldn't find any documentation about it) but I think you're right, looking back at older PR's it looks like the build is included, and I forgot to update the version as well, so I've now done that. Thanks for your help! :)

ZeroCool2u:
This is beautiful. I'm a big fan of the dual component design. Great work!

kyleabeauchamp:
Thanks, with the latest commits I can now import the `dcc.Tabs` component.  This will help folks test this out.

kyleabeauchamp:
I've found a minor issue where I see "Error loading dependencies" whenever I have a `dcc.Tabs` with only a single tab.  I've added a test for this in https://github.com/plotly/dash-core-components/pull/217/files

kyleabeauchamp:
Otherwise, I was able to successfully port my previous Dash project to the latest version of tabs!  It also seems to have cut the number of lines down, in that my interactivity with the tabs is now handled under the hood rather than by callbacks.

chriddyp:
:tada: Congrats on your first PR! :tada:

This is really great stuff. Here are a couple of general comments, one on the API, the other on the the style of the tabs:

***

**API**

> is the container component, which renders it's children as tabs, complete with selection logic - you no longer have to set this up yourself in Dash, it's handled internally, which I hope is more easy to use. Would love to hear some feedback on this, especially if this is giving you trouble or prevents a certain use case for you.

This is very cool, in many cases setting up the boiler-plate tab logic could feel pretty tedious. A few community members have commented on that fact.

Now, one subtle difference between this API and the old API is that the old API prevented the content _in each tab_ from being transported to the browser. If you have 5 tabs, the only the layout of the active tab was transported. In this implementation, the layout of all 5 tabs necessarily has to be transported. This can be problematic if the layout is large (larger network request) or, perhaps more commonly, if there is some time-consuming computation that needs to happen to generate the content for each tab (e.g. if each tab's content has a 2 second computation or network request, then loading 5 tabs takes 10 seconds instead of 2 seconds).

Given that, I believe that we should offer a method to load the content "on-demand" through a callback. What I'm not sure about is whether we should support two APIs for tabs or just one.

One idea that I had a while back was to have an API like this:
```
options: [
    {
        'label': 'Tab 1',
        'value': 'tab-1',
        'children': html.Div(...)
    },
    {
        'label': 'Tab 2',
        'value': 'tab-2',
        'children': html.Div(...)
    },    
]
```

where `children` would be optional. So, if you wanted to lazy-load the `children` through a callback, then you'd just omit them:
```
options: [
    {
        'label': 'Tab 1',
        'value': 'tab-1'
    },
    {
        'label': 'Tab 2',
        'value': 'tab-2'
    },    
]
```

The problem with this UI is that Dash components can _only be the `children` property_. Ultimately, I'd like for us to change that ("Dash Renderer 2.0": https://github.com/plotly/dash-renderer/pull/32) but that probably won't happen for a few more months.

It seems like maybe your API could be extended to allow this case by:
- Making `Tab.children` optional
- Adding a `Tab.value` to each component
- Updating `Tabs.value` with the currently selected `value` of the children.

Would that work?

***

**Style**

- For the active tab, I prefer for the background color to be white and for there to be no border. This way, the tab bleeds into the content and gives a better indication that the tab is related to the content. As a reference on this point, see point 7 in the NN/g tabs guide: https://www.nngroup.com/articles/tabs-used-right/ ("Connect the current tab to the content area")
- I like the look of the mobile stacked tabs and I think it solves the general use case well. Looking at other mobile designs of tabs (https://uxplanet.org/tabs-for-mobile-ux-design-d4cc4d9410d1), it seems like folks solve this issue by using really short names (something that we can't control for our users). Given that, I wonder if we should make the media query style parameter configurable. Or, if we should make the mobile version of tabs an option like 
```
mobile_layout = 'stacked' | 'horizontal'
mobile_breakpoint = 450
```
and then shrink some of the padding on smaller devices.

***

Finally, let's make a prerelease so that our community members can take it for a spin :)

src/components/Tabs.react.js
```diff
@@ -0,0 +1,204 @@
...
+  selectHandler
+}) => {
+  return (
+    <div
+      className={
+        selected
+          ? `${className || ''} ${selectedClassName || ''} tab tab--selected`
+          : `${className || ''} tab`
+      }
+      style={selected ? { ...style, ...selectedStyle } : style}
```
chriddyp: Nice work with the `selectedStyle` and `selectedClassName`

src/components/Tabs.react.js
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Appends (inline) styles to the Tabs container holding the individual Tab components.
+   */
+  style: PropTypes.object,
+
+  /**
+   * Appends (inline) styles to the top-level parent container holding both the Tabs container and the content container.
+   */
+  parentStyle: PropTypes.object,
```
chriddyp: In Python, the community errs towards underscores rather than camelCase. So, I think we should do the same in our properties.

In the beginning of Dash, I wasn't completely consistent. I kept `className` and in `Graph` I used `hoverData` and `clickData`. However, lately, I've been using `_` more. So in this case, `parent_style`.

The weird case would be what to do with `contentClassName`. Since we are using `className` already, do we mix styles with `content_className`?

src/components/Tabs.react.js
```diff
@@ -0,0 +1,204 @@
...
+          .tab {
+            width: auto;
+          }
+        }
+      `}</style>
+    </div>
+  );
+};
+
+export default class Tabs extends Component {
```
chriddyp: Let's add a docstring above this component - these docstrings will get picked up by the Dash component toolchain and will become available in the Python `help(dcc.Tabs)` context

src/components/Tabs.react.js
```diff
@@ -0,0 +1,204 @@
...
+        </div>
+        <div
+          className={`${tabContentClass} ${this.props.contentClassName || ''}`}
+          style={this.props.contentStyle}
+        >
+          {this.props.children[this.state.selected].props.children}
+        </div>
+        <style jsx>{`
+          .tab-parent {
+            display: flex;
```
chriddyp: We try to support IE11 and unfortunately `flex` is only partially supported: https://caniuse.com/#feat=flexbox

We have a browserstack account that you can use to check if this subset of `flex` features are compatible with `IE11`

src/components/Tabs.react.js
valentijnnieman: `className` makes sense to me, because of React's use of `className`. I'm fine with underscores, but I do think `content_className` is a little silly and might be confusing! Hmm, tricky.

chuchuhao:
DatePickerRange would broke on this branch

Harsha-Nori:
Thank you, this looks awesome! Can we expose an "id" parameter for each individual dcc.Tab component? I'd like to conditionally update the layout of a dcc.Tab based on other inputs -- in the previous version, I could return a html.Div that contained the relevant dcc.Graphs, tables, etc. that I needed. It feels like this new Tabs paradigm makes it difficult to do dynamic callbacks, whereas the earlier version was better supported.

Or is there another, better way of doing this that I'm missing?

kyleabeauchamp:
To play devil's advocate, I personally do like the fact that I'm now spending less time with callbacks to get interactivity working across the tabs.  With the old interface, I recall having a lot more "gymnastics" to ensure things are updated appropriately.  So I guess my vote would be to support both modes if possible...

Harsha-Nori:
I think that's totally fair. Maybe having ID as a "hidden" parameter might be the way to go -- not required, but available if needed. 

Of course, I could be missing something, and there might be another supported way to do dynamic callbacks that could be better. Just would like to have some way of doing it for more "advanced"/non-standard applications.

joshtrewin7:
Hi all, is this available as a pre-release? 
How can I install and start using this? 

Thanks in advance

valentijnnieman:
@joshtrewin7 This should now be available as prerelease `0.24.0rc1`!

joshtrewin7:
@valentijnnieman thank you so much! 
Loving how fast this new version is with everything pre-loaded on each tab. Makes this much smoother and nicer. Datepickerrange is broken as already mentioned. But I applied the fix and that works well for now. 

Thanks again for this updated version. Loving it so far! 
Cheers, 

Josh

valentijnnieman:
@joshtrewin7 @chuchuhao Turns out the issue was a missing - (a missing dash, wink wink) with the NPM package versioning. Basically it tried to get the CSS files from `0.24.0rc1` instead of `0.24.0-rc1`. Sorry for the trouble! I've published a new release `0.24.0rc2` which has the fix in it. Thanks for your help!

src/components/Tab.react.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Tab = ({ children }) => <div>{children}</div>;
+
+/**
+ * Part of dcc.Tabs - this is the child Tab component used to render a tabbed page.
+ * It's children will be set as the content of that tab, which if clicked will become visible.
```
chriddyp: * Its

src/components/Tabs.react.js
```diff
@@ -0,0 +1,258 @@
...
+        selectHandler(index, value);
+      }}
+    >
+      <span>{label}</span>
+      <style jsx>{`
+        .tab {
+          display: inline-block;
+          background-color: white;
+          border: 1px solid ${greyColor};
+          padding: 20px;
```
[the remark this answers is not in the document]
chriddyp: Actually, for the default tab styles in general, I prefer something like this:
```
    'width': 'calc(100% / 3)',
    'textAlign': 'center',
    'paddingTop': 15,
    'paddingBottom': 15,
    'paddingLeft': 25,
    'paddingRight': 25,
```
Full width, slightly wider than taller, centered text. Here's what it looks like:
image

Most users aren't going to change the defaults, so I want to make sure that they look as fresh as possible 😸

src/components/Tabs.react.js
```diff
@@ -0,0 +1,258 @@
...
+            }
+          }
+        `}</style>
+      </div>
+    );
+  }
+}
+
+Tabs.defaultProps = {
+  mobile_breakpoint: 1000
```
chriddyp: I wonder if we should make this smaller. On my tiny macbook, I hit this breakpoint pretty quickly:
image

What about something like `800`? Here's how that looks:
image

src/components/Tabs.react.js
```diff
@@ -0,0 +1,258 @@
...
+          }
+          .tab-content--vert {
+            display: inline-flex;
+            flex-direction: column;
+          }
+          @media screen and (min-width: ${this.props.mobile_breakpoint}px) {
+            .tab-parent--vert {
+              display: inline-flex;
+              flex-direction: row;
+            }
```
chriddyp: This is very very cool. Huge props to leading the charge with styled components.

I think that stacking the tabs looks great.
image

Are you happy with how the mobile view turned out?

src/components/Tabs.react.js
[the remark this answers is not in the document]
chriddyp: I think that getting the tabs to share the entire width is especially important since tabs can contain text with varying lengths. For example:
image

These tabs end up having different widths which looks unconventional. However, if the share an equal amount of space it ends up looking good:
image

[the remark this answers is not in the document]
chriddyp: When we make the tabs equal width, we might have to do something about getting them to be the same height if the text in one tab overflows to multiple lines:
image

I'm actually not sure how we could solve that one. Perhaps we just provide a solution in our documentation to override the tab styles. It's not immediately apparent to me what that solution would be. I tried setting `height: 80px` but ended up with this:
image

src/components/Tabs.react.js
```diff
@@ -0,0 +1,258 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+const greyColor = '#d6d6d6';
+const offWhiteColor = '#f9f9f9';
+const primaryColor = '#1975FA';
```
chriddyp: Thoughts on making these props as well? like:
```
colors = {
    border: 
    hover: 
    primary:
}
```

src/components/Tabs.react.js
```diff
@@ -0,0 +1,258 @@
...
+      : 'tab-parent';
+
+    return (
+      <div
+        className={`${tabParentClass} ${this.props.parent_className || ''}`}
+        style={this.props.parent_style}
+      >
+        <div
+          className={`${tabContainerClass} ${this.props.className || ''}`}
+          style={this.props.style}
```
chriddyp: I had a hard time figuring out what `style` was referring to. Could this be `tab_container_style`? and similarly `tab_container_className`

test/test_integration.py
```diff
@@ -366,6 +416,38 @@ def test_gallery(self):
 
         self.snapshot('gallery - text input')
 
+    def test_tabs_without_children(self):
+        app = dash.Dash(__name__)
+
+        app.layout = html.Div([
+            html.H1('Dash Tabs component demo'),
+            dcc.Tabs(id="tabs", value='tab-1', children=[
```
chriddyp: Let's change this to `tab-2`. I just tried this out on my machine and the first tab was selected on page load:
image
```python
app.layout = html.Div([
    dcc.Tabs(
        id="tabs",
        value='tab-2',
        children=[
            dcc.Tab(
                label='Tab 1',
                value='tab-1'
            ),
            dcc.Tab(
                label='Tab 2',
                value='tab-2'
            ),
            dcc.Tab(
                label='Tab 3',
                value='tab-3'
            )
        ]
    ),
    html.Div(id='container')
])


@app.callback(Output('container', 'children'), [Input('tabs', 'value')])
def display_content(value):
    return html.H1('This is {}'.format(value))
```

test/test_integration.py
```diff
@@ -366,6 +416,38 @@ def test_gallery(self):
...
+                return html.Div([
+                    html.H3('Test content 1')
+                ], id='test-tab-1')
+            elif(tab == 'tab-2'):
+                return html.Div([
+                    html.H3('Test content 2')
+                ], id='test-tab-2')
+
+        self.startServer(app=app)
+
```
chriddyp: Let's also take a percy screenshot in here

src/components/Tabs.react.js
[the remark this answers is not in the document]
valentijnnieman: What happens if you have more than 3 tabs? Or less than? :) 
screen shot 2018-07-11 at 10 45 09
The calc(100%/3) obviously only works when there's 3 tabs, I could try it with a variable for the amount of tabs. I don't think the full width tabs look great though - they're much too wide, and too flat, like they're stretched out. I guess my solution for the multiline problem was an easy one - just have them have dynamic widths. But you're right, that is maybe a bit unconventional, and could look very silly!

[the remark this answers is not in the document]
valentijnnieman: A variable actually works pretty well. Also, if you're like me and don't like the tabs to be so stretched out, you can easily set  `width: 50%` on the `container_style`! 
screen shot 2018-07-11 at 11 02 57

src/components/Tabs.react.js
valentijnnieman: Yeah, let's do it!

valentijnnieman: ;-)

test/test_integration.py
valentijnnieman: So not sure what you meant here, but I think changing it to `tab-2` is what's breaking the tests now (or it is the mystery what's been plaguing other tests too)

valentijnnieman:
@chriddyp So just had a question about what you meant with regards to the test, and after that I think we're good! Although, you might want to take out the style changes I made. I think they're good - but am curious what you think.

chriddyp:
> Although, you might want to take out the style changes I made. I think they're good - but am curious what you think.

OK great! The tab styles have been on my mind a lot the last couple of days. I was convinced that the full width tabs was the right solution but now you've got me thinking... 😺 

Another approach I was thinking about was github style tabs: 
image
So like what you had before with flex box but instead of boxes (which I think looked odd when the tabs had different lengths), just greyed out text.
Anyway, I'll check out the current iteration today and maybe do some more research on other styles. Thanks for your patience on this one!

valentijnnieman:
@chriddyp No worries! So the tabs in Github have different lengths too - you just only see the selected tab. That's a good idea actually. This is also pretty good: https://material-ui.com/demos/tabs/#basic-tabs Notice the use of flexbox to center the text, and they just cut it off if it's too long.

T4rk1n:
Could we get an `active` prop on individual tab? Like if you click on it and it is not active it doesn't change the tab. It would also takes in an `inactive_style` prop or be greyed out by default.

valentijnnieman:
@T4rk1n Good idea - I added a `disabled` prop and `disabled_style`, `disabled_className` props.

valentijnnieman:
I just published prerelease version `0.24.0rc3` which includes the latest updates.

rosswait:
@valentijnnieman  This feature is really great!  Looking forward to the main release :)

saurabhp75:
Are tabs included in the latest build?
I installed latest dash-core-components-0.24.1
AttributeError: module 'dash_core_components' has no attribute 'Tabs'

rosswait:
@saurabhp75 I was able to access them by installing the release candidate:

`pip install dash_core_components==0.24.0rc1`

test/test_integration.py
chriddyp: What I mean here is that if `tab-2` is the initial value, then the tab with the id `tab-2` should be selected. However, right now, this tab is not selected (`tab-1` is selected no matter what). So, we should fix this issue (setting the initial tab's value should select that particular tab) and ensure that it's fixed by using `value='tab-2'` in our test

valentijnnieman:
@Nemecsek @saurabhp75 Unfortunately this PR hasn't been merged yet - the `0.24.0rc5` release is considered a prerelease, therefore the latest release (`0.25.1` I believe) does not yet include Tabs. I hope to merge this today, though! Thanks for your patience :)

valentijnnieman:
@chriddyp Custom styles like `selected_style` are being appended, not replaced, otherwise it would mean that you'd have to style everything yourself, even if you just wanted to update a color or something. My guess is that this is why you have a double border somewhere.

lilianeolmos:
Hi @valentijnnieman ! I'm having an issue: when I select the tab for the first time, the layout is correct , but when I switch tabs and then return to it, the layout has been messed up:
tabs

valentijnnieman:
@chriddyp Sorry, I missed that completely when watching your gif :) I'll take a look at it!

valentijnnieman:
@chriddyp I made a mistake where I copied the styles over - thus leaving styles that were being appended (from `selected_style`). Good catch! It should work now - note though that you set `width: inherit` as `tab_style`, and `selected_style` will still have the `width: calc(25%)` on it, which will look weird.

valentijnnieman:
@lilianeolmos Hmm that's odd! I've just published prerelease version `0.24.0rc6`, could you use that (run `pip install dash_core_components==0.24.0rc6`) and see if that fixes anything?

lilianeolmos:
@valentijnnieman  thanks for your answer. I've just tried and I'm still having the same problem. When I use the previous versions of dash_core_components this problem doesn't happen.

valentijnnieman:
@lilianeolmos Hmm! Would you perhaps create a post on our forums, [here](https://community.plot.ly/c/dash) with a short reproducible example code of the bug? Would help us out a lot!

chriddyp:
Thanks for making this changes! Let's do this 😸 

 💃 💃 💃

danigallo:
@valentijnnieman is there a way to change the tab sizes/length? I have 4 and one is carrying over to a new row

valentijnnieman:
@danigallo You could specify the `style` of a `Tab` component and define it's own `width`. So `dcc.Tab(style={'width': '25%'}, children=[])` for example.

frogger72:
Thanks for all your work on the new Tabs functionality. My apps still use the old one from the 0.21.rc1 release and was wondering if it would be safe to switch over for my use case. Typically, each tab's content takes 20 seconds or so to load because of internal data fetching, computation and rendering of a complicated plotly Figure. Additionally, most tabs are never used during a given session, so the content never needs to be generated. Would the new component slow down the app?

Chris3691:
Hi

Fairly new to Plotly and Dash but loving what I can do with it so far! Have started experimenting with tabs and I have two tabs, but when I load my dashboard it opens my "second" tab as the first tab you see.

Is there a way to set tab1 as the "default" tab that should be displayed first?

Thanks

Chris

robfraz:
@Chris3691 Yes there is.   Look at the code example in [this earlier comment](https://github.com/plotly/dash-core-components/pull/213#issuecomment-407285268) where a bug (now resolved) was spotted with this default tab functionality.

You'll see each `Tab` object has a value specified (`value='tab-1'`, `value='tab-2'`, etc), and there is also a value specified in the `Tabs` object, which is set to `'tab-3'` in that example, thus making the `Tab` object with the value `'tab-3'` the default choice.

chriddyp:
We have published the documentation for this component online here: https://dash.plot.ly/dash-core-components/tabs. Many thanks to everyone who has contributed to this effort!

If you have any other questions or issues with this component, please create a new issue or discuss in the community forum https://community.plot.ly/c/dash

PR #215, opened 2018-06-22
Improved Link component

### LINKED ISSUE #99: control key support and semantic links for `dcc.Link`
`dcc.Link` is the Dash single-page-app link component. There are still a few behaviours that make these links feel less like native links:
1. Can't right-click on them
2. Can't cntl-click to open new pages with them
3. Visiting a new page keeps the scroll location at the bottom of the page, rather than bringing the user back to the top

For 3, we may need to solve this in the `dash-renderer` library, but I'm not sure.

It looks like we could use the logic in https://github.com/ReactTraining/react-router/blob/d6ac814e273b2f3aad24b7630a6f1df3f19c1475/packages/react-router-dom/modules/Link.js#L41-L62

valentijnnieman:
This improves the Link component somewhat, so that users can right-click on it etc. It also scrolls back to top when clicking a link. The fix is basically adding a href attribute to the <a> tag rendered, so that the browser treats it as a native link. Should fix #99


src/components/Link.react.js
```diff
@@ -2,69 +2,72 @@
...
 
 export default class Link extends Component {
-    constructor(props) {
-        super(props);
-        this.updateLocation = this.updateLocation.bind(this);
-    }
+  constructor(props) {
+    super(props);
+    this.updateLocation = this.updateLocation.bind(this);
+  }
```
chriddyp: We should really enforce style with eslint, but for now our convention has been tabs with 4 spaces (to match the code that we write in our python context)

src/components/Link.react.js
```diff
@@ -2,69 +2,72 @@
 
 import PropTypes from 'prop-types';
 
-import React, {Component} from 'react';
+import React, { Component } from 'react';
```
chriddyp: Similarly here - this isn't enforced in eslint (although it should be), but our convention has been no spaces around `{}` imports

src/components/Link.react.js
```diff
@@ -2,69 +2,72 @@
...
-    return evt;
+function CustomEvent(event, params) {
+  params = params || {
+    bubbles: false,
+    cancelable: false,
+    detail: undefined
+  };
+  var evt = document.createEvent('CustomEvent');
+  evt.initCustomEvent(event, params.bubbles, params.cancelable, params.detail);
+  return evt;
```
chriddyp: In general, it'd be helpful if you split out commits that did style updates vs commits that added or removed code. As a reviewer, it takes more time to understand which lines have actual changes

src/components/Link.react.js
```diff
@@ -2,69 +2,72 @@
...
-        const {href, refresh} = this.props;
-        if (refresh) {
-            window.location.pathname = href;
-        } else {
-            window.history.pushState({}, '', href);
-            window.dispatchEvent(new CustomEvent('onpushstate'));
-        }
+  updateLocation(e) {
+    e.preventDefault(); // prevent anchor from updating location
+    const { href, refresh } = this.props;
```
chriddyp: Similarly here, our convention is no white spaces around `{}`, so `const {href, refresh} = this.props`

src/components/Link.react.js
valentijnnieman: Ah yep, I have Prettier running by default on save in VS Code, so if there's nothing in eslint then it will format it like this I guess!

src/components/Link.react.js
valentijnnieman: I think this is another case of the Prettier autoformatter doing "it's job" :) But duly noted!

chriddyp:
This is very interesting. I wonder how we can test this with our integration tests.
- For the scroll behaviour, maybe we can create a multi-page dash app example where the pages are really long and the links are towards the bottom. Ideally, after clicking on a link, we see the content at the top of the second page rather than the bottom of the second page.
- We also need to test that clicking on a link still fires an update on `dcc.Location` (rather than reloading the entire page), i.e. that the `preventDefault` is working as expected
- As for the regular link behaviour, like right clicking on them, I'm not sure how we should test.

Let's publish a prerelease version and test this on the dash-docs page. To create a prerelease:
1 - Update `version.py` to be `0.23.1rc1`
2 - Update `package.json` to be `0.23.1-rc1` (notice how the `rc` syntax is different between node and python. npm requires a `-` between the version number and the prerelease tag while python's pip just has `0.23.1rc1`)
3 - Update `__init__.py` to reference the new node version:
https://github.com/plotly/dash-core-components/blob/16515bb2279e11102f39f46781cde26b5de8c3f7/dash_core_components/__init__.py#L35
should be `0.23.1-rc1`
4 - Publish: `npm run publish-all` (I'll give you access to NPM and PyPI)
5 - Create a PR in dash-docs by editing the `requirements.txt` to contain the new prerelease version (`0.23.1rc1`): https://github.com/plotly/dash-docs/blob/master/requirements.txt#L8. In that repo. PRs automatically deploy a staged versions of the app.

src/components/Link.react.js
bpostlethwaite: I'll set up prettier and eslint in CI

bpostlethwaite: including a prettierrc in the root.

valentijnnieman:
@chriddyp I've written a test that should test if the page scrolls back up (should verify snapshot on Percy), if it updates the Location, and tests if the <a> tag that is rendered by Link contains an href attribute with the correct href. My thinking is that if this is the case, then the <a> tag should/must be right-clickable in browsers.

valentijnnieman:
I've published a prerelease version of this: `0.23.1rc1` and made a PR to the Docs [here](https://github.com/plotly/dash-docs/pull/111) so we can try it out.

chriddyp:
This looks awesome!
link-component

One thing that I notice is that the standard `cmd-click` to open a new tab doesn't seem to work. Here's a GIF. The first couple of clicks are standard `<a/>` links, the last click to the gallery is a `dcc.Link`
link-component-cntrl-click

I'd be curious to know why that was the case. It's not a blocker for me as this is already a lot better

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.23.1]
+### fixed
+- Link component now is a proper <a> tag so you can right click on it, and will scroll back to top. [#99](https://github.com/plotly/dash-core-components/issues/99)
```
chriddyp: Maybe say "Fixes [#99](...), implemented in [#215](https://github.com/plotly/dash-core-components/pull/215/)"

> Link component now is a proper <a> tag. This allows you to right-click on `dcc.Link` and see the standard html link context menu. Also, clicking on a `dcc.Link` will now scroll the next page to the top, rather than keeping the scroll position at the position of the previous page.

test/test_integration.py
```diff
@@ -494,6 +494,43 @@ def update_pathname(n_clicks, current_pathname):
...
+                'height': '200vh'
+            }),
+            html.Div(id='page-content'),
+            dcc.Link('Test link', href='/test-link', id='test-link')
+        ])
+
+        @app.callback(Output('page-content', 'children'),
+                       [Input('test-url', 'pathname')])
+        def display_page(pathname):
+            return 'You are on page {}'.format(pathname)
```
chriddyp: Let's increment a `multiprocessing.Value` in here so that we can ensure that the callback is only fired a single time. 
Here's an example of incrementing `Value` in a callback:
https://github.com/plotly/dash-renderer/blob/70bb93e8f0151ddc7d0810f3df271036a069360a/tests/test_render.py#L458-L463

and further down, the assertion:
https://github.com/plotly/dash-renderer/blob/70bb93e8f0151ddc7d0810f3df271036a069360a/tests/test_render.py#L478-L484

src/components/Link.react.js
```diff
@@ -27,28 +26,32 @@ export default class Link extends Component {
...
+    updateLocation(e) {
+        e.preventDefault(); // prevent anchor from updating location
+        const { href, refresh } = this.props;
         if (refresh) {
             window.location.pathname = href;
         } else {
             window.history.pushState({}, '', href);
             window.dispatchEvent(new CustomEvent('onpushstate'));
         }
+        window.scrollTo(0, 0); // scroll back to top
```
chriddyp: Very sweet. For some reason I thought we would need to do this _after_ the content was updated, but after playing around with it seems like this totally works 👏

ned2:
So happy that these issues are being fixed!!

PR #217, opened 2018-06-22
Add integration test for a single tab

kyleabeauchamp:
cc @valentijnnieman


kyleabeauchamp:
The CI test failure is expected, as this PR adds tests for a bug I report in #213 .

valentijnnieman:
@kyleabeauchamp Ah nice catch! Let's check back on this after #213 is merged. Thanks for your help!

PR #219, opened 2018-06-28
Set Graph id to be optional

valentijnnieman:
This sets the Graph's id prop to not be required, and defines a defaultProp value for it set to 'untitled-graph' so Plotly.js still works.


chriddyp:
Nice! I believe that we'll need the graph ID to be dynamic so that users can embed multiple graphs. There are a few places in the plot code that require unique IDs, like:
```
 Plotly.Plots.resize(document.getElementById(this.props.id));
```

and 
```
PlotMethod(id, figure.data, figure.layout, config)
```

valentijnnieman:
@chriddyp Yep, makes sense. So if the user defines more than one `dcc.Graph` without providing an `id` they will both have the `defaultProp.id` of `'untitled-graph'`. Any suggestion of how to fix that? Create a random string?

chriddyp:
> Any suggestion of how to fix that? Create a random string?

Yeah, I think that's what we'll have to do

dash_core_components/metadata.json
```diff
@@ -139,73 +139,6 @@
       }
     }
   },
-  "src/components/Confirm.react.js": {
```
bpostlethwaite: Why is this getting nuked?

T4rk1n: It was called `Confirm` at first, but I renamed it to `ConfirmDialog`.

valentijnnieman: Yeah, so that's probably what happened.

valentijnnieman:
@chriddyp Do you want to take a look at the Percy screenshot?

chriddyp:
> @chriddyp Do you want to take a look at the Percy screenshot?

Woops, sorry I missed this. In the future, feel free to go in there and approve them yourselves :) 

💃

PR #225, opened 2018-07-03
Update readme with newer information

valentijnnieman:
I've updated the readme a little bit by removing things that are no longer relevant / out of date, and with stuff I found over the last week works. When I started working in this repo I was a bit confused by the directions on how to run unit tests in Karma (when there are none) for example - so I removed that bit. After this, updating the scripts in package.json could also be neat!

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ You have to maintain the list of components in `demo/Demo.react.js`.
 ### Code quality and tests
 
 ### To run integration tests (test_integration.py)
-We run our integration tests on CircleCI with help from Tox. There’s a tox.ini file which holds the configuration, refer to [tox's documentation](http://tox.readthedocs.io/en/latest/index.html) for help. I found that I needed to set environment variables in my terminal, like `TOX_PYTHON_27` to my version of python that I wanted tox to use. So running:
+We run our integration tests on CircleCI with help from Tox. There’s a tox.ini file which holds the configuration, refer to [tox's documentation](http://tox.readthedocs.io/en/latest/index.html) for help. You may need to set environment variables in your terminal, like `TOX_PYTHON_27` to my version of python that I wanted tox to use. So running:
 
 ```sh
 export TOX_PYTHON_27=python2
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,42 +59,18 @@ You have to maintain the list of components in `demo/Demo.react.js`.
 
 ### Code quality and tests
 
-#### To run lint and unit tests:
+### To run integration tests (test_integration.py)
+We run our integration tests on CircleCI with help from Tox. There’s a tox.ini file which holds the configuration, refer to [tox's documentation](http://tox.readthedocs.io/en/latest/index.html) for help. I found that I needed to set environment variables in my terminal, like `TOX_PYTHON_27` to my version of python that I wanted tox to use. So running:
```
cldougl: This is super nit-picky 🙈 but I prefer when the READMEs use the same pro-noun scheme throughout. So instead of saying: 'I found I needed to...'
use: 'You may need to...'

README.md
```diff
@@ -142,26 +118,18 @@ publishing steps into one workflow.
...
-1. Preparing to publish to NPM
-
-        # Bump the package version
-        $ npm version major|minor|patch
-
-        # Push branch and tags to repo
-        $ git push --follow-tags
-
-2. Preparing to publish to PyPi
+1 - Update `version.py`
```
cldougl: we're using semantic versioning in this repo right? (cc @chriddyp )
If that's the case, we may want to link to docs for people not super familiar with how to bump the versioning https://semver.org/

README.md
```diff
@@ -142,26 +118,18 @@ publishing steps into one workflow.
...
 
-        # Commit to github
-        $ git add setup.py
-        $ git commit -m "Bump pypi package version to vx.x.x"
-
-3. Publish to npm and PyPi
+4 - Publish: `npm run publish-all` (Make sure you have access to NPM and PyPI)
+4b - If the `publish-all` script fails on the `twine` command, try running
+```sh
+twine upload dist/dash_core_components-xx.x.x.tar.gz # where xx.x.x is the version number
```
cldougl: @valentijnnieman I may be misunderstanding but was there a reason for the `xx` in the `xx.x.x` representation of the version? I would lean towards just using x.x.x to notate the version as I believe those digits will vary depending on what version we're on.

README.md
```diff
@@ -142,26 +118,18 @@ publishing steps into one workflow.
...
+```
 
-        $ npm run publish-all
+If you want to publish a prerelease package, change `version.py` to X.XX.Xrc1 (0.23.1rc1 for example) and
+`package.json` to X.XX.X-rc1 (notice how the rc syntax is different between node and python. npm requires a - between the version number and the prerelease tag while python's pip just has 0.23.1rc1)
```
cldougl: similarly here, I would probably suggest `X.X.X-rc1` (and maybe even `-rcX` so it's clear that you could add a subsequent prerelease if necessary).

cldougl: also probably use the same case x as the twine section above

README.md
valentijnnieman: Good point! Although 'I' is used in other parts of the readme as well, as in 'I recommend installing ...' etc. I'll change it to 'you may need to' regardless!

cldougl:
lgtm now! 💃

Issue #222, opened 2018-07-02
Ability to set a maximum number of intervals?


chriddyp:
As suggested by a community member:

> When I put it this way, the cleanest solution to my problem would be if `dcc.Interval()` supported the idea of counting a certain number of ticks and then stopping by itself, i.e. had something like a `max_intervals` parameter. I’m not sure if you have any other use-cases for that but if so, maybe it’s a feature suggestion?

Full discussion: https://community.plot.ly/t/should-i-use-an-interval-for-a-one-off-delay/11375/6

Seems like a pretty good idea, coul dbe useful for general delays and fits nicely into the existing component API

claresloggett:
Might also be useful to people trying to "turn off" the Interval, e.g. https://community.plot.ly/t/how-to-turn-off-interval-event/5565 ?

That discussion is about turning it off in response to an event, but if there were a `max_intervals` parameter we could presumably turn the timer off manually by setting `max_intervals=n_intervals`.

PR #226, opened 2018-07-04
Added max_intervals prop to Interval component

valentijnnieman:
As per #222.

edit: 
A quick example:
```
...
app.layout = html.Div([
    html.Div(id='output'),
    dcc.Interval(id='interval', interval=2000, max_intervals=2)
])


              [Input('interval', 'n_intervals')])
def update_text(n):
    return "This page has been updated {} times, max_interval=2".format(n)
...
```

### DIFF
--- a/src/components/Interval.react.js
+++ b/src/components/Interval.react.js
@@ -35,7 +35,7 @@ export default class Interval extends Component {
     }
 
     componentWillReceiveProps(nextProps) {
-        if(nextProps.n_intervals < this.props.max_intervals || this.props.max_intervals === 0) {
+        if(nextProps.n_intervals < this.props.max_intervals || this.props.max_intervals === -1) {
             if ((!this.props.fireEvent && nextProps.fireEvent) ||
                 (!this.props.setProps && nextProps.setProps)
             ) {
@@ -102,5 +102,5 @@ Interval.propTypes = {
 Interval.defaultProps = {
     interval: 1000,
     n_intervals: 0,
-    max_intervals: 0
+    max_intervals: -1
 };
--- a/src/components/Interval.react.js
+++ b/src/components/Interval.react.js
@@ -35,15 +35,21 @@ export default class Interval extends Component {
     }
 
     componentWillReceiveProps(nextProps) {
-        if ((!this.props.fireEvent && nextProps.fireEvent) ||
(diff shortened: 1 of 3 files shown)


rmarren1:
Is this supposed to automatically stop the interval from triggering the `n_intervals` property is `>=` `max_intervals`? This is the behavior I expected, but it does not seem to be working that way.
Using this code:
```
import dash

app = dash.Dash('hello')


app.layout = html.Div([
    html.Div(id='output'),
    dcc.Interval(id='interval', interval=2000, max_intervals=2)
])



              [Input('interval', 'n_intervals'),
               Input('interval', 'max_intervals')])
def update_text(n, M):
    return "This page has been updated {} times, max_intervals={}".format(n, M)


app.run_server(debug=True)

```
I get this:

image

valentijnnieman:
@rmarren1 Thanks for reviewing! You need to set `scripts.config.serve_locally = True` in order for the local changes to work - otherwise Dash will get the components from unpkg, which won't have the changes I made! I will specify that in the example next time, left that stuff out to make it shorter.

valentijnnieman:
Happens to me all the time too. That's a good point - I'll rewrite it to use `-1`! Thanks again.

rmarren1:
Great, I think the docstring for `max_intervals` should explain that -1 means no maximum, but other than that I am :dancer:

src/components/Interval.react.js
```diff
@@ -75,6 +81,11 @@ Interval.propTypes = {
      */
     n_intervals: PropTypes.number,
 
+    /**
+     * Number of times the interval will be fired
```
chriddyp: > Number of times the interval will be fired. If -1, then the interval has no limit (the default) and if 0 then the interval stops running.

test/test_integration.py
```diff
@@ -528,3 +528,22 @@ def update_graph(n_clicks):
...
+        ])
+
+        @app.callback(Output('output', 'children'),
+                    [Input('interval', 'n_intervals')])
+        def update_text(n):
+            return "{}".format(n)
+
+        self.startServer(app=app)
+
+        time.sleep(2) # wait for interval to finish
```
chriddyp: I'd make this like `time.sleep(5)` just to make sure that you didn't catch it on `2` "by chance"

PR #239, opened 2018-07-18
Change package.json scripts to work on windows.

T4rk1n:
Had problems running the `package.json` scripts from my windows machine so I came up with this solution.

- added `copyfiles` because `cp` doesn't work on windows.
- added `cross-env` because setting `NODE_ENV `directly on windows doesn't work.


bpostlethwaite:
:dancer:

PR #241, opened 2018-07-23
Fix display of ConfirmDialogProvider if not used in a callback.

### LINKED ISSUE #242: ConfirmModel - `submit_n_clicks_timestamp` is fired when the cancel button is clicked
In this example, the callback is fired when the Cancel button is clicked. It should not be fired.

```python
import dash

app = dash.Dash()

app.layout = html.Div([
    dcc.ConfirmDialogProvider(
        children=html.Button(
            'Click Me',
        ),
        id='danger-danger',
        message='Danger danger! Are you sure you want to continue?'
    ),
    html.Div(id='output')
])


              [Input('danger-danger', 'submit_n_clicks_timestamp')])
def update_output(submit_n_clicks_timestamp):
    if not submit_n_clicks_timestamp:
        return ''
    return """
        It was dangerous but we did it!
        Submitted at {}
    """.format(submit_n_clicks_timestamp)


if __name__ == '__main__':
    app.run_server(debug=True)
```

cancel-button

T4rk1n:
- Add support to use the ConfirmDialog and ConfirmDialogProvider without a callback assigned to it.
- Fix callbacks sent to both submit and cancel at the same time.

Fix #242.


src/components/ConfirmDialog.react.js
```diff
@@ -11,30 +12,54 @@ export default class ConfirmDialog extends Component {
 
     constructor(props) {
         super(props);
+        this.state = {
+            displayed: props.displayed
+        };
+        this._setStateAndProps = this._setStateAndProps.bind(this);
```
bpostlethwaite: I don't see _setStateAndProps being passed as a callback anywhere so this .bind() isn't necessary?

src/components/ConfirmDialog.react.js
```diff
@@ -11,30 +12,54 @@ export default class ConfirmDialog extends Component {
...
+        this._update = this._update.bind(this);
+        // this._update();
```
bpostlethwaite: ⚡️

chriddyp:
@T4rk1n are you planning on adding tests to this one as per @bpostlethwaite 's recommendation? Or are confirm dialog tests too hard to add with selenium?

Also, does this fix https://github.com/plotly/dash-core-components/issues/242?

This PR is blocking the documentation https://github.com/plotly/dash-docs/pull/123, so it'd be good to get it in soon 😺

T4rk1n:
I added tests for ConfirmDialog without callbacks and fixed #242.

@chriddyp please review.

chriddyp:
💃 thanks!

PR #270, opened 2018-08-21
Update circle.

T4rk1n:
### Update the config of circleci to version 2

- Removed tox
- Build with `Builder` was failing, added the configs of `dash-component-boilerplate`.
- There is nothing but the tests to pylint, and they fail with really low score, so Pylint is out for now.

The test_link_scroll https://github.com/plotly/dash-core-components/blob/a727cdec2e2021d3c6f1c5bb77941d61fde9f981/test/test_integration.py#L604-L616 is somewhat brittle and sometimes fails.


valentijnnieman:
@Marc-Andre-Rivet Could you perhaps review this if you have time? I think you know this stuff pretty well!

.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
+version: 2
+
+jobs:
+  "python-2.7": &test-template
```
Marc-Andre-Rivet: Not aware of this syntax '&test-template', you use it to refer to this job configuration for python36 and 37 so you can just override what you need?

.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
...
+          name: Build
+          command: |
+              . venv/bin/activate
+              npm run build:js
+              npm run build:py
+
+      - run:
+          name: Run tests
+          command: |
+              npm run test
```
Marc-Andre-Rivet: Not important, but you can do 'npm test' instead

package.json
```diff
@@ -17,17 +17,25 @@
     "publish-all": "npm publish && python setup.py sdist upload",
     "publish-pypi": "npm run prepublish && python setup.py sdist && twine upload --sign --skip-existing",
     "start": "./node_modules/.bin/builder run build-dev",
-    "test": "./node_modules/.bin/eslint --fix .",
+    "test": "./node_modules/.bin/eslint --fix src",
```
Marc-Andre-Rivet: I find it weird that test is used to fix linting errors. Should probably be a script of its own like 'fixlint', 'lint.fix' or something along those lines.
You should be able to call 'eslint --fix src' directly instead too, same for builder above.

I would expect eslint to be in the list of dependencies if we are calling it in our scripts. We should not trust babel-eslint to pull it along.

package.json
```diff
@@ -17,17 +17,25 @@
...
     "test-debug": "./node_modules/.bin/builder run test-frontend-debug",
-    "uninstall-local": "pip uninstall dash-core-components -y"
+    "uninstall-local": "pip uninstall dash-core-components -y",
+    "build:js": "webpack --mode production",
+    "build:py": "node ./extract-meta src/components > dash_core_components/metadata.json && copyfiles package.json dash_core_components"
   },
   "author": "Chris Parmer <chris@plot.ly>",
   "license": "MIT",
   "dependencies": {
+    "babel-core": "^6.26.3",
```
Marc-Andre-Rivet: Do we have a shrinkwrap or package-lock file for this project?

.circleci/config.yml
T4rk1n: Yes, I took that from plotly/dash#222.

webpack.config.js
```diff
@@ -0,0 +1,41 @@
...
+                        loader: 'style-loader'
+                    },
+                    {
+                        loader: 'css-loader'
+                    }
+                ]
+            }
+        ]
+    }
+};
```
valentijnnieman: What's the reason for this new `webpack.config.js` file? The webpack config stuff for the components is in `./config/webpack/`, won't it be confusing to have 2 config files for webpack?

valentijnnieman:
@T4rk1n Maybe it's a good idea to update the `.eslintrc` file too while we're at it? In #274 I've used the settings taken from `plotly/streambed`, maybe we can do that here too? What do you think?

T4rk1n:
@valentijnnieman Yes we should add the prettier here, Its in the list plotly/dash#328

webpack.config.js
T4rk1n: The old configs were for webpack 3, this is for webpack 4.

valentijnnieman:
@T4rk1n Cool! You could try merging in the #274 branch if you want!

T4rk1n:
@valentijnnieman I have added prettier and fixed the errors it was reporting, for the tabs file I put a eslint-disable at the top because it was reporting a bunch of prop types error for the EnhancedTab. Please review.

valentijnnieman:
@T4rk1n Oh yeah `<EnhancedTab/>` isn't using propTypes. I'll fix that later. Nice work! 💃

chriddyp:
💃

Issue #169, opened 2018-02-26
dcc.Input bug with decimal values


bperrenoud:
when using the input component with numbers and with a callback

    import dash
    from dash.dependencies import Input, Output
    import dash_core_components as dcc
    import dash_html_components as html

    app = dash.Dash()

    app.layout = html.Div([
        dcc.Input(placeholder='Enter a value...',
            type='number',
            value=0,
            step=0.01,
            id='input_id',
            style={'float': 'left'}),
        html.Div(id='my-div')
    ])

    @app.callback(
        Output(component_id='my-div', component_property='children'),
        [Input(component_id='input_id', component_property='value')]
    )
    def update_output_div(input_value):
        return 'You\'ve entered "{}"'.format(input_value)

    if __name__ == '__main__':
        app.run_server()

    
Try entering decimal values with some 0s in the decimal part, for instance 0.001 or 4.1004:
When the callback is triggered, any ending 0 will be removed preventing you to finish writing the number

chriddyp:
Looks like this bug was added in #100 , here:
https://github.com/plotly/dash-core-components/blob/9478e4f66a9e1ada3b4d214a6b426436ef2e0b82/src/components/Input.react.js#L34-L36

For now, you can use `type='text'` and do the casting server-side in your dash app code

pope1ni:
I think the best solution here is to debounce `onChange` for `dcc.Input()`.
This will allow the user to finish what they are typing before the event handler triggers.

moorepants:
I agree with @pope1ni, the input boxes shouldn't send numbers to the server before the user has finished typing them. I'm having this trouble when entering multi digit numbers. For example, if I want to enter 20 manually, 2 gets sent to the server before I finish typing.

steven-murray:
Any timeline for this being fixed? It is very annoying! I'd rather not do the casting server-side because it makes it harder to let the user know what went wrong to fix it.

valentijnnieman:
Version `0.37.0` now includes a new `debounce` prop on the input component, that when true, only sends back the data to the server upon 'Enter' or when losing focus. Hopefully this fixes the issues you've been having. I will close this issue now, assuming that the issue has been fixed. Thanks for your patience!

vantaka2:
@valentijnnieman  - Did .37 fix the issue? I'm using .38 and seeing this same bug.

chriddyp:
[...]

PR #277, opened 2018-08-24
fixes plotly/dash-core_components#169

Bachibouzouk:
fixes plotly/dash-core_components#169

Fix the decimal problem by using `onBlur` instead of  `onChange`. However this solution works only if the line `value={value}` in the `render` function is commented.

### DIFF
--- a/src/components/Input.react.js
+++ b/src/components/Input.react.js
@@ -25,25 +25,29 @@ export default class Input extends Component {
             setProps,
             type
         } = this.props;
-        const {value} = this.state;
+        //const {value} = this.state;
         return (
             <input
-                onChange={e => {
-                    this.setState({value: e.target.value});
-                    if (setProps) {
-                        if (type === 'number') {
-                            setProps({value: Number(e.target.value)});
+                onBlur={
+                    event => {
+                        this.setState({value: event.target.value});
+                        if (setProps) {
+                            if (type === 'number') {
+                                setProps({value: Number(event.target.value)});
+                            }
+                            else {
+                                setProps({value: event.target.value});
+                            }
                         }
-                        else {
-                            setProps({value: e.target.value});
-                        }
+                        if (fireEvent) fireEvent({event: 'blur'});
                     }
-                    if (fireEvent) fireEvent({event: 'change'});
-                }}
-                onBlur={() => {
-                    if (fireEvent) fireEvent({event: 'blur'});
-                }}
-                value={value}
+                }
+                onChange={
+                    () => {
+                        if (fireEvent) fireEvent({event: 'change'});
+                    }
+                }
+                //value={value}
                 {...omit(['fireEvent', 'setProps', 'value'], this.props)}
             />
         );
@@ -243,6 +247,11 @@ Input.propTypes = {
      * Works with the min and max attributes to limit the increments at which a numeric or date-time value can be set. It can be the string any or a positive floating point number. If this attribute is not set to any, the control accepts only values at multiples of the step value greater than the minimum.
      */
     step: PropTypes.string,
+
+    /**
+     *  Sets the number of important digits. It can be the string any or a positive floating point number. If this attribute is not set to any, the default is zero.
+     */
+    precision: PropTypes.string,
 
     /**
      * Dash-assigned callback that gets fired when the input changes.


src/components/Input.react.js
```diff
@@ -25,25 +25,29 @@ export default class Input extends Component {
             setProps,
             type
         } = this.props;
-        const {value} = this.state;
+        //const {value} = this.state;
```
valentijnnieman: If you comment this out, it will mean that a Dash user cannot provide the `value` prop to the `dcc.Input` component (or rather, it will not do anything). You can see in the `constructor` of the class that `this.state.value` is set to `props.value`, and below in `<input />`, it gets set to the value attribute: `<input value={value} />`.

src/components/Input.react.js
```diff
@@ -25,25 +25,29 @@ export default class Input extends Component {
...
+                        if (fireEvent) fireEvent({event: 'change'});
                     }
-                    if (fireEvent) fireEvent({event: 'change'});
-                }}
-                onBlur={() => {
-                    if (fireEvent) fireEvent({event: 'blur'});
-                }}
-                value={value}
+                }
+                //value={value}
```
valentijnnieman: The reason that this doesn't work without removing the `value={value}`, is probably because `onBlur` only fires the update, well, on blur, as opposed to `onChange` which fires every change. Setting `value` here will overwrite anything you put in the input (try having just `<input value={value} />` in there, and you'll see what I mean - it is as if it's 'locked' to whatever is put in as the `value` prop) so `onBlur`will fire it's update on blur, but because the input is 'locked', it fires the update with the same value. Does that make sense?

Bachibouzouk: Yes it make sense, maybe it could work with another variable, i.e. `typed` for example. I will try

Bachibouzouk:
When I try to get the details of the Percy tests I get the following error :
```
Error: Ember Data Request GET /api/v1/organizations/plotly returned a 403
Payload (Empty Content-Type)
[object Object]
```

I guess I should update the changelog by pulling the latest version of master before merging?

chriddyp:
> Error: Ember Data Request GET /api/v1/organizations/plotly returned a 403

I'll need to add you to percy first

> I guess I should update the changelog by pulling the latest version of master before merging?

Yeah, or rebase on master and then update the changelog

Bachibouzouk:
Couldn't we use `<NumericInput/>` instead of `<input type='number'/>` ? The first one provides the little stepper on the right of the input box for edge, whereas `<input type='number'/>` doesn't

https://www.npmjs.com/package/react-numeric-input

chriddyp:
> Couldn't we use <NumericInput/> instead of <input type='number'/> ? 

For this bug, I'd rather write the logic ourselves rather than introduce more dependencies

Bachibouzouk:
Assigning `value={value}` generates a warning : `A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components`

For the logic of this component do we want that the value gets changed at every keystroke? or only when the user is done writing?

chriddyp:
Considering the case for decimal numbers, trying to type out `0.1005` (and thereby needing to type `0.10`, a representation that doesn't exist in Python or Javascript), we have two options:
1 - Fire the callback at `0`, `0.1`, `0.1005`. Keeping an internal representation of the value as a `string` (so our state is `0`, `0.`, `0.1`, `0.10`, `0.100`, `0.1005`) and setting that value as the `<input value={this.state.value}/>` and only calling `setProps` when there aren't any trailing zeros after the decimal place. This would fix the bug and introduce no new properties
2 - Alternatively, we can introduce a new property similar to the Slider's `updatemode: PropTypes.oneOfType(PropTypes.oneOf(['keypress', 'blur', 'enter']), PropTypes.arrayOf(PropTypes.oneOf['keypress', 'blur', 'enter']))`. That is, the user could specify `updatemode: 'keypress'` or `updatemode: 'blur'` or combinations, e.g. `updatemode: ['blur', 'enter']`. Then, the numerical input developers could set `updatemode: ['blur', 'enter']` and not worry about the keypress case. For backwards compatability, `updatemode: 'keypress'` would have to remain the default.

Both solutions would be valid. (1) is strictly correct but (2) seems like it would satisfy the users who reported the bug in https://github.com/plotly/dash-core-components/issues/169 and provide a new feature in general.

Bachibouzouk:
Because the react `<input/>` component needs to update the `value` prop in order to display live whatever the user is typing, a solution would be to create a new prop called `inputValue` to distinguish with the `value` prop. `inputValue` would be updated in react by `onBlur` (when the user click away from the `<input/>` component or press the tab key for example). This would address the backcompatibility of the `value` prop

I tried this in React and it works, but I still have problem to trigger the callbacks in a similar fashion in Dash. I will commit a cleaned up code later today hopefully

Bachibouzouk:
@valentijnnieman I am not sure how I should make `keypress` the default value of the `updatemode` prop?

Bachibouzouk:
One thing that I am unable to control now is that the used is allowed to feed a string with content not matching a number format, i.e. `'12.2,12'` 

```
dcc.Input(
        id='input3',
        type='number',
        value='12.2,12', \\this value will be assigned but not displayed Inside the component
        updatemode='blur',
)
```


Another thing (maybe linked) is that the user is allowed to type in a wrong number format (i.e. it displays it although the value of the `value` prop is set to `''`. This was not the case when I tested the code in a [dummy example](https://codepen.io/bachibouzouk/pen/LJjYoK) on codepen.

src/components/Input.react.js
```diff
@@ -24,294 +24,297 @@ function formatValue(val, type){
...
+
+    render() {
+        const {fireEvent, setProps, type, updatemode} = this.props;
+
+        const {value} = this.state;
+        return (
+            <input
+                onBlur={() => {
+                    const newValue = formatValue(this.state.value, type);
+                    if (updatemode === 'blur') {
```
valentijnnieman: Here, if `updatemode` is an array (i.e. `['blur', 'enter']`) this will not trigger. The best way to do this would probably be to first see if `updatemode` is an array (you could use Ramda's `is` function for that, something like `R.is(Array, updatemode)`) and then check if that array contains `blur` (Ramda has a [contains](https://ramdajs.com/docs/#contains) function that can do that easily)

src/components/Input.react.js
```diff
@@ -24,294 +24,297 @@ function formatValue(val, type){
...
+                        }
+                        if (fireEvent) {
+                            fireEvent({event: 'blur'});
+                        }
+                    }
+                }}
+                onChange={e => {
+                    const newValue = formatValue(e.target.value, type);
+                    this.setState({value: newValue});
+                    if (updatemode === 'keypress') {
```
valentijnnieman: Similarly here - this `if` only works if `updatemode` is a string, not when it's an array.

src/components/Input.react.js
```diff
@@ -24,294 +24,297 @@ function formatValue(val, type){
...
+                        if (keyValue === '-' && curVal.length > 0) {
+                            e.preventDefault();
+                        }
+                        if (
+                            (keyValue === '.' || keyValue === ',') &&
+                            /[.]|,/.test(curVal.value)
+                        ) {
+                            e.preventDefault();
+                        }
+                    }
```
valentijnnieman: Could you explain a bit about what this code is supposed to do? Is it really necessary?

src/components/Input.react.js
valentijnnieman: Or maybe even `if (updatemode === 'blur' || R.contains('blur', updatemode)) {}` will work!

Bachibouzouk: actually I implemented with `updatemode` being `blur` (update when pressing tab or clicking away) or `keypress` (update at everykeystrokes), I wasn't sure the if we really need to have `['blur', 'enter']` but I kept it in the `Input.Props`

src/components/Input.react.js
Bachibouzouk: This rejects the `onKeyPress` event and the `onChange` event is therefore not triggered. This is used to make sure the user can only press on 0-9, dot and comma and that the user cannot input Something like `0.122.,233,44.4`, it prevents the use of another dot or comma when either one of them is already present

valentijnnieman:
Nice! Couple of tiny things though.
clear_with_dots
You can enter multiple `.`'s at the start - also characters like `-`

begin_with_dots
Here if you enter a number and then a `.`, it clears the input upon entering another `.`

I'm not sure if the way to go here is to prevent all these types of inputs. I think having the normal behaviour of an `<input />` tag is totally fine, we just have to be careful of when to send back the value to Dash. The new update methods are already a good work-around for those having trouble with inputs and floating point numbers, so maybe that's enough. 

@chriddyp Could you weigh in on this?

Bachibouzouk:
> Here if you enter a number and then a ., it clears the input upon entering another .
It doesn't do that for me

However I also have the problem of being able to input many dots. I can fix it, but it could also be decided that it is better to leave the user able to input more freely as you suggested :)

PR #278, opened 2018-08-24
Allow Interval to be restarted after max_interval was 0

### LINKED ISSUE #266: Can't dynamically update `max_intervals`?
I'm trying to use `dcc.Interval` `max_interval` property to create "start" and "stop" buttons that turn off the `dcc.Interval`. It seems like updating `max_interval` doesn't end up restarting the interval.

Here's a simple reproducable example:
```python
import dash

app = dash.Dash()

app.layout = html.Div([
    dcc.Interval(
        id='interval',
        interval=100,
        n_intervals=0,
        max_intervals=0
    ),

    html.Button('Start', id='start', n_clicks_timestamp=-1),
    html.Button('Stop', id='stop', n_clicks_timestamp=-1),

    html.Div(id='output')

])


@app.callback(
    Output('interval', 'max_intervals'),
    [Input('start', 'n_clicks_timestamp'),
     Input('stop', 'n_clicks_timestamp')])
def start_stop(start, stop):
    if start > stop:
        print('Returning -1')
        return -1
    else:
        print('Returning 0')
        return 0


@app.callback(Output('output', 'children'), [Input('interval', 'n_intervals')])
def display_data(n_intervals):
    return 'Updated {}'.format(n_intervals)


if __name__ == '__main__':
    app.run_server(debug=True)
```

@valentijnnieman - Could you take a look? Thanks!

valentijnnieman:
This fixes #266 by allowing the `Interval` component to be restarted by setting `max_interval` to anything but 0, after it was set to 0.


T4rk1n:
Looks good to me, but the tests are failing and they are on circle 1.

PR #284, opened 2018-08-28
Don't break Tabs if no Tab is selected

### LINKED ISSUE #282: initial tab content not rendering
In https://github.com/plotly/dash-docs/pull/152, I updated the version of `dash-core-components` to include the latest tab fixes. The issue with the 2nd tab loading by default instead of the 1st has been fixed 🎉  however it looks like there is a new issue with the tabs not loading their initial content.

See the first example here: 
image

If I click through the tabs, then the content shows up:
click-through-tabs

valentijnnieman:
It seems sometimes `selectedTab` is undefined, so this adds a little guard to prevent the `Tabs` from breaking if in some cases `selectedTab` is undefined when rendering. 

Fixes #282


chriddyp:
@valentijnnieman - is this ready for review?

valentijnnieman:
@chriddyp Yes, of course!

valentijnnieman:
Although I haven't updated these tests with the Selenium wait API yet.

valentijnnieman:
@T4rk1n Alright, I've made some changes so that both ways of working with the Tabs component (Dash-callback-controlled and internal-React-controlled) work a bit better, and have a default Tab selected. I've also fixed the tests & fixed merge conflicts. Thanks for the help! Could you please check if these changes work for you?

valentijnnieman:
Published as `0.28.2`.

PR #285, opened 2018-08-28
Fix #260, allow '{}' as Graph.figure

### LINKED ISSUE #260: `return {}` doesn't work for dcc.Graph?
See commuity discussion here: https://community.plot.ly/t/returning-an-empty-figure/12500/2

rmarren1:


src/components/Graph.react.js
```diff
@@ -77,18 +77,7 @@ export default class PlotlyGraph extends Component {
...
-                pluck('type', figure.data),
-                ['candlestick', 'ohlc']).length
-            ) {
-                PlotMethod = Plotly.newPlot;
-            } else {
-                PlotMethod = Plotly.react;
-            }
-
-            return PlotMethod(id, figure.data, figure.layout, config).then(
+            return Plotly.react(id, figure.data, figure.layout, config).then(
```
valentijnnieman: Could you clarify this for me? Isn't `Plotly.newPlot` needed?

rmarren1: I am not sure, I just reverted this commit: https://github.com/plotly/dash-core-components/commit/ea4e66caaa60a80156c6aa5437a02c819fca5997#diff-b2faf9645a4fcca0badb59ef2fba58a3

rmarren1: For context, that change was made because there was a bug in Plotly. That bug was fixed in https://github.com/plotly/plotly.js/pull/2561 but we never removed the fix from Dash.

rmarren1: https://github.com/plotly/dash-core-components/pull/170 I found this, this is why we call Plotly.react

chriddyp: For some more context:
- `newPlot` is needed for the _first plot_
- Subsequent plots can be done with `Plotly.react`. It was intended to the be faster way to handle updates but in reality we don't get many of the improvements because we are always providing new copies of data. Context here: https://github.com/plotly/plotly.js/pull/2341. If users added a `layout.datarevision`, then they would get some performance improvements but I haven't tested this and it's not really documented.
- So, while we don't get the main intended features of `Plotly.react`, we do get one very critical feature: the ability to repeatedly replot webgl plots: https://github.com/plotly/plotly.js/issues/2423
- Now, there were a lot of bugs with `Plotly.react` when it first came out. One of the bugs was that candlestick and ohlc charts didn't work with it. So, I called `newPlot` just for those chart types
- However, that bug has been fixed. So, we should standardize on all chart types using `Plotly.react`. It also might fix an issue that a community user was seeing in https://github.com/plotly/dash/issues/355#issuecomment-416429064

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.28.1]
+### Fixed
+- Fix bug where front-end error was thrown when setting `Graph.figure = {}` (fixes [#260]).
```
chriddyp: Could you add a `### Changed` that mentions that "candlestick and OHLC charts are now plotted using the Plotly.react method instead of the Plotly.newPlot method."

rmarren1:
Made all those changes

src/components/Graph.react.js
valentijnnieman: Ok, great! Thanks for clarifying!

valentijnnieman:
Cool! 💃

chriddyp:
💃

rmarren1:
released at `dash-core-components==0.28.1`

PR #289, opened 2018-08-30
Components with schema

rmarren1:
This PR is dependent on the validation PR https://github.com/plotly/dash/pull/452, and consists of the Dash Core Components generated with schemas, some tests, and some modifications to DCC prop-types to make validation work.


dash_core_components/Slider.py
```diff
@@ -54,12 +56,13 @@ def __init__(self, id=Component.UNDEFINED, marks=Component.UNDEFINED, value=Comp
         _explicit_args = kwargs.pop('_explicit_args')
         _locals = locals()
         _locals.update(kwargs)  # For wildcard attrs
-        args = {k: _locals[k] for k in _explicit_args if k != 'children'}
+        args = {k: _locals[k] for k in _explicit_args}
```
rmarren1: This lets you make the `children` prop `required`.

src/components/ConfirmDialog.react.js
```diff
@@ -99,7 +99,7 @@ ConfirmDialog.propTypes = {
     /**
      *  Set to true to send the ConfirmDialog.
      */
-    displayed: PropTypes.bool,
+    displayed: PropTypes.oneOfType([PropTypes.oneOf([null]), PropTypes.bool]),
```
rmarren1: Necessary because of [this](https://github.com/plotly/dash-core-components/blob/b7c2e4a3d35b53e0e6bd34f25610319fe50804f1/test/test_integration.py#L1059-L1063) test, since it will set `displayed=None`. Or, should we change _that_ test and keep this as only `PropTypes.bool`

src/components/DatePickerRange.react.js
```diff
@@ -173,36 +173,36 @@ DatePickerRange.propTypes = {
      * Accepts datetime.datetime objects or strings
      * in the format 'YYYY-MM-DD'
      */
-    start_date: PropTypes.string,
+    start_date: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(Date)]),
```
rmarren1: Technically a Python `datetime` object is stored, which is automatically serialized to a string; however, validation occurs before serialization, so we need this so the validation `schema` is automatically generated to accept `datetime` objects.

src/components/Graph.react.js
```diff
@@ -8,7 +8,7 @@ const filterEventData = (gd, eventData, event) => {
     if (contains(event, ['click', 'hover', 'selected'])) {
         const points = [];
 
-        if (isNil(eventData)) {
+        if (isNil(eventData) || isNil(eventData.points)) {
```
rmarren1: Was throwing front-end error is `eventData.points` was undefined.

src/components/Graph.react.js
```diff
@@ -74,7 +74,10 @@ export default class PlotlyGraph extends Component {
         const {id, figure, animate, animation_options, config} = props;
         const gd = document.getElementById(id);
 
-        if (animate && this._hasPlotted && figure.data.length === gd.data.length) {
+        if (animate && this._hasPlotted &&
+            !isNil(figure) && !isNil(gd) &&
```
rmarren1: Was throwing front-end errors.

src/components/Location.react.js
```diff
@@ -116,7 +116,10 @@ Location.propTypes = {
     id: PropTypes.string.isRequired,
 
     /** pathname in window.location - e.g., "/my/full/pathname" */
-    pathname: PropTypes.string,
+    pathname: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.oneOf([null])
```
rmarren1: Added because of [this](https://github.com/plotly/dash-core-components/blob/b7c2e4a3d35b53e0e6bd34f25610319fe50804f1/test/test_integration.py#L700-L710) test, which results in an initial callback setting `pathname=None`. We can change the test instead.

src/components/Slider.react.js
```diff
@@ -52,8 +52,8 @@ Slider.propTypes = {
      * the value should be an object which
      * contains style and label properties.
      */
-    marks: PropTypes.shape({
-        number: PropTypes.oneOfType([
+    marks: PropTypes.objectOf(
```
rmarren1: This just had the wrong schema initially.

src/components/Tabs.react.js
```diff
@@ -214,7 +214,7 @@ export default class Tabs extends Component {
             selectedTab = this.props.children.filter(child => {
                 return child.props.children.props.value === this.state.selected;
             });
-            if ('props' in selectedTab[0]) {
+            if (selectedTab[0] && 'props' in selectedTab[0]) {
```
rmarren1: Was throwing front-end exceptions.

src/components/Tabs.react.js
```diff
@@ -359,7 +359,7 @@ Tabs.propTypes = {
     /**
      * Array that holds Tab components
      */
-    children: PropTypes.node,
+    children: PropTypes.node.isRequired,
```
rmarren1: Front-end exceptions when this is `None`.

Marc-Andre-Rivet:
@rmarren1 What's the status of this PR? Is this blocked or moot? If blocked, what can we do to help unblock it. Just doing sanity checks on PRs that have been outstanding for a while. Thanks!

rmarren1:
This is dependent on https://github.com/plotly/dash/pull/452 which is still awaiting approval. I'll update the top level to explain that! (It was linked to old version of the PR).

byronz:
Close this PR as we shifted the validation in [dash-renderer #100](https://github.com/plotly/dash-renderer/pull/100)

PR #293, opened 2018-09-06
Serve unminified bundle support

T4rk1n:
Add `build:js-dev` command and support in `_js_dist`.

plotly/dash#369


rmarren1:
:dancer: I had that percy issue before, I think it goes away when you rebase. Can you do that before merging to make sure?

T4rk1n:
Renamed the bundles plotly/dash#384

T4rk1n:
@valentijnnieman Can you look at the percy diff ? I rebased on master an hour ago and it looks like there is a regression in the tabs.

valentijnnieman:
@T4rk1n Looks like the Graph is rendered differently with the new version of plotly.js, looks better to me :-)

valentijnnieman:
@T4rk1n Percy is being a little bit difficult lately it seems... Feels like the snapshots are changing slightly every time it runs. Maybe if we replace the tests with the Selenium WaitAPI instead of the time.sleep() it will be more stable?

T4rk1n:
@valentijnnieman Yea, I think you can go ahead with the Selenium WaitAPI if you have the time.

PR #296, opened 2018-09-10
Update CHANGELOG.md

rmarren1:
@valentijnnieman this is correct, right? I think there was a typo


valentijnnieman:
@rmarren1 Yes, thanks for fixing! 💃

PR #299, opened 2018-09-12
Update npm scripts used for building, testing, etc.

valentijnnieman:
Hi! This is an update for all the npm scripts we're using.

TL;DR
We should automate linting and formatting. Formatting so that we don't have to ask contributors to use 4 spaces instead of 2 for indentation for example, and so that we ensure our code looks nice and clean. We should run linting locally before the build step, so we prevent linting errors happening in CircleCI (which takes more time). I personally want to automate these things, so I don't have to spend time thinking about it.

More detailed:
- it removes the old `build-dev`, `build-dist` and `copy-lib` scripts that no longer work after updating our webpack config and adds the new `build:js` and `build:py` scripts.
- it also adds some dev build options, as well as a new `build:watch` script. This I personally find very handy when making changes to the code, and wanting to test them in an actual Dash app. This watcher just runs `build:all` when files have changed.
- it removes lingering webpack & babel config files that are no longer used, as well as 
the Archetype dependency. 
- it changes the `test` script to run our python tests instead of linting (and changes the CircleCI command)
- it adds a new `lint` command
- it adds a new `format` command that runs Prettier
- ~~It updates the `prepublish` hook with running formatting first, then linting, then building (formatting and linting might not be needed here / might be overkill, depending on where we would run it)~~
- It runs formatting and linting when building JS, this is handy for catching linting errors automatically upon building, especially when using the new `build:watch`.
- ~~it adds the `husky` dependency, which allows for git hooks. I've set up the `pre-push` hook, which fires before every `git push`.~~
- it runs formatting and linting on CircleCI, before the build step.
- it adds a new `scripts` folder with a `publish.js` script, that can be called using `npm run publish-all`, and it should get the name & version number from `package.json`, and publish to both NPM and PyPi
- it fixes the Demo app and the `npm start` command
- it updates the Readme with relevant scripts

Right now the publish script does `npm publish`, which isn't enough if you've enabled 2FA. Should we make 2FA the default, i.e. update it to run `npm publish --otp`?

~~What I would like to discuss most is the `pre-push` hook. I very much like automating stuff, because I know myself and know that I sometimes forget stuff. The tests take quite a while to finish however, so I would love to know if you think it is unnecessary to run tests before every `git push`. I know that some other big open source projects like `apollo` use this flow, and I know that it never hurts to run tests locally before pushing to Github. I also think it is a good idea to `format` and `lint` before pushing, so that the code is clean and readable in a PR.~~

~~Another point of discussion are the publish scripts `publish-all` and `publish-pypi`. I've left those as-is but I know that they've never really worked for me - I always just run `prepublish` first, then `npm publish` and then `twine upload dist/***` what are your thoughts on this?~~

@plotly/dash


chriddyp:
> I know that it never hurts to run tests locally before pushing to Github

In test driven development, to fix a bug I often first write a test that demonstrates the bug (the test should necessarily fail) and then I write the fix in another commit to demonstrate that the bug was fixed. If tests have to pass before I push, then it breaks this development pattern.

However, having `format` in the `githook` works for me.

> it adds a new lint command
> it adds a new format command that runs Prettier

Very nice. Is there any case where `format` runs but `lint` fails? If not, then can we _only_ have `format` and get rid of `lint`?

> It updates the prepublish hook with running formatting first, then linting, then building (formatting and linting might not be needed here / might be overkill, depending on where we would run it)

If formatting and linting are in CircleCI, then we shouldn't need to run them before running `prepublish`. That is, since we publish off of a passing `master` branch, then the code should be properly formatted.

> It removes the old build-dev, build-dist and copy-lib scripts that no longer work after updating our webpack config and adds the new build:js and build:py scripts.

Can we also update the package.json with the versions of webpack (and remove the archetype). i.e. update it to match https://github.com/plotly/dash-component-boilerplate?

> It changes the test script to run our python tests (with a command that I always use - does this work for everyone?) instead of linting

For tests, see the commands that are in circle: https://github.com/plotly/dash-core-components/blob/29d9f1d366b64a2759fbfc74d184a0561288284d/.circleci/config.yml#L49-L54. We also used to have `pylint` in there but it might've gotten removed when @T4rk1n added circle 2 support?

chriddyp:
> Another point of discussion are the publish scripts publish-all and publish-pypi. I've left those as-is but I know that they've never really worked for me - I always just run prepublish first, then npm publish and then twine upload dist/*** what are your thoughts on this?

Yeah, I'd say let's remove these. I'm not sure how to get twine to upload the latest package, so I always run these manually:
```
git checkout master
git pull origin master
npm run prepublish
python setup.py sdist
npm run publish
twine upload dist/dash-core-components-0.25.1.tar.gz
git tag -a 'v0.25.1' -m 'v0.25.1'
git push origin master --follow-tags
```

Or perhaps we turn it into a script that substitutes the version number, so we run:
```
npm run publish-all 0.25.1
```
and it does all of those things, subsituting the version number where appropriate

valentijnnieman:
@chriddyp Yeah fair about test driven development. Running tests locally before every push does take a long time too, so I'll remove that for now. I'm actually not sure about if `lint` catches anything that `format` doesn't - my thinking was that format won't catch/format ESLint specific rules like `no-console`, but I'll test that out a bit more. Currently, there's no formatting in CircleCI that I can see - no use of Prettier anywhere that is. Maybe I'm missing something? I got the idea from this PR #277 where the code being pushed wasn't formatted.

I like the `npm run publish-all 0.25.1` (and good catch with the tags, keep forgetting that), I'll try to implement that too.

chriddyp:
> Currently, there's no formatting in CircleCI that I can see - no use of Prettier anywhere that is. Maybe I'm missing something?

Yeah, I don't think you're missing anything. We should add prettier, eslint, pylint, and flake8 to the circle config. pylint and flake8 used to be in there I believe.

T4rk1n:
> We also used to have pylint in there but it might've gotten removed when @T4rk1n added circle 2 support?

We only ran pylint on the package `__init__`, there is almost nothing in there that warrant running pylint. Also with the generated components class files it was failing. I wanted to run it on the tests, but it was failing with a really low score so I left it out.

eslint tests runs in the npm tests. I'm not really familiar with prettier but I added it and the configs from #274, I believe it checks the prettier rules while running eslint. I think formatting should be done by pre-commit hook or running the command locally.

Considering pre-commit hooks, in my experience, they don't work well with IDE, so it needs to be togglable.

> It removes the old build-dev, build-dist and copy-lib scripts that no longer work after updating our webpack config and adds the new build:js and build:py scripts.

I have also removed the old build command in #293.

> Another point of discussion are the publish scripts

I also don't use them, I do `npm publish --otp 2222` because of the 2 factor auth. And do `python setup.py sdist` then `twine upload dist/*`.

T4rk1n:
> won't catch/format ESLint specific rules like no-console, 

The no-console is in the eslintrc, it didnt catch it because you used `window.console.log`, it would of catch `console.log`. I think normally it is stripped out by webpack, but I updated webpack and maybe we're missing a config option.

bpostlethwaite:
I would prefer not to introduce Husky into our libraries. It is great for in-house application development where devs can agree on tooling. But in OSS where there isn't that level of control forcing git to automagically run when things happen can be dangerous in certain environments.

For example some IDEs configure git hooks already (like my emacs magit IDE)

valentijnnieman:
@T4rk1n I just used no-console as an example, in that ESLint will catch different errors than Prettier.

@bpostlethwaite Ah ok, that's interesting. I can just add a `npm run format` step then in CircleCI, if that works just as well (i.e. if that formats the pushed code so that it is formatted how we want it when viewing the changed code in Github, and not when finally merging into master)

I've also removed the old webpack and babel configs, as well as the archetype from deps, since we've upgraded webpack and are no longer using them.

valentijnnieman:
I'm not sure what's causing the Percy diffs - I guess formatting the `Dropdown` code changed something, but I don't see what could've caused the difference in rendering. This is what's changed: https://github.com/plotly/dash-core-components/pull/299/files#diff-9022a6ae20c263b738f1e98ca94b74e8

valentijnnieman:
I've made some changes. I still think it's a good idea to automatically run the formatter and the linter locally, mostly because having the tests fail on CircleCI because you forgot a semicolon somewhere is a bit of a hassle, and might as well run formatting too while you're at it, so that if newcomers make changes we don't have to comment on their coding style, i.e. that we want 4 spaces instead of 2 for indentation. I'm now running formatting and linting as steps before building, when running `build:js`. 

I've also updated CircleCI to run formatting and linting before building, which is maybe unnecessary if we're already doing that in `build:js`, but it only takes a second and it's good to double check. I've also added a `build:watch` script, which is handy when making changes you want to check out in an actual Dash app.

valentijnnieman:
I've also added a new `scripts` folder with a `publish.js` script, that can be called using `npm run publish-all`, and it should get the name & version number from `package.json`, and publish to both NPM and PyPi. I think this should work, but I'm a little bit afraid to test it! I don't want to publish to NPM or PyPi accidentally! :-)

valentijnnieman:
I noticed the `npm start` command and the Demo app was broken so I fixed that too - and I updated the `README` with the new scripts. One question I have is about installing it locally: do you need to run `python setup.py install` after every build, or is that not necessary? Is that something we should do in the `build:watch` command? This much more work than I thought by the way! But good to get this sorted I think, really should make maintaining this library more pleasant and hopefully encourage others to contribute!

T4rk1n:
Can't you test it with release candidate. I can't use it since I need to input my code from the authenticator with `--otp` param.

Formatting on ci is not necessary, it's not like you are gonna push from circle are you ? It should fail if it would formatted. When I tried to integrate `black` there was a `--check` argument I could pass that would fail the build if it would of formatted. 

For pre-commit hooks that will format and you can toggle on/off, you can use https://github.com/pre-commit/pre-commit. It install the hook locally from a config file, just include the config file and those that want to use it can install it.

valentijnnieman:
@T4rk1n Oh hmm good point about the `--otp`, maybe that should be the default. The point about formatting is that I want the code to be formatted no matter what - if you can toggle hooks on/off, then there are going to be cases where the code appears unformatted in a PR, thus creating discussion in reviews about missing semicolons, newline characters, tab spaces, etc, which I want to avoid (if we automate it we don't have to discuss it and we save time!) I would also prefer the code to format itself if there's a problem, and not fail because it wasn't formatted right. Although, because now formatting & linting also happen in the `build:js` step, you're right that it might not be necessary to format _again_ on CircleCI - it's just an extra step that takes a couple of ms, and ensures code quality, so maybe it is a good thing to have.

bpostlethwaite:
Personally I do not think it is a good idea to automatically run the linter and formatter locally. Or more specifically I do think it is a good idea but it should be up to the developer and their particular development environment to enable. Otherwise we are making some assumptions about workflow and tooling.

For example my IDE formats and lints code as I work so I can make a mess and it all self-heals as I go. I would be alarmed if edits I didn't make suddenly appeared in a commit. Especially if it made it more difficult to separate commits for formatting from commits containing logic changes.

If a PR is made that contains unformatted code we should point the developer to our Contributing Guide where we document the tools we use for linting and formatting and some popular Editor integrations.

valentijnnieman:
@bpostlethwaite I hear you, but we're making those assumptions anyway (ESLint and .eslintrc etc, Prettier and .prettierrc etc), and we want contributors to follow those rules, and we want our code to be formatted with those rules, so we're kind already deciding upon workflow and tooling (which I think is good!). In theory it shouldn't lead to problems with commits, seeing as it will only affect code that you're writing at that moment (because the rest should already be formatted), it will only make the code you're committing look a little different. The formatting won't fix any linting issues by the way, so it's just the "styling" of the code. As for the linter, how we're running it now, it's not fixing anything by itself either, so it's just another check before building the code that there are no issues. It also won't affect anything in your workflow, if you're already running Prettier and ESLint in your IDE, then great, the build step will not format anything nor throw a linting error. 

What are your feelings on running Prettier on Circle?

bpostlethwaite:
We should run our tooling in CI in verify mode. And fail tests if it doesn't pass.

FWIW I am fine with Prettier and have been using it for a long time but what I am not in favour of is adding pre-commit hooks or similar bindings between code editing and code versioning. 

To me it is the job of developers, not library maintainers, to connect their development tools into the process they use for submitting code. Library maintainers can mandate that the code is formatted a certain way (in this case the spec is prettier and CI is the enforcing mechanism) but they shouldn't need  to dictate on _how_ to do it (of course documenting recommended ways are useful).

Not a big deal, it's just that I am somewhat biased as my unusual dev environment (emacs magit and lots of lisp stuff) inevitably clashes with projects that build in a lot of auto-magic stuff in the name of making things easier. 

I'll leave it at that.

chriddyp:
> Not a big deal, it's just that I am somewhat biased as my unusual dev environment (emacs magit and lots of lisp stuff) inevitably clashes with projects that build in a lot of auto-magic stuff in the name of making things easier.

I'd say that Ben is a representative developer, so if he feels like this is getting in the way then we should respect that.

Running prettier implicitly does sound like it would lead to surprising behaviour. If I'm editing code with the watcher running and the code is changing before my eyes when I'm saving, then I would find that surprising and I could imagine that affecting my flow. For example, if I'm writing a large new function, or even a new component, then I will want to write code freely without any distractions or without the lines changing _as I'm writing them_.

valentijnnieman:
Ok! I'll remove it & remove the format command from Circle. Sorry if I rubbed you the wrong way, just trying to improve the developing experience on this repo.

bpostlethwaite:
Not at all! It's good to have these conversations in public. There isn't a right way to do these things. 

We want to build the best developer experience possible. Does that mean automagic things for developers or providing flexibility and clear guidelines? Sometimes it's a bit of both.

.babelrc
```diff
@@ -1,6 +1,4 @@
 {
-    "extends": "./node_modules/dash-components-archetype/config/babel/babelrc",
-    "plugins": [
-        "styled-jsx/babel"
-      ]
-}
+    "presets": ["env", "react"],
+    "plugins": ["transform-object-rest-spread", "styled-jsx/babel"]
```
chriddyp: In our `dash-component-boilerplate`, we don't have this `plugins` anymore. Do we still need it? If so, should we add it to the `dash-component-boilerplate`?

dash_core_components/package.json
```diff
@@ -8,64 +8,64 @@
...
+    "test": "python -m unittest test.test_integration",
+    "format": "prettier --config .prettierrc --write src/**/*.js",
     "uninstall-local": "pip uninstall dash-core-components -y",
     "build:js": "webpack --mode production",
-    "build:py": "node ./extract-meta src/components > dash_core_components/metadata.json && cp package.json dash_core_components && npm run generate-python-classes"
+    "build:js-dev": "webpack --mode development",
+    "build:py": "node ./extract-meta src/components > dash_core_components/metadata.json && cp package.json dash_core_components && npm run generate-python-classes",
+    "build:all": "npm run build:js && npm run build:py",
+    "build:all-dev": "npm run build:js-dev && npm run build:py",
+    "build:watch": "watch 'npm run build:all' src"
```
chriddyp: should this be `build:all-dev` since we'd most likely use the `watch` while we're in development?

.circleci/config.yml
```diff
@@ -34,17 +34,18 @@ jobs:
           paths:
               - "venv"
               - "node_modules"
+
+      - run:
+          name: npm run lint
+          command: npm run lint
+
```
chriddyp: and `prettier --list-different` would be another `run` step here?

chriddyp:
This is looking really good. I made a couple of comments, once those are resolved, :dancer: from me.

After this is merged, could we make the same set of changes in `dash-html-components` and in `dash-component-boilerplate`?

chriddyp:
> Should we make 2FA the default, i.e. update it to run npm publish --otp?

I'm 👍  with that

.babelrc
valentijnnieman: We definitely need `styled-jsx/babel` here because of it's dependency in the `Tabs` component. `transform-object-rest-spread` is also needed if we want to use the [object spread operator](https://babeljs.io/docs/en/babel-plugin-transform-object-rest-spread/)

dash_core_components/package.json
valentijnnieman: Yeah good point, I wasn't sure about that. Having `build:js-dev` has some value if you just want to run the JS side (although, what does the `development` flag actually do for us here? It's not watching anything or rebuilding or hot-reloading etc) but I guess `build:all-dev` doesn't really have that much value. I wondered however, if `build:all-dev` makes it clear enough that you're running a watcher, as opposed to `build:watch` makes it pretty clear that you are. My suggestion would be to drop the `build:all-dev` altogether, and keep `build:watch`.

.circleci/config.yml
valentijnnieman: We could do that. That will mean that, if you haven't formatted properly, the build will fail, right? In that case, we're still running Prettier here and enforcing formatting - wouldn't it be better/nicer/friendlier to, instead of having the build fail (and forcing users to run formatting themselves), auto-format it in this step?

scripts/publish.js
```diff
@@ -0,0 +1,24 @@
...
+
+console.log(`Publishing version ${version} of ${name} to NPM & PyPi\n`);
+
+console.log('>', 'python setup.py sdist');
+execSh([
+        'npm publish',
+        `python setup.py sdist`,
+        `twine upload dist/${name}-${version}.tar.gz`,
+        `git tag -a 'v${version}' -m 'v${version}'`,
+        `git push origin master --follow-tags`
```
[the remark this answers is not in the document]
valentijnnieman: Oh yes, very good points! I'll look into it some more.

dash_core_components/package.json
T4rk1n: I use `build:js-dev` in #293, also added to the prepublish command because we need the dev bundle in the package.

valentijnnieman:
@bpostlethwaite Could you review as well? Thanks :)

.circleci/config.yml
chriddyp: > That will mean that, if you haven't formatted properly, the build will fail, 

Correct, similar to the `lint` step.

> auto-format it in this step?

How would it get added to the code though? Would CircleCI commit it back to this branch? I'd rather that CI didn't make any changes to the code, that the user is forced to make the changes.

valentijnnieman: I thought that code pushed to Github would pass through CircleCI first, so whatever changes happen in CircleCI would be reflected on Github. But now that I think about it, I guess that's not how it works. Ok! I'll add the `--list-different` command.

chriddyp: Yeah, CircleCI just runs a copy of the code in the repo at each commit, CircleCI doesn't write back to GitHub at all (unless you program to do that explicitly)

bpostlethwaite: Yes and that would be too magical. I would expect the code I write to be the code that ends up in Github :)

bpostlethwaite:
What are the Percy visual diffs about though?

T4rk1n:
There are regressions in the percy diff, the checklist seems broken.

rmarren1:
I think the tabs screenshots are a bit finicky, I got that `Tabs with Graph - clicked tab 1 (graph should not resize)`  regression before and have also seen it opposite (graph 1 in tab 2).

valentijnnieman:
@rmarren1 Yeah, the test isn't great for that one. I rewrote it (and am merging it now) in #306

valentijnnieman:
@T4rk1n @bpostlethwaite Running the code in the test in a Dash app looks fine to me, I have no idea what's up with those Percy diffs. 
screen shot 2018-09-21 at 4 19 12 pm

T4rk1n:
@valentijnnieman I confirm the same, working normally in a dash app.

valentijnnieman:
Perhaps the changes in #312 will give us some better insights into what's going on here.

T4rk1n:
@valentijnnieman Can we change the webpack config like @bpostlethwaite did here https://github.com/plotly/dash-component-boilerplate/pull/14#discussion_r221983392 ?

PR #303, opened 2018-09-13
fixes plotly/dash-core_components#292 Input type “number” increment bug

Bachibouzouk:
added a test for the disfunction of the increment buttons on the input when `type` is `number` and when the prop `min` is set to a decimal value, the test fails in both increment/decrement cases

### DIFF
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -422,6 +422,70 @@ class Tests(IntegrationTests):
 
         self.snapshot('gallery - text input')
 
+    def test_num_input_up_increment(self):
+        app = dash.Dash(__name__)
+
+        step_size = 0.5
+
+        app.layout = html.Div([
+            html.Div(id='waitfor'),
+            html.Label('Upload'),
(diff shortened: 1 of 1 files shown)


Bachibouzouk:
soThe `<input type="number" />` component should have the stepUp and stepDown method according to [this webpage](https://www.w3schools.com/jsref/dom_obj_number.asp). I managed to trigger the function by using the `ref="dccInput"` prop and then Calling `this.refs.dccInput.setpUp(10)` for example. I could nevertheless not successfully replace this function with another one by typing `this.refs.dccInput=dccstepUp`

a `console.log(this.refs.dccInput)` displays that `dccInput : input#my-input.my-input-wrapper` has a `HTMLInputElementPrototype` object which itself contains `{ stepUp: stepUp(), stepDown: stepDown() ...}` 
so maybe we could manage to override the stepUp function there, I will try more later, just wanted to describe what I've done so far.

PR #306, opened 2018-09-18
Fix Tabs and Tabs tests

### LINKED ISSUE #300: console.log in dcc.Tabs
There is console log in the Tabs component render, it should be removed.

https://github.com/plotly/dash-core-components/blob/29d9f1d366b64a2759fbfc74d184a0561288284d/src/components/Tabs.react.js#L171-L184

valentijnnieman:
Looks like I made a mistake when merging in master in previous PR #279 , and the fix wasn't included (`clone(figure.layout)`). I've put it back in, as well as removed the console.log, so this closes #300 too. Apologies, I won't merge & publish when I have a fever and should've been in bed anymore!


test/test_integration.py
```diff
@@ -629,6 +638,11 @@ def render_content(tab):
         self.snapshot("Tabs with Graph - initial (graph should not resize)")
         tab_two.click()
 
+        # wait for content to be loaded after clicking
+        WebDriverWait(self.driver, 10).until(
+            EC.visibility_of_element_located((By.ID, "tabs-content-example"))
```
chriddyp: should you be looking for data that's actually inside the tab rather than the tab container itself? We might even want to wait for the graph to be rendered. i.e. something like `#graph-`-tabs .svg-container`. Similarly below when you switch tabs (`#graph-2-tabs .svg-container`)

PR #308, opened 2018-09-20
Fix external_url bundles.

T4rk1n:
Change the name of the external_bundle, fix broken unpkg 0.30.0.

plotly/dash#369


chriddyp:
💃

PR #316, opened 2018-10-02
Table of contents component

T4rk1n:
# Table of contents component

- Generate a table of content from the children of an element.
- Regenerate the table on content change using a MutationObserver.

@chriddyp asked for a table of contents in dash docs plotly/dash-docs#212. I remember I had made a gist with a table of content builder from markdown. I adapted the gist to a dash component.

https://gist.github.com/T4rk1n/9a44e6d0a2da88cd3b54d3328df29bb8

### Props:

- `id`
- `content_selector`, The css selector to build the table of contents from it's children.
- `headings` The tag names that will be included in the table of contents.

### Basic usage:

```python
import dash

app = dash.Dash(__name__)

md = dcc.Markdown('''
# level one

content

## level two

content

### level 2-1

content

### level 2-2

content

### level 2-3

content

## level two-two

### level three

content

#### level four

content

##### level five

content
'''
)

app.layout = html.Div([
    dcc.TableOfContents(content_selector='#content'),
    html.Div(id='content'),
    html.Button('fill', id='fill')
])


@app.callback(Output('content', 'children'), [Input('fill', 'n_clicks')])
def on_fill(n_clicks):
    if n_clicks is None:
        raise PreventUpdate

    return md
```

### DIFF
--- a/src/components/TableOfContent.react.js
+++ b/src/components/TableOfContent.react.js
@@ -1,0 +1,91 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const promiseWrap = (func, options={rejectNull: false}) => new Promise((resolve, reject) => {
+    const { rejectNull } = options;
+    let result;
+    try {
+        result = func();
+    } catch(e) {
+        reject(e);
+    }
+    if (rejectNull && !result) {reject('Expected promise result is null');}
+    else {resolve(result);}
+});
+
+const selectElement = (selector) => promiseWrap(() => document.querySelector(selector));
+
+const buildToc = (contentSelector, options={headings: ['h1', 'h2', 'h3', 'h4', 'h5']}) => selectElement(contentSelector).then(element => {
+    const { headings } = options;
+    let currentNode;
+    // noinspection JSCheckFunctionSignatures
+    const nodeIterator = document.createNodeIterator(element, NodeFilter.SHOW_ELEMENT,
+        (node) => headings.map(h => node.nodeName.toLowerCase().match(h)).reduce((a, e) => a || e) ?
+            NodeFilter.FILTER_ACCEPT :
+            NodeFilter.FILTER_REJECT);
+    let lastNodeId = 0;
+    const children = [];
+    // eslint-disable-next-line no-cond-assign
+    while(currentNode = nodeIterator.nextNode()) {
+        const nodeId = 'toc-' + lastNodeId++;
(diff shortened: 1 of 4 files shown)


chriddyp:
This is pretty cool, and I could see this getting immediate use in dash-docs.

However, I'm not sure if we should include this in `dash-core-components`. It feels like a fairly specialized use case and I'd like to keep `dash-core-components` more generic, that way:
1. We have less to document. All dash core components need to get documented in dash-docs, and the more components that we have in there, the harder it is to navigate.
2. We have less to maintain. This component feels less "critical" than the other components that have, and might not need the same level of maintenance.
3. It's easier to describe what this component package offers. I've rejected PRs in the past for specialized components (e.g. ) under the principle that `dcc` should remain small, focused, and essential. Other components can be in other packages, and this type of decentralization is a key feature of Dash.

That being said, we could still publish this under a separate repo and use this in the dash-docs. It could be it's own package like `dash_table_of_components.TableOfContents` or we could make a "miscellaneous" package like `dash_misc_components.TableOfComponents` that we use at Plotly for random components, prototypes, and ideas. If any of them end up being popular, then we can fold them into the more "official" fold.

I'd also appreciate thoughts from @plotly/dash and @nicolaskruchten on this one.

cldougl:
I like the idea of publishing in a separate repo: dash_table_of_components.TableOfContents

T4rk1n:
I wasn't sure about putting it here, but I figured it goes along with the Markdown component because that was what the original gist was about (generating a table of content from markdown). I have some other components that I made, I don't mind putting them together in another repo.

T4rk1n:
Here's a few suggestions for a new repo name:

- `import dash_extra_components as extra`
- `import dash_misc_components as misc`
- `import dash_random_components as drc` 

I feel the SuggestionsInput would go there as well.
@bpostlethwaite You told me it was requested for dcc, think it would matter if it was in a new repo ?

cldougl:
I like `import dash_extra_components` or maybe `import dash_supplemental_components as dsc` (supplemental may be too long though).
@bpostlethwaite can you confirm that this doesn't need to be in `ddc` for a client or link to an issue where this was discussed

bpostlethwaite:
It wasn't requested for dcc. I probably wasn't clear about where it should go and recommended dcc just so we could start it somewhere. No problems hosting it and components like it in another repo

T4rk1n:
https://github.com/plotly/dash-extra-components/pull/4

PR #319, opened 2018-10-04
SuggestionInput component

T4rk1n:
Wrapper around [react-mentions](https://github.com/signavio/react-mentions)

Input that offers suggestions when a trigger character is typed.

Example:

```python
dcc.SuggestionInput(suggestions=[{
    'trigger': '$',
    'options': [
        {
            'id': 'bob',
            'display': 'bob',
        },
        {
            'id': 'bill',
            'display': 'bill'
        }
    ],
}])
```

Todo:

- [x] add integration test


bpostlethwaite:
Some things that came up during our offline discussion:

- Trigger mention even when there is a character directly before the trigger character
- More of a Dash pattern for styling
- stylable description string for options in `mention markup` so something like `__id__ - __description__`

T4rk1n:
Closing in favor of #329, a custom built SuggestionsInput.

PR #326, opened 2018-10-12
Add `n_blur/n_submit` to Input.

### LINKED ISSUE #230: Trying to add 'returnPressed' event for Input
(this is a follow-up of the issue I added on [dash project](https://github.com/plotly/dash/issues/283), I was not sure where to put it)

I tried to implement the feature I asked for, here is the diff on Input component:

```
.react.js`
diff --git a/src/components/Input.react.js b/src/components/Input.react.js
index b064581..860a4ce 100644
--- a/src/components/Input.react.js
+++ b/src/components/Input.react.js
@@ -43,6 +43,11 @@ export default class Input extends Component {
                 onBlur={() => {
                     if (fireEvent) fireEvent({event: 'blur'});
                 }}
+                onKeyPress={e => {
+                    if (e.charCode === 13) {
+                        if (fireEvent) fireEvent({event: 'returnPressed'});
+                    }
+                }}
                 value={value}
                 {...omit(['fireEvent', 'setProps', 'value'], this.props)}
             />
@@ -254,5 +259,5 @@ Input.propTypes = {
      */
     setProps: PropTypes.func,
 
-    dashEvents: PropTypes.oneOf(['blur', 'change'])
+    dashEvents: PropTypes.oneOf(['blur', 'change', 'returnPressed'])
 };
```

It seems pretty straightforward, however it does not work.

Here is the code I used to give it a try:

```
@app.callback(dash.dependencies.Output('output', 'children'),
    [],
    [dash.dependencies.State('cmdline', 'value')],
    [dash.dependencies.Event('cmdline', 'returnPressed')])
def execute_cmd(cmd):
    return [html.Pre(cmd)]
```
I get error messages on the web browser console:

 [Show/hide message details.] Error: Node does not exist: cmdline.change
 [Show/hide message details.] Error: Node does not exist: cmdline.blur

I am new to Dash -- any help would appreciated to understand the problem, in order for me
to be able to propose a pull request (if project maintainers are interested ?)

T4rk1n:
Add four props to dcc.Input:

- `n_submit`, The number of times `enter` was pressed when the component had focus.
- `n_submit_timestamp` The last timestamp enter was pressed.
- `n_blur`, The number of times the component lost focus.
- `n_blur_timestamp`, The last time the component lost focus.

resolves #230


src/components/Input.react.js
```diff
@@ -41,6 +41,20 @@ export default class Input extends Component {
                     if (fireEvent) {
                         fireEvent({event: 'blur'});
                     }
+                    if (setProps) {
```
valentijnnieman: A small improvement here could be that if `setProps` is set, the component doesn't really need to keep state itself, so `this.setState` should only be called if `setProps` is empty.

valentijnnieman: A small improvement here could be that if `setProps` is set, the component doesn't need to keep state itself, i.e. if `setProps` is set you don't need to fire `setState`. If you were to implement that, you'd also need to add that logic to `componentWillReceiveProps`. We had a small discussion about that in #331 - it won't cause any bugs like that here but it's still a small gain to not keep unnecessary state in this component. Up to you if you want to add that now, we could always add it later if you want to just merge this first.

T4rk1n: But it's not using the state.

valentijnnieman: Ah I was looking at line 29. That's something for another time then, feel free to merge!

T4rk1n: Is there a note about that in our react dev guide ? In the SuggestionsInput, I was using `componentWillReceiveProps` to setState on the value, but also setState before calling setProps on the value and the value in the props and state was always different so it lead to a bug where no suggestions beyond the first was registered so it a very valid concern to think about when developing dash component.

valentijnnieman: Exactly, like I mentioned above I found a bug with that recently too in #331. In the Input component I don't think it'll lead to any bugs though. But it's not in the React for Python devs guide, no. Would be good to add that!

dash_core_components/Location.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, id=Component.REQUIRED, pathname=Component.UNDEFINED, search=C
         _locals.update(kwargs)  # For wildcard attrs
         args = {k: _locals[k] for k in _explicit_args if k != 'children'}
 
-        for k in ['id']:
+        for k in [u'id']:
```
Akronix: why did you do this?

dash_core_components/Store.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, id=Component.REQUIRED, storage_type=Component.UNDEFINED, data
         _locals.update(kwargs)  # For wildcard attrs
         args = {k: _locals[k] for k in _explicit_args if k != 'children'}
 
-        for k in ['id']:
+        for k in [u'id']:
```
Akronix: why did you do this?

T4rk1n: You commenting on auto generated code, I didn't do that.

Akronix: Ah ok.

PR #329, opened 2018-10-15
Add SuggestionsInput component.

T4rk1n:
**Available in version:**

`$ pip install dash-core-component==0.36.0rc1`

Shows a menu with suggestions when a trigger is entered on the input/textarea,

Example:
```python
import dash
import flask

from dash.dependencies import Output, Input, State

app = dash.Dash(__name__)

suggestion_list = [
    'one',
    'two',
    'three',
    'five',
]

callback_suggestions = [
    'callback',
    'react',
    'dash'
]


app.layout = html.Div([
    dcc.SuggestionsInput(
        id='suggestions',
        suggestions=[
            {
                'trigger': '$',
                'options': [
                    {
                        'value': 'Sylvester Stallone',
                        'display': 'Rocky',
                        'description': 'Bob description'
                    },
                    {
                        'value': 'Arnold',
                        'display': 'Terminator',
                        'description': 'Arnold description'
                    }
                ],
            },
            {
                'trigger': '#',
                # This route will be used to retrieve the options.
                'suggestion_route': '/suggestions'
            },
            {
                'trigger': '@',
                'options': []
            }
        ],
        multi_line=True,
        # Style for the component single suggestions.
        suggestion_style={'padding': '0.5rem'},
        # Style for when a suggestion is selected.
        suggestion_selected_style={'backgroundColor': 'green', 'color': 'white'},
        # Style for the modal list of suggestions.
        suggestions_style={
            'backgroundColor': '#dcdcdc',
            'border': 'solid black 1px',
        },
        include_trigger=True
    ),
    html.Br(),
    dcc.SuggestionsInput(
        id='triggerless',
        suggestions=[
            {
                'trigger': '',
                'options': [
                    {'value': 'Rambo'},
                    {'value': 'Alien'},
                    {'value': 'Terminator'},
                    {'value': 'Predator'}
                ]
            }
        ],
        triggerless=True,
    ),
])


# Backend powered suggestions
@app.server.route('/suggestions')
def route_suggestions():
    payload = flask.request.args

    captured = payload.get('captured')
    if not captured:
        # Return the whole list since it's the first capture.
        return flask.jsonify([{'value': x} for x in suggestion_list])

    return flask.jsonify([
        {'value': x} for x in suggestion_list if captured in x
    ])


# Callback powered suggestions
@app.callback(Output('suggestions', 'filtered_options'),
              [Input('suggestions', 'captured')],
              [State('suggestions', 'current_trigger')])
def on_captured(captured, current_trigger):
    if captured is None or current_trigger != '@':
        raise PreventUpdate

    return [{'value': x} for x in callback_suggestions if captured in x]


if __name__ == '__main__':
    app.run_server(port=10960, threaded=True, debug=True)
```

Styling the component can be accomplished with the style props or className. If you style with className, you might need to set !important on the props that set by default in inline-style.

- `suggestions_className`/`suggestions_style`: the modal list of suggestions.
- `suggestion_className`/`suggestion_style`: single suggestion element in the list.
- `selected_suggestion_className`/`selected_suggestion_style`: when a suggestion is selected.

**Navigating/selecting the suggestions.**

When a trigger is entered in the input, a menu will pop up with the available suggestions. You can click on a suggestion or navigate the list with the arrow keys. Pressing tab or enter will confirm the selection and concat the associated suggestion value to the `value` prop.

### DIFF
--- a/src/components/SuggestionsInput.react.js
+++ b/src/components/SuggestionsInput.react.js
@@ -137,7 +137,6 @@ export default class SuggestionsInput extends React.Component {
         this.onKeyUp = this.onKeyUp.bind(this);
         this.onChange = this.onChange.bind(this);
         this.onSuggestion = this.onSuggestion.bind(this);
-        this.resetSuggestions = this.resetSuggestions.bind(this);
         this.preventTabNavigation = this.preventTabNavigation.bind(this);
         this._suggestionsRequest = null;
     }
--- a/src/components/SuggestionsInput.react.js
+++ b/src/components/SuggestionsInput.react.js
@@ -492,4 +492,6 @@ SuggestionsInput.propTypes = {
      * CSS class for a suggestion while it is selected.
      */
     suggestion_selected_className: PropTypes.string,
+    
+    setProps: PropTypes.any,
 };
--- a/src/components/SuggestionsInput.react.js
+++ b/src/components/SuggestionsInput.react.js
@@ -248,7 +248,7 @@ export default class SuggestionsInput extends React.Component {
                     }
                     break;
                 case 'Backspace':
-                    if (this.state.captured.length - 1 <= 0) {
+                    if (this.state.captured.length - 1 < 0) {
                         this.resetSuggestions();
                     } else {
                         this.updateSuggestions({
--- a/src/components/SuggestionsInput.react.js
+++ b/src/components/SuggestionsInput.react.js
@@ -217,8 +217,10 @@ export default class SuggestionsInput extends React.Component {
             } else {
                 this.setFilteredOptions(trigger.options);
             }
-        } else if (currentTrigger) {
-            const trigger = triggers[currentTrigger];
+        } else if (currentTrigger || this.props.triggerless) {
+            const trigger = this.props.triggerless
+                ? this.props.suggestions[0]
+                : triggers[currentTrigger];
             const options = trigger.options;
             switch (e.key) {
                 case 'Enter':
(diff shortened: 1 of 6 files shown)


bpostlethwaite:
Something funky with the triggerless trigger. You pointed out this commit https://github.com/plotly/dash-core-components/pull/329/commits/13fd12884f8bdad21f408eb77eeea9039a9153da

src/components/SuggestionsInput.react.js
```diff
@@ -0,0 +1,598 @@
...
+const OptionsShape = PropTypes.shape({
+    value: PropTypes.string.isRequired,
+    display: PropTypes.string,
+    description: PropTypes.string,
+});
+
+SuggestionsInput.propTypes = {
+    id: PropTypes.string,
+    /**
+     * Css class for the container of the input and suggestions modal.
```
bpostlethwaite: capitals - CSS

src/components/SuggestionsInput.react.js
```diff
@@ -0,0 +1,598 @@
...
+     */
+    className: PropTypes.string,
+    /**
+     * Style object given to the topmost container.
```
bpostlethwaite: is topmost container the same node as the target above "container of the input and suggestions modal"? If so, let's make the descriptions of the target node the same.

Marc-Andre-Rivet:
@T4rk1n What's the status on this? Is there something blocking us from going forward? Temporarily or indefinitely? Just doing sanity check on outstanding PRs. Thanks.

T4rk1n:
This component and the table of contents is now for dash-extra-components:

https://github.com/plotly/dash-extra-components/pull/2

PR #337, opened 2018-10-19
Use selenium WebDriverWait in tests.

### LINKED ISSUE #336: Issue with a Tabs test case
https://github.com/plotly/dash-core-components/blob/b4ddfa592d4f4b6ff7056dbf25f43d5cc2c47c9a/test/test_integration.py#L580-L586

This portion of the test is not executing consistently. There is a moment where the state of the `#tabs-content` is `''` and is present on the screen before the initial callback is fired, so sometimes `''` is compared with `'Default selected Tab content 1'` and throws an error. I got this to work temporarily by putting `time.sleep(1)` after `self.startServer(app=app)`

T4rk1n:
- Refactor `Tests.wait_for_element_by_css_selector` and `Tests.wait_for_text_to_equal` to use the WebDriverWait api.
- Replace manual text assert with `wait_for_text_to_equal`

Close #336

### DIFF
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -39,35 +39,23 @@ from multiprocessing import Value
 # export PERCY_PROJECT=plotly/dash-integration-tests
 # export PERCY_TOKEN=...
 
+TIMEOUT = 20
+
 
 class Tests(IntegrationTests):
     def setUp(self):
         pass
 
     def wait_for_element_by_css_selector(self, selector):
-        start_time = time.time()
-        error = None
-        while time.time() < start_time + 20:
-            try:
-                return self.driver.find_element_by_css_selector(selector)
-            except Exception as e:
-                error = e
-                pass
-            time.sleep(0.25)
-        raise error
+        return WebDriverWait(self.driver, TIMEOUT).until(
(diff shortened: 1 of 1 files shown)


test/test_integration.py
```diff
@@ -1021,6 +1005,10 @@ def display_data(n_intervals):
         # get the output after it's stopped, it shouldn't be higher than before
         output_stopped = self.wait_for_element_by_css_selector('#output')
 
+        self.wait_for_text_to_equal("#output", output_stopped.text)
+
+        # This test logic is bad
+        # same element check for same text will always be true.
         self.assertEqual(output.text, output_stopped.text)
```
valentijnnieman: Can we not remove line `1012` then, since you added `wait_for_text_to_equal` on line `1008`?

T4rk1n: The wait_for_text I added is as bad, it check the text against itself. I think we should refactor the last few lines of that test.

valentijnnieman: Ok sounds good!

valentijnnieman: 💃

rmarren1:
This is great, fixed a bunch of stuff in the my validation branch

Issue #335, opened 2018-10-19
Tabs uses old PropTypes syntax


rmarren1:
These lines would break in React 16, can we just use the normal `PropTypes` way?
https://github.com/plotly/dash-core-components/blob/b4ddfa592d4f4b6ff7056dbf25f43d5cc2c47c9a/src/components/Tabs.react.js#L377-L379

@valentijnnieman

valentijnnieman:
Which is what exactly? I see the same syntax on all other components, what am I missing?

rmarren1:
Okay great, I was wondering if there was a functional difference between the two here since plain `PropTypes` was used for the other `Tabs` props. I'll just edit those

valentijnnieman:
I don't think so, it was something I copied over from somewhere, but you're right in that we should use the `prop-types` package for React 16 support! Thanks for making a PR :)

rmarren1:
https://github.com/plotly/dash-core-components/pull/338

PR #338, opened 2018-10-19
Change `React.PropTypes` to just `PropTypes` in `Tabs.children`.

rmarren1:
https://github.com/plotly/dash-core-components/issues/335


valentijnnieman:
Nice, thanks 💃 The Percy snapshot is weird, but unrelated!

rmarren1:
Yeah I get that diff a bunch, seems to be an innocuous race condition

PR #342, opened 2018-10-20
Add flake8 to CircleCI

jbampton:


jbampton:
Hey @rmarren1 or @T4rk1n  I think the CircleCI tests need to be restarted  👍 

Message was -> Error response from daemon: received unexpected HTTP status: 503 Service Unavailable

rmarren1:
pinging @valentijnnieman too.
I don't think there is any hand-written Python code in the `dash_html_components` module (unless any has been added recently??), in which case I don't think we need to lint the class files. Any formatting issues with these files should just be taken care of in the class template in `dash/development/base_component.py`, and all of the JS code from which the Python classes are derived are already linted.

Also, I know we do not lint tests in any of the Dash repos, although I am not completely opposed to doing this, but I'd like to hear other developers opinions before adding it.

jbampton:
Hey @rmarren1  yes I understand now about the generated Python files.  

I can remove the flake8 check for that folder and just keep the flake8 check for the `test` folder.

Also there is another pull request here -> https://github.com/plotly/dash/pull/352  which has been approved and I was adding linting support for the tests folder.

jbampton:
Hey @rmarren1 and @valentijnnieman flake8 is now just enabled for the `test` folder 👍 

Also pinging @T4rk1n referencing https://github.com/plotly/dash/pull/352

valentijnnieman:
Apologies for the late reply! Is `flake8` a linter or a formatter? It seems to be both. I am not sure on the use of running a formatter in circle - what we have for the JS code is formatting checks with Prettier that will cause Circle to fail if it detects formatting issues, and linting with ESLint that will cause Circle to fail if any linting errors. Does this do that too, or does this format the code as it goes into Circle? If the latter, that's not of much use, since the code isn't piped back from Circle to Github or anything.

jbampton:
@Marc-Andre-Rivet the tests are passing now.  Thanks :)

Marc-Andre-Rivet:
@jbampton Thanks for your contribution :)

jbampton:
Thank you @Marc-Andre-Rivet 👍

PR #348, opened 2018-10-26
Changed incorrect readonly prop of Input by right one: readOnly

Akronix:
the prop readonly of Input component is incorrect and therefore, it does not work. Replaced by correct one: readOnly.

### DIFF
--- a/src/components/Input.react.js
+++ b/src/components/Input.react.js
@@ -234,7 +234,7 @@ Input.propTypes = {
     /**
      * This attribute indicates that the user cannot modify the value of the control. The value of the attribute is irrelevant. If you need read-write access to the input value, do not add the "readonly" attribute. It is ignored if the value of the type attribute is hidden, range, color, checkbox, radio, file, or a button type (such as button or submit).
      */
-    readonly: PropTypes.string,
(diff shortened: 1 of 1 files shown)


T4rk1n:
@valentijnnieman The example in the react doc have it as `readOnly`.

https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes

Akronix:
Exactly. As @valentijnnieman pointed out, readonly is not recognized as a DOM attribute for the *React input element* (here we are using React or not rendering directly to a html tag). This is because of their camelCase convention:

> In React, all DOM properties and attributes (including event handlers) should be camelCased. For example, the HTML attribute tabindex corresponds to the attribute tabIndex in React. The exception is aria-* and data-* attributes, which should be lowercased. For example, you can keep aria-label as aria-label.

Source: https://reactjs.org/docs/dom-elements.html

Actually, if you try to use readonly attribute in this component (Input) it doesn't work, but it does work for [the Textarea component of dash-html-components](https://github.com/plotly/dash-html-components/blob/master/src/components/Textarea.react.js) which has this attribute as readOnly. That's how I found the bug.

PR #352, opened 2018-10-30
Loading component

valentijnnieman:
This is the new Loading component for the new Loading States api [issue 267](https://github.com/plotly/dash/issues/267). For now it has a basic default spinner. 

*Please note!*
I've had to make a small change to `extract-meta` to not pick up `.jsx` files, so that we can have `.jsx` files in `src/` that are not supposed to be mapped to Dash components. This is so that I could split up all the spinners into separate files, that we _don't want Dash to pick up as components_ - instead they should be JS only components that get used under the hood. There's probably a better fix here, let me know your thoughts please.

TODO:
- [x] Add more default spinners, and a prop for selecting which one to use (`loading_screen_style = 'basic' | 'overlay' | 'spinner' | 'empty-div' |` or something similar)
- [x] Support for full-screen spinners
- [x] Add props for styling the default spinner's colours and other
- [x] Add support for handling which prop caused loading time (once in renderer PR)
- [x] Allow for React components that don't get picked up by extract-meta, so we can have a file structure that allows separating of concerns.
- [ ] Look into adding support for custom loading components, with [this PR perhaps](https://github.com/plotly/dash-renderer/pull/92)

SPINNERS:
- Graph spinner : https://codepen.io/Chunkydory/pen/qQOrwB
- Cube spinner : https://codepen.io/Chunkydory/pen/aQdpPL

edit: I'm also showing the component and prop Name in the HTML, just to show off that that information is being passed. In the final version of this component, I'll remove it (or at least change how it looks)

edit2: I've fixed the Demo app and I've added a Loading Demo app that you can fire up to check out all the spinners. `npm start`!


src/components/Loading.react.js
```diff
@@ -0,0 +1,104 @@
...
+
+Loading.propTypes = {
+    id: PropTypes.string,
+
+    /**
+     * Array that holds components to render
+     */
+    children: React.PropTypes.oneOfType([
+        React.PropTypes.arrayOf(React.PropTypes.node),
+        React.PropTypes.node,
```
T4rk1n: Use PropTypes not React.PropTypes.

src/components/Loading.react.js
```diff
@@ -0,0 +1,104 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * A slider component with a single handle.
+ */
+export default class Loading extends Component {
+    constructor(props) {
+        super(props);
+    }
```
T4rk1n: :zap: Default constructor.

T4rk1n:
@valentijnnieman You wrap the component you want to have a loading with this loading component ?

valentijnnieman:
@bpostlethwaite I think you mentioned specific loading spinners for Plotly on Friday? If there are any that you think we should incorporate into this component, could you link them here? Thanks in advance :)

bpostlethwaite:
https://github.com/plotly/streambed/blob/master/shelly/webapp/static/webapp/images/plotly-logo-loop.gif

But this one doesn't actually seem to have looping enabled. You'll need to open a GIF editor and enable looping.

I am pretty sure we should edit this one in `streambed` to support looping as well...

bpostlethwaite:
If you create a loop version let me know and I'll PR that back into `streambed`

valentijnnieman:
Hmm it's very small and blurry. I don't think we should have that as a default loading spinner, would rather have something in like we have here now! Taken from http://tobiasahlin.com/spinkit/ .

valentijnnieman:
I made this https://codepen.io/Chunkydory/pen/qQOrwB which is a little bit campy! But I kind of like it. @cldougl also pointed out that this exists! https://codepen.io/doeg/pen/RWGoLR

valentijnnieman:
Here's the pen of the cube spinner I just added: https://codepen.io/Chunkydory/pen/aQdpPL

Marc-Andre-Rivet:
@valentijnnieman No specific suggestion, just a curated list of spinners
https://github.com/yangshun/awesome-spinners

dash_core_components/package.json
```diff
@@ -33,6 +34,7 @@
     "babel-preset-env": "^1.7.0",
     "babel-preset-react": "^6.24.1",
     "builder": "3.2.2",
+    "color": "^3.1.0",
```
valentijnnieman: New dependency used for darkening hex colors, used in the `CubeSpinner` component. It allows users to provide a color, and the component will figure out the appropriate darker colors that are used in the cube.

demo/LoadingDemo.react.js
```diff
@@ -0,0 +1,50 @@
...
+                <Loading status={status} type="graph" fullscreen={true} />
+            </Tab>
+            <Tab value='tab-7' label='Fullscreen Cube spinner'>
+                <Loading status={status} type="cube" fullscreen={true} />
+            </Tab>
+        </Tabs>
+    );
+};
+
+export default LoadingDemo;
```
valentijnnieman: New Demo app that shows off the spinners.

extract-meta
```diff
@@ -39,7 +39,7 @@ function parseFile(filepath) {
     const urlpath = filepath.split(path.sep).join('/');
     let src;
 
-    if (!['.jsx', '.js'].includes(path.extname(filepath))) {
+    if (!['.js'].includes(path.extname(filepath))) {
```
valentijnnieman: We currently do not have any `.jsx` files in our source code, so I figured they could be used in a way to denote components that _should not be picked up by extract-meta_, so that we can use them internally in JS. It's an easy solution to have the spinners, in this case, in separate files and not in one huge `Loading.react.js` file.

index.html
```diff
@@ -4,22 +4,13 @@
 <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/codemirror/5.0.0/codemirror.min.css" />
-    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/codemirror/5.0.0/theme/xq-light.min.css" />
-    <link rel="stylesheet" href="https://unpkg.com/react-select@1.0.0-rc.3/dist/react-select.css" />
-    <link rel="stylesheet" href="https://cdn.rawgit.com/chriddyp/abcbc02565dd495b676c3269240e09ca/raw/816de7d5c5d5626e3f3cac8e967070aa15da77e2/rc-slider.css"
-    />
```
valentijnnieman: These are no longer needed since we supply these CSS files ourselves in the source code (using `style-loader` etc)

index.html
```diff
@@ -4,22 +4,13 @@
...
 
 <body>
     <div id="react-demo-entry-point"></div>
 </body>
 
 <footer>
-    <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/codemirror/5.0.0/codemirror.min.js"></script>
-    <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/codemirror/5.0.0/mode/javascript/javascript.min.js"></script>
-    <script type="text/javascript" src="https://unpkg.com/react@15/dist/react.js"></script>
-    <script type="text/javascript" src="https://unpkg.com/react-dom@15/dist/react-dom.js"></script>
```
valentijnnieman: We're supplying React in the `devDependencies` now, so this is no longer needed.

webpack.config.js
```diff
@@ -55,7 +55,7 @@ module.exports = (env, argv) => {
         module: {
             rules: [
                 {
-                    test: /\.js$/,
+                    test: /\.jsx?$/,
```
valentijnnieman: `.jsx` files weren't being picked up in the Webpack configs, this will add both `.js` and `.jsx` files.

webpack.serve.config.js
```diff
@@ -1,7 +1,11 @@
 const config = require('./webpack.config.js');
+const path = require('path');
 
 config.entry = {main: './demo/index.js'};
-config.output = {filename: 'output.js'};
+config.output = {
+    filename: './output.js',
+    path: path.resolve(__dirname),
+};
```
valentijnnieman: Fixes the Demo app.

src/components/Loading/Loading.react.js
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Array that holds components to render
+     */
+    children: PropTypes.oneOfType([
+        PropTypes.arrayOf(PropTypes.node),
+        PropTypes.node,
+    ]),
+
+    type: PropTypes.oneOf(['graph', 'cube', 'circle', 'dot', 'default']),
```
T4rk1n: Need to add a description docstring here.

src/components/Loading/spinners/CircleSpinner.jsx
```diff
@@ -0,0 +1,182 @@
...
+import PropTypes from 'prop-types';
+
+const CircleSpinner = ({status, color, fullscreen, debug}) => {
+    let debugTitle;
+    if (debug) {
+        debugTitle = (
+            <h3 className="dash-loading-title">
+                Loading {status.component_name}
+                's {status.prop_name}
+            </h3>
```
T4rk1n: :smile_cat:

index.html
Marc-Andre-Rivet: So I dug through this because it was not obvious to me how the dependencies had changed based on the webpack.config.js and webpack.server.config.js changes in this PR -- they would not affect these script dependencies.

webpack.serve.config.js sets ```config.externals = undefined``` which makes webpack.config.js evaluate ```'externals' in overrides``` to true which does not set react, react-dom, etc. as defaults.

Not going to ask for changes as this is out of scope, just documenting my own confusion. I think this could be improved on later.

src/components/Input.react.js
```diff
@@ -63,13 +63,20 @@ export default class Input extends Component {
...
                     }
                     if (setProps) {
-                        const castValue =
-                            type === 'number' ? Number(value) : value;
-                        setProps({
-                            n_blur: this.props.n_blur + 1,
-                            n_blur_timestamp: new Date(),
-                            value: castValue,
-                        });
+                        if (debounce) {
```
Marc-Andre-Rivet: How is this debounce change related to the loading component?

src/components/Loading/Loading.react.js
```diff
@@ -0,0 +1,88 @@
...
+                case 'cube':
+                    return <CubeSpinner status={loading_state} color={color} debug={debug} fullscreen={fullscreen} />;
+                case 'circle':
+                    return <CircleSpinner status={loading_state} color={color} debug={debug} fullscreen={fullscreen} />;
+                case 'dot':
+                    return <DotSpinner status={loading_state} color={color} debug={debug} fullscreen={fullscreen} />;
+                default:
+                    return <DefaultSpinner status={loading_state} color={color} debug={debug} fullscreen={fullscreen} />;
+            }
+        }
```
Marc-Andre-Rivet: Since all spinners have the same props, you could simplify this
```
function getSpinner(type) {
    switch (type) {
        case 'graph':
            return GraphSpinner;
        case 'cube':
            return CubeSpinner;
        case 'circle':
            return CircleSpinner;
        case 'dot':
            return DotSpinner;
        default:
            return DefaultSpinner;
    }
}
```

```
const Spinner = getSpinner(this.props.type); // Capitalized for React's sake!

return <Spinner
    status={loading_state}
    color={color}
    debug={debug}
    fullscreen={fullscreen}
/>;
```

src/components/Loading/spinners/CircleSpinner.jsx
```diff
@@ -0,0 +1,182 @@
...
+                    }
+                      .dash-loading-title {
+                          text-align: center;
+                      }
+                    .sk-circle {
+                        margin: 1rem auto;
+                        width: 40px;
+                        height: 40px;
+                        position: relative;
+                    }
```
Marc-Andre-Rivet: Would suggest changing ```.sk-circle``` for ```.dash-spinner-container .sk-circle``` and similarly for other selectors and spinners.

src/components/Loading/spinners/CircleSpinner.jsx
```diff
@@ -0,0 +1,182 @@
+import React from 'react';
```
Marc-Andre-Rivet: From talking with @valentijnnieman some of these seem to be inspired from MIT licensed sources, I think we should add attribution to the code if only for future devs to know where they come from and for follow up purposes.

valentijnnieman: Yes, good call. Completely forgot - most of them come from https://github.com/plotly/dash-core-components/pull/352#issuecomment-437483287 so we should credit them. Just as comments in the source code?

Marc-Andre-Rivet: I'm fine with code comments except if we do explicit attributions normally through a Notice.txt or similar

T4rk1n:
Tested with the simple.py in dash-renderer, looking good.  Here's a few thing I'd like to see:

- [ ] integration test with diverse components wrapped in dcc.Loading
- [ ] percy snapshot for all the spinners.

Marc-Andre-Rivet:
@T4rk1n Not sure what percy snapshots for the spinners would look like -- is the execution time perfectly predictable or will the snapshots be slightly different every time due to execution unpredictability?

T4rk1n:
Just put a sleep in the callback, take the screenshot right after the callback fires.

T4rk1n:
Or maybe try to take the snapshot from inside the callback ?

Marc-Andre-Rivet:
That's exactly what I mean, that does not seem 100% predictable :)

Marc-Andre-Rivet:
I agree with integration tests -- I would say "various components" are not strictly required to start with except if there are multiple loading scenarios that differ greatly -- proving sanity with one component would be a good start.

T4rk1n:
If you sleep for 2 seconds in the callback and wait for 0.5 in the test it should be predicable. I would try snapshot from the callback.

T4rk1n:
Forget about the percy snapshot...

Marc-Andre-Rivet:
After discussing the integration tests situation with @valentijnnieman I think this can be committed without an integration test for now -- when dash-renderer will be updated after  https://github.com/plotly/dash-renderer/pull/93 we can write a meaningful integration test.. @valentijnnieman  Please open a followup issue for the tests.

valentijnnieman:
@chriddyp version `0.41.0rc1` should be up now.

valentijnnieman:
Released version `0.41.0rc2`, which fixes the indent error in `_imports_.py` that occurs on release candidate 1.

valentijnnieman:
Also noticed I didn't add `className` and `style` props yet, which might come in handy for some users. Added them in `0.41.0rc4`

RayPalmerTech:
Hello, I don't usually post on Github, so I didn't know exactly where to post this, but I am a big fan of Dash and currently experimenting with it. I also noticed this Loading core component release and am very happy with the ease of use & the look of it! 
However, I did bump into some issues when combining it with a hidden Div to store data in. Namely in the app I'm building, I'm doing an expensive SQL query and store that data in the Hidden Div.  Afterwards I update some graphs based on that data. The problem is that when I place the hidden Div within the dcc.Loading component (together with a graph), the loading gif only seems to work on page initialization and not anymore afterwards. I made a small code example that mimics this behavior  to clarify what I mean:

```
import json
import time

import dash

app.layout = html.Div([
    dcc.Dropdown(id='dropdown',
                 options=[{'label': i, 'value': i} for i in [1,2,3,4,5]],
                 value=1),
    dcc.Loading(id='text-content-and-hidden-div',
                 type='graph',
                 fullscreen=True,
                 children=[
                         html.Div(id='text-div'),
                         # Hidden div inside the app that stores the intermediate value
                         html.Div(id='intermediate-value', style={'display': 'none'})
                 ])
])

@app.callback(Output('intermediate-value', 'children'),
              [Input('dropdown', 'value')])
def calculate_new_text_and_store_in_hidden_div(value):
     # some expensive computing step normally (or in my case an SQL data fetch)
     text = str(value*10)
     time.sleep(3)
     return json.dumps(text)

@app.callback(Output('text-div', 'children'),
              [Input('intermediate-value', 'children')])
def update_text_in_div(jsonified_text):
    json_string = json.loads(jsonified_text)
    return json_string

if __name__ == '__main__':
    app.run_server(debug=True)
```

Am I doing something wrong in my code, is my view on how to use the Loading component wrong, or is this perhaps a bug? 

Thanks in advance and merry christmas!

Akronix:
@RayPalmerTech I think the proper place to ask this kind of questions would be in the forum: https://community.plot.ly/ using the Dash tag.

dash_core_components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dash-core-components",
-  "version": "0.40.0",
+  "version": "0.41.0-rc4",
```
Marc-Andre-Rivet: remove rc

Marc-Andre-Rivet:
Would like to see some improvements to the tests, specifically:
- testing for the presence of html content is not sufficient -- should test to see if the content corresponds to the loading or loaded state

Also, would like confirmation as to the issue experienced by user above (https://github.com/plotly/dash-core-components/pull/352#issuecomment-449851808)

test/unit/Loading.test.js
```diff
@@ -4,75 +4,79 @@ import {render} from 'enzyme';
...
+            <div>Child 1</div>
+            <div>Child 2</div>
+            <div>Child 3</div>
+        </Loading>
     );
 
-    expect(loading.html()).toBeDefined();
-});
\ No newline at end of file
+    expect(loading.html()).toMatchSnapshot("Loading with is_loading=true");
```
Marc-Andre-Rivet: A lot better. Although I'm concerned about test stability -- updating the styling or the loading component will systematically require updating the snapshots. Is there any gain vs. simply checking if the HTML contains `dash-default-spinner`?

valentijnnieman: The snapshots are pretty easy to update - running `npm run test-unit` and if there's a change in snapshots, will tell you to run something like `npm run test-unit -- -u` to update the snapshots. So the gain is that it's a bit easier to update if we were to change our UI. Another gain is of course that it tests if the output is exactly the way we want as opposed to testing if it has a class of `dash-default-spinner`.

valentijnnieman:
I've made a few changes that should hopefully fix the problems that users such as @RayPalmerTech  are having. The loading component now traverses its children and checks for any loading state, and displays a spinner accordingly. Here's a small example app:
```
# -*- coding: utf-8 -*-
import dash
import plotly.graph_objs as go
import time

app = dash.Dash(__name__)

app.layout = dcc.Loading(
    id="loading-1",
    children=[
        html.Div(id="output-1"),
        dcc.Input(id="input-1", value='Input triggers fullscreen spinner'),
        dcc.Loading(id="loading-2", children=[html.Div(id="output-2")], type="default"),
        dcc.Input(id="input-2", value='Input triggers local spinner'),
        html.Div(
            [
                dcc.Loading(
                    id="loading-3",
                    children=[html.Div([html.Div(id="output-3")])],
                    type="circle",
                ),
                dcc.Input(id="input-3", value='Input triggers nested spinner'),
            ]
        ),
    ],
    type="cube",
    fullscreen=True,
)


@app.callback(Output("output-1", "children"), [Input("input-1", "value")])
def input_triggers_fullscreen(value):
    time.sleep(1)
    return value


@app.callback(Output("output-2", "children"), [Input("input-2", "value")])
def input_triggers_spinner(value):
    time.sleep(1)
    return value


@app.callback(Output("output-3", "children"), [Input("input-3", "value")])
def input_triggers_nested(value):
    time.sleep(1)
    return value


if __name__ == "__main__":
    app.run_server(debug=False)
```

It is still slightly brittle (order of the children seem to matter, where if the output is below the input in the example above's layout, the fullscreen spinner doesn't seem to trigger) so I would like to do more testing and hopefully fix this before finally releasing.

valentijnnieman:
Released as prerelease `0.44.0rc2`

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.44.0] - 2019-02-13
+### Added
+- Loading component
```
Marc-Andre-Rivet: Unreleased, revert version change for final merge. Add issue link.

dash_core_components/CircleSpinner.py
```diff
@@ -0,0 +1,58 @@
+# AUTO GENERATED FILE - DO NOT EDIT
+
+from dash.development.base_component import Component, _explicitize_args
```
Marc-Andre-Rivet: The Loading component uses the various spinner flavours so it makes sense to have a component for each flavour, but I'm not sure exposing the various flavours as top-level dcc components makes sense.

Marc-Andre-Rivet: The Loading component uses the various spinner flavours so it makes sense to have a component for each flavour, but I'm not sure exposing the various flavours as top-level dcc components makes sense.

src/components/DatePickerRange.react.js
```diff
@@ -117,46 +117,63 @@ export default class DatePickerRange extends Component {
...
-                endDatePlaceholderText={end_date_placeholder_text}
-                firstDayOfWeek={first_day_of_week}
-                focusedInput={focusedInput}
-                initialVisibleMonth={() => {
-                    if (initial_visible_month) {
-                        return initial_visible_month;
-                    } else if (end_date && focusedInput === 'endDate') {
-                        return end_date;
-                    } else if (start_date && focusedInput === 'startDate') {
+            <div
```
Marc-Andre-Rivet: I have concerns over the impact these wrapper divs will have on the overall application display / flow.
These will default to `display: block` while the now-wrapped component may or may not have been that way.

What was the impact on the visual tests?

... Oh! Is see the default style above! As this been done for all comps?

src/components/Loading/Loading.react.js
```diff
@@ -0,0 +1,165 @@
...
+            className,
+            style,
+            fullscreen,
+            debug,
+            type,
+        } = this.props;
+
+        const initial_loading_state = R.isNil(loading_state)
+            ? false
+            : loading_state.is_loading;
```
Marc-Andre-Rivet: The first impression I got from reading this variable's name was that this had to do with initial vs. subsequent loading. A more precise name would clarify the meaning.

src/components/Loading/Loading.react.js
```diff
@@ -0,0 +1,165 @@
...
+         */
+        component_name: PropTypes.string,
+    }),
+};
+
+function getLoadingStateInChildren(children, initial_loading_state) {
+    if (!Array.isArray(children)) {
+        children = [children];
+    }
+    let isLoading = initial_loading_state;
```
Marc-Andre-Rivet: Is the new variable necessary with the `"no-param-reassign": ["off"],` rule above?

src/components/DatePickerRange.react.js
valentijnnieman: Just for these ones, because those were the only ones that showed visual changes.

dash_core_components/CircleSpinner.py
valentijnnieman: Fixed this by having a new `fragments/` folder. Our tools only look in `components/` for components to parse, so our spinners are now in `fragments/Loading/....` so that they won't get exposed.

src/components/Loading/Loading.react.js
valentijnnieman: In a way it is, because the `initial_loading_state` is the loading state of the Loading component itself. Any suggestions? Not sure what naming would be more precise.

Marc-Andre-Rivet: Would probably just make it `isLoading` like elsewhere

src/components/Loading.react.js
```diff
@@ -0,0 +1,165 @@
...
+            child.props &&
+            child.props.loading_state &&
+            child.props.loading_state.is_loading
+        ) {
+            isLoading = true;
+        }
+        if (child.props && child.props.children) {
+            isLoading = getLoadingStateInChildren(
+                child.props.children,
+                isLoading
```
Marc-Andre-Rivet: Passing the previous isLoading result seems to make the code more complicated for no gain. Would refactor it out.

valentijnnieman: The `isLoading` here is needed because we don't want to lose any loading state when we're recursing through `child.props.children` - I'm having a hard time refactoring that out at least.

valentijnnieman:
(Re) released as rc `0.43.0rc3` including fixes

src/components/Loading.react.js
```diff
@@ -0,0 +1,148 @@
...
+                    debug={debug}
+                    fullscreen={fullscreen}
+                />
+            );
+        }
+        if (
+            R.type(this.props.children) !== 'Object' ||
+            R.type(this.props.children) !== 'Function'
+        ) {
+            return <div className={className}>{this.props.children}</div>;
```
Marc-Andre-Rivet: We're not using the id passed through props for id+key of this component. Might cause similar problems as what we've seen with other comps

valentijnnieman: I don't think we're setting `key` anywhere in the components, so I don't think it'll cause problems. The `key` is set in dash-renderer, however. But, you're right - I should add the `id` prop here!

eromoe:
I have installed `0.44.0rc2` , but no loading shows, 

Is any wrong withinstalled  packages ? : 
```
dash==0.36.0  # The core dash backend
dash-html-components==0.13.5  # HTML components
dash-core-components==0.44.0rc2  # Supercharged components
dash-table==3.1.11  # Interactive DataTable component (new!)
dash-daq==0.1.0  # DAQ components (newly open-sourced!)
dash-bootstrap-components
```

valentijnnieman:
Hi @eromoe - the packages you'd need are: 
``` 
dash_html_components==0.14.0rc21
dash_renderer==0.18.0rc4
dash_core_components==0.43.0rc3
```
It looks like you're missing `dash_renderer==0.18.0rc4`!

Here's a [link to the forums](https://community.plot.ly/t/loading-states-and-loading-component/19650/10) with a more detailed example of how to use this.

Issue #173, opened 2018-03-12
Input 'number' with min and max not checking manual input


radekwlsk:
The `dcc.Input` component set to `type='number'` with `min` and `max` defined checks the value only when using the input's arrows. On manual input any value can be entered, only regexp for `number` is checked.

I think it should apply `min` when entered manually value is lower than allowed and `max` when it is greater than allowed. For example on focus lose or mouse click or even after some delay of not typing.

sobayed:
I agree. When I saw the min/max parameters I assumed that the user inputs would be checked in some way. Allowing any input anyway kind of defeats the purpose (or at least make the forced check configurable).

Austin-PAVrus:
Me too. Trying solve it by making something like:

@app.callback(Output('agentT', 'value'),
[Input('agentT', 'value')],
[State('agentT', 'min'),
State('agentT', 'max')])
def check_agentT_value(val,min,max):
	if val >= min or val<= max:
		raise PreventUpdate
	elif val<min:
		return min
	else:
		return max

leads me to a failure of layout loading with no logs in console...

Is there other ways to check the limits of a manual input? I have 46 numeric fields in my project...

valentijnnieman:
A fix is included in `0.37.0`!

PR #356, opened 2018-10-31
Input component updates

### LINKED ISSUE #169: dcc.Input bug with decimal values

### LINKED ISSUE #292: Input type “number” increment bug
https://community.plot.ly/t/input-type-number-increment-issue/13277

If Min is set to 0.0001 and one tries to increase the value from 1 to 2 it instead goes to 1 to 1.0001. The next number it goes to is 0.0001.

valentijnnieman:
edit:
This PR is becoming much bigger than expected, apologies in advance if the commit history or comments are not super clear.

This updates the `Input` component with a new `debounce` prop that determines when `setProps` is called, should fix #169. It also adds a bunch of unit tests, and fixes an issue where props coming from `dash-renderer` would overwrite the current state of the input. That last bug is more common than we think - components that use `state` as well as `setProps` get out of sync often.

Here's an example app with the new `debounce` prop:

```
import dash

app = dash.Dash()

app.layout = html.Div([
    dcc.Input(placeholder='Enter a value...',
        type='number',
        value=0,
        step=0.01,
        debounce=True,
        id='input_id',
        style={'float': 'left'}),
    html.Div(id='my-div')
])

@app.callback(
    Output(component_id='my-div', component_property='children'),
    [Input(component_id='input_id', component_property='value')]
)
def update_output_div(input_value):
    return 'You\'ve entered (+1) {}'.format(input_value+1)

if __name__ == '__main__':
    app.run_server()
```

It also fixes #292 by fixing the `step` prop and #173 by not updating value if < `min` or > `max`.

Also renamed a couple of props that weren't being picked up by React, similarly to #348.


test/unit/Input.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+    input.find('input').simulate('change', {target: {value: 'new value'}});
+
+    expect(input.find('input').getNode().value).toEqual('new value');
+
+    // dash-renderer could rerender a component with it's original
+    // props, if dash-renderer is not informed of prop updates
+    input.setProps({value: 'initial value'});
+
+    expect(input.find('input').getNode().value).toEqual('new value');
```
Marc-Andre-Rivet: Could you explain what is being tested here?

valentijnnieman: Yes! In some cases, `dash-renderer` will need to re-render the entire layout. It will, at that point, push the props in the components it wants to render. If, Dash isn't informed of any prop updates (in the case were `setProps()` is not being used, but the component's internal state is) then it will push in the old prop, or the initial prop.

dash_core_components/Location.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, id=Component.REQUIRED, pathname=Component.UNDEFINED, search=C
         _locals.update(kwargs)  # For wildcard attrs
         args = {k: _locals[k] for k in _explicit_args if k != 'children'}
 
-        for k in [u'id']:
+        for k in ['id']:
```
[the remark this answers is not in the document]
T4rk1n: That the dash generate component, I think I generated from py3 and it added that unicode, I regened from py2 and they were gone.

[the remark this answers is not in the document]
valentijnnieman: Ah ok! That's probably fine then!

valentijnnieman:
There were still some open issues on the Input component, figured I'd get some fixes for those in. This should close #169 now (when merge conflicts are resolved). I intend to look into #169 as well, while I'm at it. I'll change the status of this to WIP!

T4rk1n:
@valentijnnieman this fix #362, can you add this test to test_integration:

```python
    def test_input_lose_focus(self):
        app = dash.Dash(__name__)
        app.layout = html.Div([
            dcc.Input(
                id='input',
                value='initial value'
            ),
            dcc.Input(id='input-2'),
            html.Div(
                html.Div([
                    1.5,
                    None,
                    'string',
                    html.Div(id='output-1')
                ])
            )
        ])

        @app.callback(Output('output-1', 'children'), [Input('input', 'value')])
        def update_output(value):
            return value

        self.startServer(app)

        self.wait_for_text_to_equal('#output-1', 'initial value')

        input1 = self.wait_for_element_by_css_selector('#input')
        input1.clear()
        clicker = self.wait_for_element_by_css_selector('#input-2')
        clicker.send_keys('bad')

        input1.send_keys('hello world')
        time.sleep(1)

        self.assertEqual('hello world', input1.get_attribute('value'))
```

:dancer:

valentijnnieman:
@T4rk1n Sorry, I'm not sure what this test is testing, could you explain? I ask because I want to try to split up tests into (1) Unit tests that go into `test/unit/{componentName}.test.js` and (2) Integration tests that use an actual Dash app that go into `test/test_integration.py`. I'm not sure if your test should be a unit test or an integration test.

T4rk1n:
It's an integration test for a bug introduced by n_blur and setState in componentWillReceiveProps, when the component lost focus it would prepend the value with the initial value. 

So I wrote this test that change the focus to another input and assert the first input is still the same value, it failed with the dcc version on master. `wait_for_text_to_equal` is actually `wait_for_text_to_be_present`, so the previous tests in dash would pass with those [percy diffs](https://percy.io/plotly/dash/builds/1118236). With your changes to props handling this test passed.

valentijnnieman:
@T4rk1n In that case, the behaviour of new props and `setState` in `componentWillReceiveProps` is already being tested in the unit test, I feel. I'm kind of wary of adding (slow) integration tests that test behaviour that is (or should be) already tested in that component's unit tests. If you want you can make a PR later that adds this test, or better yet, update the new unit tests!

T4rk1n:
Testing time used to be not so slow, I investigated and found a few issues. [Dropped the test time from ~8:30 avg to ~4:30.](https://circleci.com/gh/plotly/dash-core-components/1998).

The test is already written and [proven to fail](https://circleci.com/workflow-run/ff0648d9-eab9-4ace-a672-159cee29d44f), it only adds at most a couple seconds. I've been hunting this bug for two weeks, finally found the culprit and wrote the test with failing behavior with the intention to fix it, was out of idea and noticed you changed things here so I tried and it passed. I'd prefer not having to rewrite it, so I'll merge master after this PR on the failing branch and add it in another PR.

Have yet to take a good look at the new unit tests, I don't know much about jest/enzyme. They look good at first glance with the mocked setProps.

valentijnnieman:
@T4rk1n Yeah the time isn't really that much of an issue for me, it's more that I think that what you're testing in your test is already covered in the unit tests now. I want to really start dividing tests between unit and integration, testing the behaviour of the component in the unit tests and the integration with Dash in the integration tests. That way, we can more easily do test driven development, writing unit tests that fail and make them pass, then assert that the behaviour in Dash stays the same later in the integration tests. It's just a lot easier to do that as opposed to make a change to the code, do a build:js and build:py, and then fire up the Selenium tests each time to make sure it works. With Jest you're testing the `src` code, not the build. Having said all that, I don't think there are a lot of `Input` integration tests, so that's something that we should change.

CHANGELOG.md
```diff
@@ -2,8 +2,23 @@
...
+### Added
+- Unit tests for `Input` component.
+- New `debounce` prop for `Input` component that determines if the input should update to a Dash server immediately, or on 'Enter' key. Fixes [#169](https://github.com/plotly/dash-core-components/issues/169)
+### Changed
+- `min` and `max` prop now won't update the input when values are lower than or greater than `min` and `max` respectively. Fixes[#173](https://github.com/plotly/dash-core-components/issues/173)
+- `step` prop can now be a `number` and is therefor set correctly on the corresponding `<input/>` tag. Fixes [#292](https://github.com/plotly/dash-core-components/issues/292)
+
 ## [0.36.0] - 2018-11-01
-### Fixed
+### Fixe
```
Marc-Andre-Rivet: The `d` is gone!

src/components/Input.react.js
```diff
@@ -12,58 +12,91 @@ import {omit} from 'ramda';
...
                 onChange={e => {
-                    this.setState({value: e.target.value});
-                    if (setProps) {
-                        if (type === 'number') {
-                            setProps({value: Number(e.target.value)});
-                        } else {
-                            setProps({value: e.target.value});
-                        }
+                    const newValue = e.target.value;
+                    if (((min || max) && newValue < min) || newValue > max) {
```
Marc-Andre-Rivet: I think this check is wrong. (1) `0` is falsy and max is compared always

CHANGELOG.md
Marc-Andre-Rivet: I unno!

test/unit/Input.test.js
```diff
@@ -0,0 +1,305 @@
...
+                min: 0,
+                max: 2,
+            };
+            beforeEach(() => {
+                input = mount(<Input type="number" {...props} />);
+            });
+            test('Input can not be updated lower than props.min', () => {
+                input
+                    .find('input')
+                    .simulate('change', {target: {value: props.min - 1}});
```
Marc-Andre-Rivet: As discussed, these tests are incorrect in so far as they are sending a number to the component when an actual onchange event would send a string

test/unit/Input.test.js
```diff
@@ -0,0 +1,314 @@
...
+                );
+            });
+        });
+        describe('without min and max props', () => {
+            let input;
+            beforeEach(() => {
+                input = mount(<Input type="number" value={0} />);
+            });
+            test('Input can be updated', () => {
+                input.find('input').simulate('change', {target: {value: -1}});
```
Marc-Andre-Rivet: make it a string

test/unit/Input.test.js
```diff
@@ -0,0 +1,314 @@
...
+                expect(Number(input.find('input').getNode().value)).toEqual(-1);
+                input.find('input').simulate('change', {target: {value: 100}});
```
Marc-Andre-Rivet: make it a string

Issue #346, opened 2018-10-26
`npm start` does not work for me


Akronix:
I've followed the instructions written in the README.md file for installing and setting up the development environment.

However, when I do `npm start` my browser page is black and I only see an error in the console saying:

> output.js:1 Failed to load resource: the server responded with a status of 404 (Not Found)

Here is the screenshot:

![image](https://user-images.githubusercontent.com/2631430/47558060-86dbf180-d912-11e8-805c-10e9b848d884.png)

valentijnnieman:
@T4rk1n Could you take a look at this? I think I just copied the webpack config from what you did at `dash-component-boilerplate`.

T4rk1n:
I'd just delete the script, there's no point in running those demos, they don't run the dash-renderer so you don't get `setProps` and all that jazz. You're always better off running a `usage.py`.

valentijnnieman:
I think there's still value in being able to run just the React code, without having to use Dash. If, for example, all you are doing is changing some UI code, it is faster to just run the JS code and not bother with any Python at that point. In any case, if we have a demo app, as we do in `demo/Demo.react.js`, we should support that and make sure that it works, or remove it entirely (including the webpack configs).

Akronix:
Ups, I lost the context and I did a mistake. I thought that this was dash-renderer repo when I read that comment and I responded accordingly :sweat_smile: .

For dash core components it is useful to have a working pure JS Demo, because I want to test the React components in pure JS before to do the whole python binding, actually, this is why I opened the issue: I wanted to test the input component in react before to make any further changes.
So yes, a working `npm start` script would be nice.

valentijnnieman:
Similarly happening in dash-component-boilerplate, here [#27](https://github.com/plotly/dash-component-boilerplate/issues/27)

T4rk1n:
It was resolved in plotly/dash-component-boilerplate#28

PR #359, opened 2018-11-01
Fix npm start to run Demo app

valentijnnieman:
@T4rk1n I've changed the webpack config to resemble https://github.com/plotly/dash-component-boilerplate/pull/28, can you please review? There's some more noise again in the build, can you confirm that these changes are coming from builds with py3 (like you mentioned [here](https://github.com/plotly/dash-core-components/pull/356#discussion_r229840252)) or is there more going on? Is there a way of preventing this from happening?


valentijnnieman:
Closes #346

PR #361, opened 2018-11-02
Datepickersingle prop typos

tcbegley:
This simple PR corrects a couple of typos in props being passed to `SingleDatePicker` from `react-dates` inside the definition of `DatePickerSingle`.

See [here](https://github.com/airbnb/react-dates/blob/v12.7.1/src/shapes/SingleDatePickerShape.js#L47) and [here](https://github.com/airbnb/react-dates/blob/v12.7.1/src/shapes/SingleDatePickerShape.js#L63)

In particular this means the `show_outside_days` prop of `DatePickerSingle` from `dash-core-components` has the desired effect.


rmarren1:
:dancer:

Marc-Andre-Rivet:
@tcbegley I see this PR has been sitting here for a few weeks now. All that's missing is a merge from master. Do you want to bring it through the finish line or can we do it ourselves? Whatever your preference, thanks for your contribution! 😄

Marc-Andre-Rivet:
> I don't have permission to merge
Ah! Didn't think of that. I guess that makes a lot of sense.

@tcbegley The contribution guideline is missing a few key points that were not raised during the review.

What's needed:
1. resolve conflicts (just accept what comes from master and rebuild, push the rebuild's result -- the build has changed so there will be additional files generated)
- bump patch version in package.json and version.py (we're currently at 0.40.3, so 0.40.4)
2. update changelog
```
## [0.40.4] - 2018-12-11
### Fixed
3. Fix typos in DatePickerSingle props [#361](https://github.com/plotly/dash-core-components/pull/361)
```
4. updated build artefacts (yes, the build artefacts need to be updated in the PR with our current process)

If you'd rather I do the changes in the fork I'm more than willing too. Don't hesitate if you have additional questions.

tcbegley:
@Marc-Andre-Rivet That makes sense, happy to take a pass at that now and push for you to review.

Marc-Andre-Rivet:
@tcbegley Sorry if this follow up is a bit haphazard, this is the first community PR I pick up here 😧

tcbegley:
@Marc-Andre-Rivet no worries at all! You've been very helpful so far.

I've made the changes you suggested (with the exception of bumping to `0.40.5` as `0.40.4` appears to already be on master and PyPI, is that right?).

tcbegley:
I was wondering about that. I'm guessing to do with the fact that `show_outside_days` previously defaulted to `true` but didn't actually do anything because it was passed to `enableOutSideDays` which wasn't recognised by react-dates.

Anyway, thanks for the help and for merging!

PR #364, opened 2018-11-05
Improve test time

T4rk1n:
There was `self.driver.implicitly_wait(2)` that slipped in, this PR removes it and further decrease sleep times.

### DIFF
--- a/test/IntegrationTests.py
+++ b/test/IntegrationTests.py
@@ -95,11 +95,10 @@ class IntegrationTests(unittest.TestCase):
         else:
             self.server_process = multiprocessing.Process(target=run)
             self.server_process.start()
-        time.sleep(5)
+        time.sleep(3)
 
         # Visit the dash page
         self.driver.get('http://localhost:8050')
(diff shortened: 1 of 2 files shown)


chriddyp:
Yeah, I'm fine with it as long as tests pass and as long as we're using `wait_for_test_to_equal`

T4rk1n:
> Yeah, I'm fine with it as long as tests pass and as long as we're using wait_for_test_to_equal

Which reminds me that it's not `wait_for_text_to_equals` but `wait_for_text_to_be_present`, I'll fix that later.

T4rk1n:
@chriddyp I think the additional sleep time for the tabs test is because of the percy diffs ? I think there's a delay before those graphs auto resize.

T4rk1n:
@valentijnnieman @chriddyp The new percy snapshots got the css styles ???

wbrgss:
Shaving off the sleep time causes no regressions as far as I can tell. :dancer:

PR #366, opened 2018-11-06
security updates

rmarren1:
Update packages in response github security alerts.

This resolves the alerts here: https://github.com/plotly/dash-core-components/network/alerts (not sure who has access to view these)


rmarren1:
Is there something like greenkeeper.io that could automate this process? I've seen https://dependabot.com/ used in some open source repos for Python.

rmarren1:
Not going to make a release because this is just used in development / testing.

PR #372, opened 2018-11-08
Update requirements

rmarren1:
Quick cleanup of the requirements files.


rmarren1:
I wanted to mirror the changes [here](https://github.com/plotly/dash/pull/439).
I think it is important to 
1) keep the location of these files consistent throughout all main repos (dcc, dash, html)
2) only have one requirements file to install dev requirements (or > 1 in the case that there are significant differences between python versions)
3) keep these files minimal, preferably with the `install_requires` packages plus `selenium` and `percy` so that the most up to date packages are installed in CI, and so the github security alerts go away

That PR does all of these well, I agree that the location is probably not the best. I think a plain `requirements.txt` file at the top level is expected, finding it in the hidden `.circleci` folder might make it difficult for brand new developers to get started.

rmarren1:
What do you think about the same changes just with `.circleci/requirements/dev.txt` moved to `requirements.txt`? I could mirror these in `dash` and `dash-html-components` as well.

T4rk1n:
It was asked to move them in dash because it caused confusion for people looking at the repo and the old `requirements.txt` was all locked with old versions that would cause problems when installed. Maybe add it back with all unlocked versions of true dependencies with a comment on top saying theses requirements are for dev purpose only.

rmarren1:
Okay that makes sense. I'll add a comment, and perhaps we should name it `dev-requirements.txt` instead to be super explicit as to what they are there for.

rmarren1:
@valentijnnieman Moved that file back, also figured out the Percy error (the package to load excel files wasn't included in the dev requirements :see_no_evil:)

valentijnnieman:
Cool, that looks good to me!

rmarren1:
Not making release as this PR only changes testing / dev requirements.

PR #373, opened 2018-11-08
Update dependencies to their latest

valentijnnieman:
This PR updates most of the packages to their latest version.

This also removes some unused packages, like builder and old babel packages that are no longer needed.

Note that the move from react-dates 12.3 to 18.2 is a slightly bigger one - not only does it change the way styling is handled, but it also just looks different in general. I've updated the way we style our DatePicker components with some simple overwrites, but it still will look a little bit different. I don't think that's a bad thing though, necessarily. Check out the Percy diffs to see the changes in the DatePickers.


rmarren1:
Would this require moving `dash-renderer` to react 15?

rmarren1:
Those percy diffs look okay to me. Looks like a few components grew a pixel or two

src/components/DatePickerRange.react.js
```diff
@@ -1,3 +1,4 @@
+import 'react-dates/initialize';
```
valentijnnieman: This is the new (or old, in any case the documented) way of using `react-dates`, as is on the [github page](https://github.com/airbnb/react-dates#initialize)

bpostlethwaite:
We should either support package.lock and `npm` or `yarn` and `yarn.lock`. I think moving to `yarn` should happen as a separate project.

Can you remove `yarn.lock` and all references and commits to `yarn` and `yarn.lock`?

test/unit/Input.test.js
```diff
@@ -51,61 +51,61 @@ describe('Props can be set properly', () => {
 
     test('props.id is set as the <input> id', () => {
         // test if id is in the actual HTML string
-        const inputTag = input.render().children();
+        const inputTag = input.render();
```
wbrgss: What's the difference between `render()` and `render().children()` ? Does the change to the latter have to do with the React 16 upgrade?

test/unit/Input.test.js
```diff
@@ -118,15 +118,15 @@ describe('Input with (default) type=text', () => {
             input = mount(<Input value="initial value" />);
         });
         test('Input updates value', () => {
-            expect(input.find('input').getNode().value).toEqual(
+            expect(input.find('input').instance().value).toEqual(
```
wbrgss: Similar question here, what is the difference between `getNode()` and `instance()`? An API change in `enzyme`?

src/components/css/react-dates-overrides.css
```diff
@@ -0,0 +1,32 @@
...
+* From the react-dates docs:
+* "Note: This component assumes box-sizing: border-box is set globally in your page's CSS."
+* If box-sizing: border box isn't set globally, the picker appears a bit broken.
+*/
+html {
+  box-sizing: border-box;
+}
+*, *:before, *:after {
+  box-sizing: inherit;
+}
```
wbrgss: This is a good idea in principle but it's a bit dangerous to set the wildcard selector. Other components might be expecting the default `content-box`; we should double check if the layout looks different. I'd prefer if the above lines (6-11) were more explicitly in a "global" or "base" CSS file with the above comment explaining the rationale.

test/test_integration.py
```diff
@@ -188,10 +188,10 @@ def test_upload_gallery(self):
                 'Drag and Drop or ',
                 html.A('Select a File')
             ], style={
-                'widatetimeh': '100%',
+                'width': '100%',
                 'height': '60px',
                 'lineHeight': '60px',
-                'borderWidatetimeh': '1px',
+                'borderWidth': '1px',
```
wbrgss: :+1:

.circleci/config.yml
```diff
@@ -19,7 +19,7 @@ jobs:
                   command: echo $CIRCLE_JOB > circlejob.txt
 
             - restore_cache:
-                  key: deps1-{{ .Branch }}-{{ checksum "dev-requirements.txt" }}-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "circlejob.txt" }}
+                  key: deps1-{{ .Branch }}-{{ checksum "dev-requirements.txt" }}-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "circlejob.txt" }}
```
wbrgss: This should be reverted right? As per @bpostlethwaite's [comment](https://github.com/plotly/dash-core-components/pull/373#issuecomment-437486409) and your [latest commit](https://github.com/plotly/dash-core-components/pull/373/commits/2e4b84fe03dc22cdaf8142d82c0d87cfda719bb1)

.circleci/config.yml
```diff
@@ -30,7 +30,7 @@ jobs:
                       npm install --ignore-scripts
 
             - save_cache:
-                  key: deps1-{{ .Branch }}-{{ checksum "dev-requirements.txt" }}-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "circlejob.txt" }}
+                  key: deps1-{{ .Branch }}-{{ checksum "dev-requirements.txt" }}-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "circlejob.txt" }}
```
wbrgss: same as above here

test/unit/Input.test.js
valentijnnieman: It has to do with the Enzyme upgrade.

test/unit/Input.test.js
valentijnnieman: Exactly. Enzyme got updated, and a bunch of the methods I used before didn't work anymore after upgrading.

.circleci/config.yml
valentijnnieman: Yeah, forgot that one, thanks!

jest.config.js
```diff
@@ -107,15 +107,15 @@ module.exports = {
...
+  roots: [
+    "<rootDir>/test/unit"
+  ],
 
   // Allows you to use a custom runner instead of Jest's default test runner
   // runner: "jest-runner",
 
   // The paths to modules that run some code to configure or set up the testing environment before each test
-  // setupFiles: [],
+  setupFiles: ['<rootDir>/test/setupTests.js'],
```
valentijnnieman: The Jest and Enzyme updates now require us to specify a React adapter to use - so we create this `setupTests.js` setup file to run before our tests.

jest.config.js
```diff
@@ -107,15 +107,15 @@ module.exports = {
   // rootDir: null,
 
   // A list of paths to directories that Jest should use to search for files in
-  // roots: [
-  //   "<rootDir>"
-  // ],
+  roots: [
+    "<rootDir>/test/unit"
```
valentijnnieman: We don't want to look in other folders for JS tests - specifically the new `setupTests.js` file shouldn't be run as a unit test with Jest.

src/components/Link.react.js
```diff
@@ -50,10 +50,10 @@ export default class Link extends Component {
...
         const {className, style, id, href} = this.props;
         /*
-            * ideally, we would use cloneElement however
-            * that doesn't work with dash's recursive
-            * renderTree implementation for some reason
-            */
+         * ideally, we would use cloneElement however
+         * that doesn't work with dash's recursive
+         * renderTree implementation for some reason
+         */
```
valentijnnieman: Sorry for this noise - it seems `Prettier` still picks up some formatting issues here and there from time to time. I think the formatting tests in Cirlce are maybe not 100% right, as it seems to miss these formatting issues.

src/components/SyntaxHighlighter.react.js
```diff
@@ -1,4 +1,4 @@
-import {arduinoLight, monokai} from 'react-syntax-highlighter/dist/styles';
+import {arduinoLight, monokai} from 'react-syntax-highlighter/dist/styles/hljs';
```
valentijnnieman: New syntax for updated `react-syntax-highlighter`

valentijnnieman:
@chriddyp We could still merge this branch, since almost all of the packages used are updated. If dash-renderer provides React, and if the dcc bundles build with these updated packages/dependencies work without failing, then surely that won't break anything? I've also checked the updated packages's dependencies, and none specify that they don't work below React 16. If you're good with it, I can take out the references to version 1.0.0 and we can merge this, closing all of the open greenkeeper ones (which was the goal of this PR when I started with it anyway)

valentijnnieman:
@Marc-Andre-Rivet Reverted the changes to the client-facing react packages such as react-dates, only internal tooling updates now. Again apologies for the messy PR / commit history!

Issue #379, opened 2018-11-09
clickData isn't registered when the graph is embedded in a tab?


chriddyp:
See https://community.plot.ly/t/creating-multi-tab-app-with-clickdata-functionality/14044

PR #381, opened 2018-11-13
Refactor Graph.defaultProps.id code into function

valentijnnieman:
The code for generating a random id in the `Graph` component's `defaultProps` was inline - causing trouble for R component generation. This PR factors the code out into it's own function, so that it is a little bit neater.


src/components/Graph.react.js
```diff
@@ -494,13 +505,7 @@ PlotlyGraph.propTypes = {
...
 
 PlotlyGraph.defaultProps = {
-    /* eslint-disable no-magic-numbers */
-    id:
-        'graph-' +
-        Math.random()
-            .toString(36)
-            .substring(2, 7),
-    /* eslint-enable no-magic-numbers */
+    id: generateId(),
```
Marc-Andre-Rivet: What about multiple graphs in the same page? They way they will all have the same id. Similar to when this was added to the dash-table.

src/components/Graph.react.js
```diff
@@ -69,6 +69,19 @@ export default class PlotlyGraph extends Component {
         super(props);
         this.bindEvents = this.bindEvents.bind(this);
         this._hasPlotted = false;
+
+        this.props.id = this.props.id ? this.props.id : this.generateId();
```
Marc-Andre-Rivet: This looks wrong. A component should never override its props.

If using state, setProps or another approach brakes tests, maybe we need to dig and find if that is correct or incorrect behavior from either the test or sub-components.

valentijnnieman: I agree that it's wrong. 

However, seeing as the test do indeed break, I'm not sure how to go from here. There are no sub-components, and the tests seem correct. Any help would be appreciated!

src/components/Graph.react.js
```diff
@@ -190,13 +207,13 @@ export default class PlotlyGraph extends Component {
...
         return (
-            this.props.id !== nextProps.id ||
+            this.state.id !== nextProps.id ||
             JSON.stringify(this.props.style) !== JSON.stringify(nextProps.style)
         );
     }
 
     componentWillReceiveProps(nextProps) {
-        const idChanged = this.props.id !== nextProps.id;
+        const idChanged = this.state.id !== nextProps.id;
```
Marc-Andre-Rivet: Since we pull the id from the state everywhere now, shouldn't we update the state to the new props.id value?

chriddyp:
Does the current code cause duplicate IDs (since it's in `defaultProps` instead of in runtime?). If so, then maybe this PR will fix https://github.com/plotly/dash-core-components/issues/379?

Marc-Andre-Rivet:
@chriddyp You're right. @valentijnnieman There's a id: generatedId() again in ```graphDefaultProps```, when it tries to validate at ```const id = props.id ? props.id : generateId();``` won't it pick up that default and defeat the purpose of the check?

valentijnnieman:
> Does the current code cause duplicate IDs (since it's in `defaultProps` instead of in runtime?). If so, then maybe this PR will fix #379?

It does create duplicate IDs, although I'm not sure if that will fix #379, since the examples given there set the `id` of the graphs explicitly.

Issue #362, opened 2018-11-02
Input n_blur lose focus get the initial values back in the component.


T4rk1n:
Source of the input percy diffs I have been trying to debug for the last week. The component lose focus after text assert and the initial value gets prepended somehow to the value.

https://percy.io/plotly/dash/builds/1118236

T4rk1n:
Bug is due to how selenium clear the input, it doesn't send an onChange event to react.

PR #384, opened 2018-11-14
Add more Input tests, remove duplicate value callbacks.

T4rk1n:
This PR adds a failing `Input` test and fix `wait_for_text_to_equal` that only checked if the text was present and not equal.

Tests are failing in `dash` and `dash-renderer` if we update the dash-core-components version to the latest.


valentijnnieman:
@T4rk1n The input has been changed to fire `setProps()` `onBlur`, causing these tests to fail - they're asserting `call_count` to be 2, when it is now 3, for example. We fire `setProps` `onBlur` now because we've added the `debounce` feature, allowing users to only fire `setProps` upon submit or `onBlur`, fixing the problems users had entering floats like 0.001, where `setProps` was being fired `onChange` caused `setProps` to truncate the 0.0 to just 0, overwriting the state of the input from 0.0 to 0. In other words, the tests need to be changed, or we only fire `setProps(value)` `onBlur` when `debounce` is true (that's probably the better solution).

T4rk1n:
@valentijnnieman That is fine for the call count tests. But there's still the bug in #362, it's hard to reproduce by hand because you have to type and change focus very fast, but it can happen. The only way I had that test not fail is by removing the `state.value` from `props.value` and that would remove the debounce logic. Or by keeping the initial value and checking it's not the same, but that is not a good solution either.

T4rk1n:
Both of those errors comes from that. The input get cleared ok, but it get the initial value back in componentWillReceiveProps caused by `n_blur` sending the value back or something when the input lose focus. It has to lose focus real fast when there is a value callback still going on to happen.

valentijnnieman:
@T4rk1n Are you still working on this? Would be good to get this fixed!

T4rk1n:
The only fix for that bug is removing the componentWillReceiveProps, it would remove the debounce logic for the number type. Maybe we can have another logic instead for that case. Like caching the value when the type is number and only setting it while setting n_blur/n_submit and only use the props for render.

T4rk1n:
After investigating a bit more the input clear bug, it boils down to the way selenium clear the input. It doesn't send the onChange event for the cleared value so somehow the initial value get backs in. This is not something that would happen if the user cleared the input manually with backspaces or selecting the whole text and deleting. Thus, it is safe to patch the failing tests to fix that selenium bug.

The callback counts were wrong because we added the debounce logic and the value was set all the time in onBlur and onSubmit. I added a check that only set the value if debounce is true, removing the duplicate value callbacks.

T4rk1n:
@valentijnnieman Please review

src/components/Input.react.js
```diff
@@ -65,22 +65,27 @@ export default class Input extends Component {
                     if (setProps) {
                         const castValue =
                             type === 'number' ? Number(value) : value;
```
valentijnnieman: For consistency, the way the `payload` is set up should probably be the same in the `onBlur` method as well as in the `onKeyPress` method. I might prefer the way it's done on line 86 - the `castvalue` const also shouldn't have to be defined if it's not being used here.

test/test_integration.py
```diff
@@ -1446,3 +1452,127 @@ def _insert_cookie(rep):
         self.wait_for_text_to_equal('#content', 'Logged out')
 
         self.assertFalse(self.driver.get_cookie('logout-cookie'))
+
+    def test_lose_focus_input(self):
```
valentijnnieman: Is this test really different from the `test_simple_callback` one? There, on line 1569, the input also loses focus. Perhaps this test could be removed?

PR #388, opened 2018-11-20
Add Logout button

T4rk1n:
Add a logout button, performs a form post request to a `logout_url`.

Prop | Type | Default | description
-- | -- | -- | --
id | string | - | Id of the button.
label | string | logout | Text of the button
logout_url | string | - | Url to submit a post logout request.
style | object | - | style of the button.
className | string | - | CSS class for the button.


T4rk1n:
https://github.com/plotly/streambed/issues/11817

T4rk1n:
@scjody 

There is no python hooks for the components so we can't do it at the component level. Can add it from a dds app with `dcc.LogoutButton(logout_url=os.getenv('DASH_LOGOUT_URL'))`.

scjody:
That URL doesn't look so bad, except you'll want a default so it doesn't crash in test environments and for non-DDS users. So something like: `dcc.LogoutButton(logout_url=os.getenv('DASH_LOGOUT_URL', 'https://dash.plot.ly/docs_on_logout'))`

Is it impossible to check `DASH_LOGOUT_URL` in [this function](https://github.com/plotly/dash-core-components/pull/388/files#diff-02b032de2fd7f7e82055f0f33b09fdcaR19)? (I don't know how `LogoutButton.py` gets generated so I accept that it may not be, but it seems strange that we wouldn't have control over the backend of all our components.)

T4rk1n:
The python component class is auto-generated, it will get overwritten when the components are regenerated.

scjody:
I realize that, but I'm surprised we don't have any control over the generation process or what goes into the component at that time.

Anyway if my suggestion is unreasonable I'm OK leaving things as they are.

T4rk1n:
W could have some kind of hooks, it's just not implemented right now. I created an issue:

https://github.com/plotly/dash/issues/464

src/components/LogoutButton.react.js
```diff
@@ -0,0 +1,58 @@
...
+    render() {
+        const {id, logout_url, label, className, style} = this.props;
+        return (
+            <form
+                action={logout_url}
+                method="post"
+                className="dash-logout-frame"
+            >
+                <button
+                    className={`dash-logout-btn ${className}`}
```
valentijnnieman: Here if `className` is undefined, it will add "undefined" to the class. Would be cleaner to check if `className` exists before appending it.

T4rk1n:
@valentijnnieman 

There's an example in the test:

```python
@app.server.route('/_logout', methods=['POST'])
def on_logout():
      rep = flask.redirect('/logged-out')
      rep.set_cookie('logout-cookie', '', 0)
      return rep
```

T4rk1n:
I'll make `https://dash.plot.ly/dash-core-components/logout_button` the default url and adds documentation and examples there when released.

valentijnnieman:
@T4rk1n Maybe add some more info (or a link to those docs) in the docstring for the component as well, so that it's more clear for users what this component is for.

valentijnnieman:
Great - looks good to me!

PR #392, opened 2018-11-26
Fix Tabs vertical mode

### LINKED ISSUE #390: New Tabs/Tab components displays wrong in vertical mode.
Hi,

The new Tabs/Tab core components have display issues when used in vertical=True mode. 
The following are differences between the old tabs component and the new one:

* For unselected tabs, the gray right side border is missing.
* Selected tabs (except the last one) have an incorrect extra gray bottom border.
* The last selected tab has an incorrect extra top blue border.
* The bottom part of the tabs component beneath the tabs does not have  a border on the right side.

Also, both in vertical and horizontal mode the tabs displays wrong at small sizes.
This can be worked around by setting mobile_breakpoint=0.

[tabs.zip](https://github.com/plotly/dash/files/2565102/tabs.zip)

valentijnnieman:
This fixes and closes #390 by providing better styles for the Tabs component in vertical mode (`Tabs(vertical=true)`), and adds an integration test for vertical mode.


valentijnnieman:
@wbrgss I thought you might wanted to take a look at this, to make sure this doesn't change anything for DDK.

test/test_integration.py
```diff
@@ -480,6 +480,22 @@ def test_gallery(self):
...
+            dcc.Tabs(id="tabs", value='tab-3', children=[
+                dcc.Tab(label='Tab one', value='tab-1', id='tab-1'),
+                dcc.Tab(label='Tab two', value='tab-2', id='tab-2'),
+                dcc.Tab(label='Tab three', value='tab-3', id='tab-3'),
+                ], vertical=True),
+            html.Div(id='tabs-content')
+        ])
+
+        self.startServer(app=app)
+        self.wait_for_text_to_equal('#tab-3', 'Tab three')
```
Marc-Andre-Rivet: Could you add a child to that tab so we know it's displaying it when selected?

dash_core_components/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.40.0'
+__version__ = '0.40.2'
```
Marc-Andre-Rivet: This version is inconsistent with the others!

valentijnnieman: Fixed typo :)

src/components/Tabs.react.js
```diff
@@ -287,17 +293,22 @@ export default class Tabs extends Component {
                         :global(.tab-container--vert .tab) {
                             width: auto;
                             border-right: none !important;
-                            border-bottom: none;
+                            border-bottom: none !important;
```
wbrgss: Is this overriding a `border` property set inline with JS? If not the `!important` could probably be avoided

PR #394, opened 2018-11-27
Updates react-markdown to most recent version

edwardreed81:
This is mainly to support tables in Markdown. Resolves #393.

PR #395, opened 2018-11-27
Adds boolean to Markdown for HTML escaping

edwardreed81:
This allows the developer to render raw HTML inside Markdown if desired.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "react-dates": "^12.3.0",
     "react-docgen": "^2.20.1",
     "react-dropzone": "^4.1.2",
-    "react-markdown": "^4.0",
+    "react-markdown": "^2.4.5",
     "react-select": "^2.1.0",
     "react-select-fast-filter-options": "^0.2.2",
     "react-syntax-highlighter": "^5.0.0",
--- a/src/components/Markdown.react.js
+++ b/src/components/Markdown.react.js
@@ -13,7 +13,7 @@ function DashMarkdown(props) {
         props.children = props.children.join('\n');
     }
 
-    return <Markdown source={props.children} {...props} />;
+    return <Markdown source={props.children} escapeHtml={!props.dangerously_allow_html} {...props} />;
 }
 
 DashMarkdown.propTypes = {
@@ -35,7 +35,7 @@ DashMarkdown.propTypes = {
      * inadvertently expose your users to a cross-site scripting (XSS)
      * (https://en.wikipedia.org/wiki/Cross-site_scripting) attack.
      */
-    escapeHtml: PropTypes.bool,
+    dangerously_allow_html: PropTypes.bool,
 
     /**
      * A markdown string (or array of strings) that adhreres to the CommonMark spec
@@ -47,7 +47,7 @@ DashMarkdown.propTypes = {
 };
 
 DashMarkdown.defaultProps = {
-    escapeHtml: true,
+    dangerously_allow_html: false,
 };
 
 export default DashMarkdown;
--- a/src/components/Markdown.react.js
+++ b/src/components/Markdown.react.js
@@ -13,7 +13,7 @@ function DashMarkdown(props) {
         props.children = props.children.join('\n');
     }
 
-    return <Markdown source={props.children} escapeHtml={true} {...props} />;
+    return <Markdown source={props.children} {...props} />;
 }
 
 DashMarkdown.propTypes = {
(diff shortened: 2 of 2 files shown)


valentijnnieman:
Thanks for making this PR! This looks good to me.

@chriddyp Is there any reason (security?) why the Markdown component doesn't support the `escapeHtml` prop already?

chriddyp:
Yeah, we might want to do a quick rename on this prop, like `dangerouslyEscapeHtml`, similar to https://github.com/plotly/dash-dangerously-set-inner-html

chriddyp:
and our docstring for the prop should describe why this can be dangerous as well

edwardreed81:
I'm happy to rename the prop to remind people it is dangerous, but I don't feel qualified to issue warnings about the dangerousness of this... someone else want to take a shot at it?

package.json
```diff
@@ -50,7 +50,7 @@
     "react-dates": "^12.3.0",
     "react-docgen": "^2.20.1",
     "react-dropzone": "^4.1.2",
-    "react-markdown": "^2.4.5",
+    "react-markdown": "^4.0",
```
valentijnnieman: Could you remove this line in favour of doing it in a separate PR, possibly #394?

src/components/Markdown.react.js
```diff
@@ -13,7 +13,7 @@ function DashMarkdown(props) {
         props.children = props.children.join('\n');
     }
 
-    return <Markdown source={props.children} escapeHtml={true} {...props} />;
+    return <Markdown source={props.children} {...props} />;
```
valentijnnieman: I think you're missing something like `escapeHtml={props.dangerouslyEscapeHtml}` here - the `{...props}` spread here isn't enough because of the naming difference (dangerouslyEscapeHtml vs escapeHtml). Could you add that?

src/components/Markdown.react.js
```diff
@@ -29,6 +29,14 @@ DashMarkdown.propTypes = {
      */
     containerProps: PropTypes.object,
 
+    /**
+     * A boolean to control raw HTML escaping.
+     * Setting HTML from code is risky because it's easy to 
+     * inadvertently expose your users to a cross-site scripting (XSS)
+     * (https://en.wikipedia.org/wiki/Cross-site_scripting) attack.
+     */
+    dangerouslyEscapeHtml: PropTypes.bool,
```
valentijnnieman: 👍

edwardreed81:
I just remembered that `escapeHtml = true` for `react-markdown` is the safe choice; rendering raw HTML requites `escapeHtml = false`.

I could make the option inside `dcc` something like `dangerouslyRenderHtml` and have the boolean values reversed from how `react-markdown` handles it. Or keep paradigm and name the parameter `safelyEscapeHtml`.

What do you guys think?

package.json
edwardreed81: Whoops, I got my branches mixed up apparently. This'll be reverted in the next commit.

src/components/Markdown.react.js
edwardreed81: Thanks for pointing this out, I forgot. Once I resolve the naming/boolean question below, it'll be in the next commit.

valentijnnieman:
In that case, I think it would be best to stick to the naming `react-markdown` uses. That way, the prop in Dash maps 1-to-1 to `react-markdown`.

edwardreed81:
Do you mean just use `escapeHtml` like `react-markdown` or use `safelyEscapeHtml`?

edwardreed81:
I reverted back to using just `escapeHtml` to match what `react-markdown` does. I added a default value of `true` for this prop (the safe choice).

Not sure why the checks failed; I don't know what `prettier` is.

chriddyp:
I would still prefer to warn users that providing raw HTML is dangerous. For many dash users, this will be their first encounter with XSS-like attacks, and I'd like to be very upfront about how dangerous this can be.

Keeping the props matching 1-1 is nice, but it's not required by us. There is already a precedent for us _not_ doing this, e.g. with the `DatePickerSingle`, we convert the prop types to underscore-case.

So, I'd say we name the prop something like `dangerously_allow_html` (which gets mapped as `escapeHtml = !dangerously_allow_html`). This is similar to the language that we use elsewhere ("dangerouslySetInnerHtml")

edwardreed81:
I'm happy with that @chriddyp. I made those changes, but again I don't understand why the checks are failing.

edwardreed81:
What needs to happen to get this merged?

edwardreed81:
@valentijnnieman the Percy check fails now, but I get "Something went wrong" when I click Details. Any idea?

valentijnnieman:
@edwardreed81 Ah that's probably because you don't have access. I've taken a look and looks like it's all good - Percy takes snapshots of what it thinks are changes, but it all looks good to me.

Issue #357, opened 2018-11-01
R version


nicolaskruchten:
Once https://github.com/plotly/dash/issues/445 is completed, this repo should be upgraded such that it runs the R code generation by default, and is installable via `install_github()`.

rpkyle:
PR #411 addresses this issue.

nicolaskruchten:
A couple of Github workflow recommendation:

1. It's generally not expected that you would comment "PR #411 addresses this issue" if you mention #357 in the PR itself: as you can see, the back-reference is automatically listed here already
2. If you modify your comment in #411 to say `closes #357` instead of `addresses issue #357` then you get some Github magic which will automatically close #357 when #411 is merged to `master`: https://help.github.com/articles/closing-issues-using-keywords/`

nicolaskruchten:
(You'll know the magic is correctly set up because you get little dotted lines under the word "closes")

Here's an example pair: https://github.com/plotly/dash/issues/434 was auto-closed by https://github.com/plotly/dash/pull/439 (note the special Github-generated comments and backlinks and formatting etc)

PR #411, opened 2018-12-07
Update R dashCoreComponents package to 0.40.2

rpkyle:
The JSON >> R 'transpiler' has been rewritten in Python, and was used to generate an installable R package for version 0.40.2 of `dash-core-components`. This PR addresses issue #357.

@nicolaskruchten @T4rk1n


package.json
```diff
@@ -20,6 +25,7 @@
     "uninstall-local": "pip uninstall dash-core-components -y",
     "build:js": "webpack --mode production",
     "build:js-dev": "webpack --mode development",
+    "build:r": "node ./extract-meta src/components > dash_core_components/metadata.json && cp package.json dash_core_components && npm run generate-python-classes && npm run generate-r-classes",
```
alexcjohnson: Does `generate-r-classes` actually require `generate-python-classes` to be run first? regardless, the `extract-meta` and `cp package.json` parts are going to be shared so I wonder if we just want to have one command `build:backend` or something that does both languages (and any others we add later)? And use that in `build:all` (which is currently an exact copy of `prepublish` but perhaps shouldn't be cc https://github.com/plotly/dash-html-components/pull/83)

man/coreChecklist.Rd
```diff
@@ -0,0 +1,32 @@
...
+
+\item{options}{An array of options}
+
+\item{className}{The class of the container (div)}
+
+\item{values}{The currently selected value}
+
+\item{id}{}
+}
+
```
alexcjohnson: There is inconsistent whitespace at the end of these `.Rd` files, as well as trailing whitespace on some of the blank lines above. Can we get rid of these?

dash_core_components/_imports_.py
```diff
@@ -1,47 +1,47 @@
...
     "RangeSlider",
-    "Slider",
-    "Store",
-    "SyntaxHighlighter",
-    "Tab",
-    "Tabs",
     "Textarea",
-    "Upload"
+    "Tabs",
+    "Slider"
```
alexcjohnson: These were sorted before, now they're not. What changed?

dash_core_components/Upload.py
```diff
@@ -9,44 +9,44 @@ class Upload(Component):
...
+        self._prop_names = ['children', 'disabled', 'style', 'multiple', 'contents', 'disable_click', 'accept', 'id', 'min_size', 'className', 'style_reject', 'style_active', 'max_size', 'last_modified', 'className_active', 'filename', 'style_disabled', 'className_reject', 'className_disabled']
         self._type = 'Upload'
         self._namespace = 'dash_core_components'
         self._valid_wildcard_attributes =            []
         self.available_events = []
-        self.available_properties = ['children', 'id', 'contents', 'filename', 'last_modified', 'accept', 'disabled', 'disable_click', 'max_size', 'min_size', 'multiple', 'className', 'className_active', 'className_reject', 'className_disabled', 'style', 'style_active', 'style_reject', 'style_disabled']
+        self.available_properties = ['children', 'disabled', 'style', 'multiple', 'contents', 'disable_click', 'accept', 'id', 'min_size', 'className', 'style_reject', 'style_active', 'max_size', 'last_modified', 'className_active', 'filename', 'style_disabled', 'className_reject', 'className_disabled']
```
alexcjohnson: Similarly, these ones were not sorted before and they're a *different* unsorted order now. I don't really care about the sort order here, but I don't want unnecessary diffs. What changed?

package.json
rpkyle: Sorry for not replying to these comments late last month; when I last spoke with @nicolaskruchten and @T4rk1n about component generation in Python vs. R, I believe we decided that support for the latter remains experimental. 

Currently this requires setting the `--r-prefix` flag when calling `dash-generate-components`, but we could eventually remove the flag and generate both at once. 

It does (for now) require that the Python classes are generated first, since JS dependency information is sourced from `__init__.py` and the R package version number from `version.py`.

dash_core_components/_imports_.py
rpkyle: The 'transpiler' was rewritten in Python, and the sort order changed relative to the R version. I will correct this; sorry for the unnecessary diffs this generated...

rpkyle: OK, I've looked into this. The reshuffled order isn't present when cloning the repo, but occurs after running `dash-generate-components ./src/components/ dash-core-components`, with or without the `--r-prefix`. 

It seems to be outside of the R component generation methods, but I might be mistaken.

@T4rk1n @rmarren1

dash_core_components/Upload.py
rpkyle: Please see above comment, same situation.

package.json
nicolaskruchten: remind me again why the R transpiler gets the version info from a generated Python file? I know we've talked about this before but I don't recall.

dash_core_components/_imports_.py
nicolaskruchten: if that's the case, we should make a separate PR that commits that batch of changes, so we can review/approve the Python and R changes separately :)

rpkyle: @alexcjohnson By commenting about this change, you revealed two distantly related issues that should be addressed. Thanks -- talking with @rmarren1, this is probably related to swapping out `_get_metadata` for `json.loads`, and inadvertently not using an `OrderedDict` anymore.

I'm going to verify this, will then open PR for Dash if true, and link back to here.

PR #412, opened 2018-12-10
Update renovate configuration

Marc-Andre-Rivet:
- Align with current configuration for Dash-Table (https://github.com/plotly/dash-table/blob/master/renovate.json)

- Exclude major version bump on known problematic UI packages

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,8 +1,72 @@
 {
     "extends": [
         ":onlyNpm",
-        "group:all"
+        "schedule:daily"
     ],
-    "enabledManagers": ["npm"],
-    "schedule": "on monday"
+    "enabledManagers": [
+        "npm"
+    ],
+    "groupName": "all dependencies",
+    "groupSlug": "all",
+    "lockFileMaintenance": {
+        "enabled": false
+    },
+    "packageRules": [
+        {
+            "packageNames": [
+                "rc-slider"
+            ],
+            "allowedVersions": "<9.0.0"
+        },
+        {
+            "packageNames": [
+                "react-dates"
+            ],
+            "allowedVersions": "<13.0.0"
+        },
+        {
+            "packageNames": [
+                "react-dropzone"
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,8 +1,72 @@
...
     "extends": [
         ":onlyNpm",
-        "group:all"
+        "schedule:daily"
     ],
-    "enabledManagers": ["npm"],
-    "schedule": "on monday"
-}
+    "enabledManagers": [
+        "npm"
```
Marc-Andre-Rivet: Let's just do package.json and get it right for now -- we can cover requirements.txt later

renovate.json
```diff
@@ -1,8 +1,72 @@
...
+                "react-syntax-highlighter"
+            ],
+            "allowedVersions": "<6.0.0"
+        },
+        {
+            "packageNames": [
+                "react-virtualized-select"
+            ],
+            "allowedVersions": "<4.0.0"
+        },
```
Marc-Andre-Rivet: Prevent UI packages that need human intervention to be updated from attempting to update automatically

renovate.json
```diff
@@ -1,8 +1,72 @@
...
+        {
+            "packagePatterns": [
+                "*"
+            ],
+            "depTypeList": [
+                "dependencies",
+                "devDependencies"
```
Marc-Andre-Rivet: For all others, combine into a single PR

renovate.json
```diff
@@ -1,8 +1,72 @@
...
+            ],
+            "groupName": "all dependencies",
+            "groupSlug": "all"
+        },
+        {
+            "depTypeList": [
+                "peerDependencies",
+                "engines"
```
Marc-Andre-Rivet: Only do deps and devDeps -- engine and peerDeps are environment or Dash related and need to be taken care of by humans

PR #417, opened 2018-12-12
prepublish -> prepare

rmarren1:


Marc-Andre-Rivet:
'prepare' is run in on more situation than `prepublish`: `npm install my-package-with-prepare-script` vs only on `npm publish` and local-to-project `npm install`
https://docs.npmjs.com/misc/scripts

Is this intended or an unplanned side-effect? For standalone component scenarios (this is a desired scenario for the dash-table, could be valid for dcc), this might be an issue.

rmarren1:
The goal is to get rid of the deprecation warnings for `prepublish`. Seems the two options are `prepublishOnly` and `prepare`. I chose `prepare` since it builds on install which was how `prepublish` worked before and I thought developers would be used to that, also since I couldn't think of a situation where someone would `npm install dash-core-components`.

Would be totally fine making it `prepublishOnly` too. I end up running `npm run build:all` after install anyways by habit, but I was trying to make no changes to peoples workflows.

Marc-Andre-Rivet:
I re-read the code. I misread the additional use case. Prepare does keep the existing behavior.

Issue #422, opened 2018-12-21
Can't put null values in store array?


rmarren1:
Not sure what is happening with the Store here (is this a bug or a limitation?)

```
import numpy as np
import dash
import dash_html_components as html
import dash_core_components as dcc
from dash.dependencies import Input, Output, State

app = dash.Dash(__name__)

app.layout = html.Div([
    html.Div(id='output'),
    dcc.Store(id='store', data={'hello': np.array([1.1, 1.1, None])})
])

@app.callback(Output('output', 'children'),
              [Input('store', 'data')])
def hello(data):
    print(type(data['hello']))
    print(data['hello'])
    return data['hello']  # same happens if you don't render array, e.g just return 'hello'

app.run_server(
    port=8888,
    dev_tools_serve_dev_bundles=True
)

```

This app seems to work since the array will be printed correctly in the `hello` callback, suggesting that everything worked out (np.array was serialized to a JSON array with `null` instead of `np.nan`, this made it to the redux store, then the `hello` callback was triggered and the `null` value was converted to a `None`.

But this strange error is thrown, no idea why
![image](https://user-images.githubusercontent.com/10272301/50364344-ab80ce00-053d-11e9-8f46-cd3022641d00.png)

T4rk1n:
Two different bugs.

- null values in lists results in a infinite recursive loop.
- changing the root type in a callback also result in a infinite recursive loop.

rmarren1:
> changing the root type in a callback also create an infinite recursive loop.
What does this mean? That if my store is `{'data': [1, 2, 3]}` and I change it to `{'data': 'hello'}`, it will break? If that what is meant by changing the root type?

T4rk1n:
No it shouldn't break, just not update.

T4rk1n:
> `{'data': [1, 2, 3]}` and I change it to `{'data': 'hello'}`

That worked when I tried, maybe it was just the null check that was missing. I'll still add a test for that use case.

T4rk1n:
The tests for the store data types changes will need to assert that changes from a combination of the different types will update the output.

```python
>>> pprint.pprint(list(itertools.combinations(['str', 'number', 'dict', 'list', 'None'], 2)))
[('str', 'number'),
 ('str', 'dict'),
 ('str', 'list'),
 ('str', 'None'),
 ('number', 'dict'),
 ('number', 'list'),
 ('number', 'None'),
 ('dict', 'list'),
 ('dict', 'None'),
[...]

PR #424, opened 2018-12-26
Fix store null values in lists

T4rk1n:
Assert that both old and new data is not null when checking if the data needs update.

Part of fix for #422.

@rmarren1 please review.

### DIFF
--- a/src/components/Store.react.js
+++ b/src/components/Store.react.js
@@ -4,7 +4,9 @@ import PropTypes from 'prop-types';
 
 function dataCheck(data, old) {
     // Assuming data and old are of the same type.
-    if (R.isNil(old) || R.isNil(data)) {
+    const oldNull = R.isNil(old);
+    const newNull = R.isNil(data);
+    if ((oldNull || newNull) && !(oldNull && newNull)) {
         return true;
     }
     const type = R.type(data);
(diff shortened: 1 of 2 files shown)


rmarren1:
:dancer:

src/components/Store.react.js
```diff
@@ -4,7 +4,9 @@ import PropTypes from 'prop-types';
 
 function dataCheck(data, old) {
     // Assuming data and old are of the same type.
-    if (R.isNil(old) || R.isNil(data)) {
+    const oldNull = R.isNil(old);
+    const newNull = R.isNil(data);
+    if ((oldNull || newNull) && !(oldNull && newNull)) {
```
alexcjohnson: I found this a bit hard to read... how about `if (oldNull !== newNull)` ? Or I suppose you could short-circuit the logic below with something like `if (oldNull || newNull) { return oldNull !== newNull; }`

Separately, it looks to me as though the checks below implicitly assume `old` and `data` have the same type, and may fail if for example `data` is an `Array` and `old` is an `Object`. Should we include a step `if (R.type(old) !== type) { return false; }`?

T4rk1n: > Separately, it looks to me as though the checks below implicitly assume old and data

That's the second part of #422, I'll add a test and fix it in another PR. Initial implementation assumed data was of the same shape.

https://github.com/plotly/dash-core-components/blob/fa82eeb17ee7831d54fa08ea84f2054e7b5fd027/src/components/Store.react.js#L6

PR #427, opened 2018-12-28
Test store data types update & fixes

### LINKED ISSUE #422: Can't put null values in store array?

T4rk1n:
Closes #422


src/components/Store.react.js
```diff
@@ -2,27 +2,36 @@ import R from 'ramda';
...
 
-function dataCheck(data, old) {
+/**
+ * Deep equality check to know if the data has changed.
+ *
+ * @param {any} newData - New data to compare
+ * @param {any} oldData - The old data to compare
+ * @returns {boolean} The data has changed.
+ */
+function dataChanged(newData, oldData) {
```
alexcjohnson: Beautiful, thank you!

src/components/Store.react.js
```diff
@@ -2,27 +2,36 @@ import R from 'ramda';
...
     // Assuming data and old are of the same type.
-    const oldNull = R.isNil(old);
-    const newNull = R.isNil(data);
+    const oldNull = R.isNil(oldData);
+    const newNull = R.isNil(newData);
```
alexcjohnson: `isNil` catches both `undefined` and `null` - and mostly that distinction is moot because `undefined` can't be sent via JSON, I guess. But we should be careful about it in objects, like `{a: null} <-> {}`

src/components/Store.react.js
```diff
@@ -2,27 +2,36 @@ import R from 'ramda';
...
             return true;
         }
-        for (let i = 0; i < data.length; i++) {
-            if (dataCheck(data[i], old[i])) {
+        for (let i = 0; i < newData.length; i++) {
+            if (dataChanged(newData[i], oldData[i])) {
                 return true;
             }
         }
     } else if (R.contains(type, ['String', 'Number'])) {
```
alexcjohnson: `'Boolean'`?

T4rk1n: Forgot about `Boolean` in all this type checks, thanks!

alexcjohnson:
> I added a type check ... But it failed the test.

Can you say more? I don't see how that makes sense and I worry about remaining strange cases like https://github.com/plotly/dash-core-components/issues/422#issuecomment-450278298 if this is omitted.

T4rk1n:
> Can you say more?

Failure: `AssertionError: '1' != '{"data": [1, 2, 3]}'`

T4rk1n:
@alexcjohnson Please review.

test/test_integration.py
```diff
@@ -1580,3 +1581,62 @@ def hello(data):
...
+            ('str', 'hello'),
+            ('number', 1),
+            ('dict', {'data': [2, 3, 4]}),
+            ('list', [5, 6, 7]),
+            ('null', None),
+            ('bool', True),
+            ('bool', False),
+            ('empty-dict', {}),
+            ('list-dict-1', [1, 2, {'data': [55, 66, 77], 'dummy': 'dum'}]),
+            ('list-dict-1', [1, 2, {'data': [111, 99, 88]}]),
```
alexcjohnson: Can you add the pathological items in my comments https://github.com/plotly/dash-core-components/issues/422#issuecomment-450278298 and https://github.com/plotly/dash-core-components/issues/422#issuecomment-450279988? I think you fixed all those, but here's another one to add, that I think still fails:
```js
dataChanged({a: 1, b: null}, {a: 1, c: 1}) // returns false, should be true
```

This is due to `isNil` conflating `undefined` and `null` - come to think of it, is there really a reason to treat those as equivalent? Perhaps we can just change the null check to:
```js
if (oldNull || newNull) {
    return oldData !== newData; // which will distinguish null/undefined
}
```

T4rk1n:
@alexcjohnson Added the undef edge case and checking the data instead of the null fixed it. I think we're good now.

PR #435, opened 2019-01-11
Issue434 - Clear date in DatePickerSingle and DatePickerRange

### LINKED ISSUE #434: Clear DatePickerSingle from callback?
**➥ Is it possible to clear the `date` from a `dcc.DatePickerSingle` via a callback?**

---
**Background**  
Say I have a date picker, and a button. I want to clear the selected date when I click on the button. (Ok, this looks a lot like a `clearable=True` DatePickerSingle, but my real case is of course a bit more complex).

**Problem**  
I have tried to return either `''` (an empty string) or `None` in the callback, but neither seems to work: each time, the select date disappear, but I get an "Invalid date" placeholder and the calendar dialog doesn't appear anymore.

screenshot from 2019-01-11 11-24-54

**MWE**  
```
# import external modules
import dash
import pytz

# import external functions
from datetime import datetime, timedelta
from dateutil.parser import parse

def serve_layout():

    return html.Main(
        [
            dcc.DatePickerSingle(
                id='datepicker',
            # layout settings
                first_day_of_week=1,
                display_format='DD/MM/YY',
                placeholder='Start date',
            # date defaults
                date=datetime.now(),
            # date limits
                min_date_allowed=datetime(2018, 1, 1),
                # today
                max_date_allowed=datetime.now(),
                initial_visible_month=datetime(
                    datetime.now().year,
                    datetime.now().month,
                    1
                ),
            ),
            html.Button(
                'Clear date',
                id='button'
            )
        ],
    )

app = dash.Dash(
    __name__,
)

# re-compute the layout everytime the page is refreshed — cf. https://dash.plot.ly/live-updates
app.layout = serve_layout

@app.callback(
    Output('datepicker', 'date'),
    [
        Input('button', 'n_clicks'),
    ],
    [
        State('datepicker', 'date'),
    ]
)
def clear_date(n_clicks, current_selected_date):
    ''' clears the date when button is clicked'''

    if (n_clicks is not None) and (n_clicks > 0):
        # =============================== neither of both following lines work 
        # return ''
        return None
    else:
        return current_selected_date


if __name__ == '__main__':
    app.run_server(debug=True)

```

Marc-Andre-Rivet:
Fixes #434 

Currently, DatePickerSingle and DatePickerRange do not process `null` dates correctly.

Handle the null case by assigning the value as-is to the underlying react-dates component instead of wrapping it with moment.


dash_core_components/metadata.json
```diff
@@ -2789,6 +2789,9 @@
             },
             {
               "name": "string"
+            },
+            {
+              "name": "bool"
```
Marc-Andre-Rivet: In some previous commit the metadata was not updated correctly.

src/index.js
```diff
@@ -22,6 +22,8 @@ import Tab from './components/Tab.react';
 import Store from './components/Store.react';
 import LogoutButton from './components/LogoutButton.react';
 
+import './components/css/react-dates@12.3.0.css';
```
Marc-Andre-Rivet: Moving this style dependency to the top level -- jest is not able to handle css files correctly atm

src/utils/convertToMoment.js
```diff
@@ -0,0 +1,26 @@
...
+    const dest = {};
+
+    momentProps.forEach(key => {
+        const value = newProps[key];
+
+        switch (R.type(value)) {
+            case 'Undefined':
+                return;
+            case 'Null':
+                dest[key] = null;
```
Marc-Andre-Rivet: Common logic between Single and Range + the new `Null` case

Issue #325, opened 2018-10-11
disabled prop for dcc.Interval is not used.


T4rk1n:
We have a `disabled` prop in the prop type definition of the Interval component, but it is not actually used by the code to disable the interval. Either remove it as we have the max_intervals=0 logic to stop the interval or make it stop the interval.

https://community.plot.ly/t/interval-component-cannot-be-disabled-via-callback/14455

Proposed solution:
- [ ] replace the props argument in setInterval by `this.props`.
- [ ] remove the disabled check from the interval.
- [ ] check for disabled prop in `componentWillReceiveProps`.stop the loop if false .
- [ ] restart the loop on disabled = True in componentWillReceiveProps if it was stopped.

T4rk1n:
Actually, we check for disabled only while doing setProps in the interval, but it takes the props from arguments and not `this` so it isn't updated. It's also the wrong place for that check as the loop will still run after setting disabled.

The disabled props should actually be checked in `componentWillReceiveProps` and stop the loop if false.

shammamah:
@T4rk1n By "stop the loop", do you mean that `window.clearInterval` should be called? Or is there some other behaviour that should occur?

T4rk1n:
Yes that is what I meant.

philip-peterson:
This issue is addressed by https://github.com/plotly/dash-core-components/pull/436

alexcjohnson:
Closed by #436 - thanks @philip-peterson 🎉

PR #436, opened 2019-01-16
Tests and tweak for Interval component

philip-peterson:
Hi! Thank you for making Dash, it is a very helpful platform for quickly creating visualizations. I was using the Interval component the other day and found out that there is some trouble with the `disable` prop (I made an example to show how). So I added some tests and made some tweaks to the Interval component, hopefully the fix is useful to you!

Cheers,
- p


alexcjohnson:
Thanks @philip-peterson! The fix looks good, and quite a bit clearer code too 🎉

I linted and rebuilt (we're intending to change the process so built files are published only with releases, but for now we're building in every PR). That said, `eslint` is still complaining about the [`console.warn`](https://github.com/plotly/dash-core-components/pull/436/files#diff-f65765bc3238f99a4990e2117b92bc0fR19) on a redundant `startTimer` call. I think we have that rule in place because console messages are not a level of debugging we want to persist in the final code - either a) this is a severe enough problem that it should throw an actual error, b) it's not an error at all and we should just return silently, or c) we should treat this as equivalent to `resetTimer`, and clear and restart the interval. I don't think (a) makes sense, probably I'd go for (b) but could be convinced for (c).

FYI I should just note that events are about to disappear (see https://github.com/plotly/dash/issues/531) but I'll take care of reconciling that with the work and tests in this PR after merging.

philip-peterson:
Hey Alex, thanks for taking a look at this so quickly! I thought about it a bit, and actually I think option (A) is what would make the most sense. Since we are talking about a situation that should never occur -- the code is written in a way that should not allow it to happen -- we would want to know if something has gone wrong, because it means something in the code isn't written the way it was intended. That eliminates option B. As for option C, if someone is managing to go down this path, it is by accident (either on their part or ours), and the last thing a developer wants alerting them of a mistake that they made is the presence of some subtle, slightly different behavior from what constitutes normal operation. So that would rule out option (C).

The argument against option (A) I can see as being that, well, if the component is broken then we should at least try to proceed with execution. But, I would counter that attempting to proceed with executing code known to be broken is doomed to fail.

Let me know what you think of all that. Pushed up a couple new commits to fix that lint issue mentioned and also ran prettier on all the tests. Thanks again!

alexcjohnson:
OK, as long as we have sufficient test coverage to be comfortable that this error is never thrown, which it looks like we do, I can support throwing an error.

@Marc-Andre-Rivet this is a 💃 from my standpoint, any comments before I merge it? (I added a changelog entry just now FYI)

Marc-Andre-Rivet:
💃. Thanks for reaching out to us with these fixes and improvements @philip-peterson!

philip-peterson:
Thanks @Marc-Andre-Rivet  and @alexcjohnson ! This ticket can be marked as solved btw: https://github.com/plotly/dash-core-components/issues/325

PR #439, opened 2019-01-18
Update R dashCoreComponents package to 0.43.0

### LINKED ISSUE #357: R version

rpkyle:
The JSON >> R 'transpiler' has been rewritten in Python, and was used to generate an installable R package for version 0.43.0 of `dash-core-components`. This PR closes #357.

@alexcjohnson @nicolaskruchten @T4rk1n @Marc-Andre-Rivet


dash_core_components/Location.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, id=Component.REQUIRED, pathname=Component.UNDEFINED, search=C
         _locals.update(kwargs)  # For wildcard attrs
         args = {k: _locals[k] for k in _explicit_args if k != 'children'}
 
-        for k in ['id']:
+        for k in [u'id']:
```
alexcjohnson: Where did this `u` (and in `Store` below) come from? Is this intentional, or a result of running the build script in py2 instead of py3?

rpkyle: Just to wrap this up, the crux of the difference is that Python 3 does away with `unicode` and `str` is essentially `unicode` by default; in Python 2, importing via `json.load` will return Unicode strings. There are workarounds, and the one I've chosen to use is Mark Amery's  [`byteify`](https://stackoverflow.com/questions/956867/how-to-get-string-objects-instead-of-unicode-from-json/13105359#13105359) function.

dash_core_components/_imports_.py
```diff
@@ -1,47 +1,47 @@
...
-    "Slider",
-    "Store",
-    "SyntaxHighlighter",
-    "Tab",
-    "Tabs",
     "Textarea",
-    "Upload"
+    "Tabs",
+    "Slider"
 ]
```
T4rk1n: Generated without newline.

dash_core_components/metadata.json
```diff
@@ -1,3690 +1 @@
...
-      {
-        "name": "clearTimer",
-        "docblock": null,
-        "modifiers": [],
-        "params": [],
-        "returns": null
-      },
-      {
-        "name": "handleInterval",
-
```
T4rk1n: Regenerated without indentation.

R/coreChecklist.R
```diff
@@ -0,0 +1,18 @@
...
+        type = 'Checklist',
+        namespace = 'dash_core_components',
+        propNames = c('labelClassName', 'style', 'inputClassName', 'inputStyle', 'labelStyle', 'dashEvents', 'options', 'className', 'values', 'fireEvent', 'id', wildcard_names),
+        package = 'dashCoreComponents'
+        )
+
+    component$props <- filter_null(component$props)
+    
+    structure(component, class = c('dash_component', 'list'))    
+}
```
T4rk1n: Generated without newline at end of file.

rpkyle:
OK, I've regenerated the components; if you could take a look and see if we're ready to merge, that'd be great.

@T4rk1n @alexcjohnson

R/coreChecklist.R
rpkyle: The newline isn't required for R scripts.

alexcjohnson: I don't think newline is required for *any* language we use, yet in some cases we actually set our linters to force it. Why?
- It's sometimes useful on the command line (otherwise if you `cat` a file for some reason you'll find your next prompt at the end of the last line of the file, rather than on a new line)
- In some editors, the last line may get intermittently covered up by a horizontal scroll bar
- Github adds 🚫in diffs and nobody wants to annoy Github!

I don't see this as a big deal (unless you have anything to add to this list @T4rk1n), but it's mildly nice and - unless there's a good reason not to - probably easier to just add the newline than to argue about it 😅

T4rk1n: No, it's not really important, just a pet peeves.

rpkyle:
> The props order for python component shouldn't change.

@T4rk1n If you could look over the recent commit, that'd be great. Hopefully we're nearing :dancer:.

R/coreChecklist.R
rpkyle: Good points, easy to modify the transpiler to add in the newlines next time around!

PR #444, opened 2019-01-23
No events

### LINKED ISSUE #433: Remove Event system
Companion issue to https://github.com/plotly/dash/issues/531

Events are used only lightly in this repo. The changes I see to remove them completely:

- `Graph` has events, and corresponding properties with the data, for things like click events, hover events... I don't see any need to use the events instead of just letting the property trigger the callback, but folks using the events would need to alter their code. Do we need to add timestamps for these, or is the existence of the data properties enough?

- A bunch of components expose `change` events, fired with every character typed for example. The underlying value will also trigger a callback, I believe in all of these cases, so I don't think we're losing any functionality, but we may want to add timestamps - to *all* core components in fact - to help distinguish most-recent-user-input, for cases of multiple parallel ways to update the same output, as described in https://github.com/plotly/dash/issues/531

- `Input` already has `n_blur[_timestamp]` and `n_submit[_timestamp]` with no corresponding events. It doesn't even have a `change` event, but the note above ^^ still may apply to it.

- `Interval` currently fires an event *and* increments `n_interval` - so we simply need to remove the event. No timestamp property is provided, would be an easy add if anyone was interested.

- `Textarea` currently exposes `blur` and `click` events, which need to be updated to properties.

alexcjohnson:
Closes #433 - last in the series for https://github.com/plotly/dash/issues/531

In addition, as in the other no-events PRs, I tried to unify the local and CI test runners, and make it so on CI all tests run, even if an earlier one fails, so you get a complete failure report - see https://github.com/plotly/dash-core-components/commit/3f0b9dfef9807621ffb977ed5ef947ad3fd0fd96, @T4rk1n @Marc-Andre-Rivet are you comfortable with this?


alexcjohnson:
@T4rk1n the `dcc` css test:

https://github.com/plotly/dash-core-components/blob/674fd072f0349053ed5604514ca152380e2a65dc/test/test_integration.py#L1421-L1429

is failing, seems to be due to https://github.com/plotly/dash/pull/547 - can you take a look at it, see if it's just the test that needs updating or if there's a real issue with the new `assets_folder` logic? The same issue that percy caught in this PR, that the extra css is not showing up, I see it locally running this test both on master and on this branch.

dash_core_components/package.json
```diff
@@ -62,6 +65,7 @@
     "exec-sh": "^0.3.0",
     "jest": "^23.6.0",
     "lodash": "^4.17.11",
+    "npm-run-all": "^4.1.5",
```
Marc-Andre-Rivet: I do so love this package 🥇

Marc-Andre-Rivet:
- [Interval tests](https://github.com/plotly/dash-core-components/blob/no-events/test/unit/Interval.test.js) still contain `fireEvent` logic
- Technically out of scope, but while updating the test scripts, could you change npm's `lint` command to do `eslint src test` instead of just src

Marc-Andre-Rivet:
Test run in dash-docs seems to be behaving fine. Once the ☝️ are fixed I'm fine with this PR.

alexcjohnson:
> Can you confirm what version of `dash` is used in your environment during the components py file generation?

Right, this needs to be built off the `no-events` branch of `dash` https://github.com/plotly/dash/pull/550. Also I believe it needs to be built with py3 (which you're doing anyway) since that branch doesn't yet have @rpkyle's `byteify` fix from https://github.com/plotly/dash/pull/545

dash_core_components/package.json
```diff
@@ -13,7 +13,10 @@
     "publish-all": "node scripts/publish.js",
     "start": "webpack-serve ./webpack.serve.config.js --open",
     "lint": "eslint src",
-    "test": "python -m unittest test.test_integration",
+    "lint:py": "flake8 --ignore=E501,F401,F841,F811 test",
+    "test": "run-s -c test-unit test:py test:pyimport lint format:test lint:py",
+    "test:py": "python -m unittest test.test_integration",
+    "test:pyimport": "python -m unittest test.test_dash_import",
```
T4rk1n: Can combine them with `python -m unittest discover`

alexcjohnson: A good to know! But `test_dash_import` is funny, since it rewrites files (specifically, making a `dash.py` in the test folder) you can't run it in conjunction with any of the other tests. So if we wanted to use `discover` we'd have to make `test_dash_import` undiscoverable one way or another, and still run it separately.

T4rk1n: Hmm, that test is strange, not sure we really need to test that.

alexcjohnson:
> [Interval tests](https://github.com/plotly/dash-core-components/blob/no-events/test/unit/Interval.test.js) still contain `fireEvent` logic

Ah good call, that came in with a community PR after my initial sweep through the code... fixed in https://github.com/plotly/dash-core-components/pull/444/commits/7d822569112d898efe07ec9930ee8077856aa993

> Technically out of scope, but while updating the test scripts, could you change npm's `lint` command to do `eslint src test` instead of just src -- and fix the associated lint errors

Sure, easy enough -> https://github.com/plotly/dash-core-components/pull/444/commits/948c7dc94ba01d53ea16d0812b3597a78b6e64e0

dash_core_components/package.json
alexcjohnson: I guess someone put that error message in the code, so they wanted to see it in a test. But yeah, not the most useful test to be lugging around.

src/components/Textarea.react.js
```diff
@@ -28,26 +28,37 @@ export default class Textarea extends Component {
...
             />
         );
     }
 }
 
+Textarea.defaultProps = {
+    n_blur: 0,
+    n_blur_timestamp: -1,
+    n_clicks: 0,
+    n_clicks_timestamp: -1,
```
T4rk1n: :smiley_cat: This addition of props should be noted somewhere, had to read the code to see it.

alexcjohnson: Thanks @T4rk1n -> added to `CHANGELOG` in https://github.com/plotly/dash-core-components/pull/444/commits/8b4936b62534bf4fc539c269a070c824d3337dc3

Kind of amazing that these two are the only events we had that weren't covered by an existing regular prop already.

PR #478, opened 2019-03-05
Rework dash-renderer setProps

Marc-Andre-Rivet:


dash_core_components/package.json
```diff
@@ -73,6 +73,7 @@
     "prettier": "^1.14.2",
     "react": "^16.6.1",
     "react-dom": "^16.6.1",
+    "react-dot-fragment": "^0.2.8",
```
Marc-Andre-Rivet: React 16 ```<Fragment />``` shim

src/components/ConfirmDialogProvider.react.js
```diff
@@ -31,20 +32,18 @@ export default class ConfirmDialogProvider extends React.Component {
...
-            });
+        const wrapClick = child => {
+            const props = clone(child.props);
+            props.__dashlayout__.props.onClick = () => {
+                const update = {displayed: true};
+                this.setState(update);
+                if (setProps) {
+                    setProps(update);
+                }
+            };
```
Marc-Andre-Rivet: Like before, inserting `onClick` prop for the real child to use

src/components/ConfirmDialogProvider.react.js
```diff
@@ -53,9 +52,9 @@ export default class ConfirmDialogProvider extends React.Component {
                     (loading_state && loading_state.is_loading) || undefined
                 }
             >
-                {realChild && realChild.length
-                    ? realChild.map(wrapClick)
-                    : wrapClick(realChild)}
+                {children && children.length
+                    ? children.map(c => wrapClick(c))
+                    : wrapClick(children)}
```
Marc-Andre-Rivet: The child is not nested -- operate on the tree container directly

alexcjohnson: Why the change from `children.map(wrapClick)` to `children.map(c => wrapClick(c))`?

Also, not new but perhaps `Array.isArray(children)` would be better than `children && children.length`.

test/unit/Tabs.test.js
```diff
@@ -177,12 +177,12 @@ describe('Tabs handle Tab selection logic', () => {
...
     test('Tab can be clicked and will display its content', () => {
         tabs.find('[value="tab-2"]').simulate('click');
-        const renderedContent = tabs.find('.tab-content > div').html();
+        const renderedContent = tabs.find('.tab-content div').html();
         expect(renderedContent).toEqual('<div>Tab 2 child</div>');
     });
     test('Tab without value will still be clickable', () => {
         tabs.find('[value="tab-2"]').simulate('click');
-        const renderedContent = tabs.find('.tab-content > div').html();
+        const renderedContent = tabs.find('.tab-content div').html();
```
alexcjohnson: Is this change still necessary? Or did `<Fragment>` restore the previous structure?

src/components/ConfirmDialogProvider.react.js
Marc-Andre-Rivet: Guessing I did a console.log at some point and didn't revert it back correctly. Reverting.

test/unit/Tabs.test.js
```diff
@@ -177,12 +177,12 @@ describe('Tabs handle Tab selection logic', () => {
     });
     test('Tab can be clicked and will display its content', () => {
         tabs.find('[value="tab-2"]').simulate('click');
-        const renderedContent = tabs.find('.tab-content > div').html();
+        const renderedContent = tabs.find(Tab).html();
```
Marc-Andre-Rivet: Checking for the Tab component instead of some arbitrary selector

test/unit/Tabs.test.js
```diff
@@ -177,12 +177,12 @@ describe('Tabs handle Tab selection logic', () => {
...
         expect(renderedContent).toEqual('<div>Tab 2 child</div>');
     });
     test('Tab without value will still be clickable', () => {
         tabs.find('[value="tab-2"]').simulate('click');
-        const renderedContent = tabs.find('.tab-content > div').html();
+        const renderedContent = tabs.find(Tab).html();
```
Marc-Andre-Rivet: This test makes no sense. It's the exact copy of the one above with a different name. It's clearly not testing that case.. removing.

test/unit/Tabs.test.js
```diff
@@ -177,12 +177,7 @@ describe('Tabs handle Tab selection logic', () => {
     });
     test('Tab can be clicked and will display its content', () => {
         tabs.find('[value="tab-2"]').simulate('click');
-        const renderedContent = tabs.find('.tab-content > div').html();
-        expect(renderedContent).toEqual('<div>Tab 2 child</div>');
-    });
-    test('Tab without value will still be clickable', () => {
-        tabs.find('[value="tab-2"]').simulate('click');
-        const renderedContent = tabs.find('.tab-content > div').html();
```
Marc-Andre-Rivet: This test is an exact copy of the test above. Looks like copy-paste that was never modified. Removing.

test/unit/Tabs.test.js
Marc-Andre-Rivet: See final change below

PR #481, opened 2019-03-07
Issue479 - Fix Vertical Slider regression

### LINKED ISSUE #479: Vertical Slider has stopped working
Hello,
Vertical Sliders have stopped working (I think this happened in the most recent update, but am not certain)
However, vertical RangeSlider's still work in vertical mode.
Thanks,
Peter

```python
import dash

app = dash.Dash(__name__)


    html.H1(children='Vertical Slider has stopped working.'),
    html.Div([
        dcc.Slider(
            id='mhd_window_slider',
            min=0.05,
            max=1.4,
            step=0.005,
            value=0.5,
            vertical=True,
        )
    ], style={ 'width':'50px', 'height':'500px' }),
    
    html.H1(children='Vertical RangeSlider still works.'),
    html.Div([
        dcc.RangeSlider(
            id='mhd_clim_slider',
            count=1,
            min=0,
            max=3,
            step=0.01,
            value=[0.02, 1],
            vertical=True,
        )
    ], style={ 'width':'50px', 'height':'500px' }),
])

if __name__ == '__main__':
    app.run_server(debug=True)
```

Marc-Andre-Rivet:
Fixes #479 

The test fails without the fix: https://circleci.com/gh/plotly/dash-core-components/3577
```
Exception: ('browser error logged during test', {'level': 'SEVERE', 'message': "http://localhost:8050/_dash-component-suites/dash_renderer/react-dom@15.4.2.min.js?v=0.20.0&m=1551989220 13:5540 Uncaught TypeError: Cannot read property 'toFixed' of undefined", 'source': 'javascript', 'timestamp': 1551989561259})
```

### DIFF
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -26,7 +26,7 @@ export default class Slider extends Component {
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
-                style={{display: 'initial'}}
+                style={{display: 'inline'}}
             >
                 <ReactSlider
                     onChange={value => {
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -26,7 +26,7 @@ export default class Slider extends Component {
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
-                style={{ display: 'initial' }}
+                style={{display: 'initial'}}
             >
                 <ReactSlider
                     onChange={value => {
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -26,7 +26,7 @@ export default class Slider extends Component {
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
-                style={vertical ? { height: '100%' } : {}}
+                style={vertical ? {height: '100%'} : {}}
             >
                 <ReactSlider
                     onChange={value => {
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -18,7 +18,7 @@ export default class Slider extends Component {
     }
 
     render() {
-        const {id, setProps, updatemode, loading_state} = this.props;
+        const {id, loading_state, setProps, updatemode, vertical} = this.props;
         const {value} = this.state;
         return (
             <div
@@ -26,7 +26,7 @@ export default class Slider extends Component {
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
-                style={{display: 'inline'}}
+                style={vertical ? { height: '100%' } : {}}
             >
                 <ReactSlider
                     onChange={value => {
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -26,6 +26,7 @@ export default class Slider extends Component {
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
+                style={{ display: 'initial' }}
             >
                 <ReactSlider
                     onChange={value => {
--- a/test/IntegrationTests.py
+++ b/test/IntegrationTests.py
@@ -14,35 +14,6 @@ from selenium.webdriver.chrome.options import Options
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 
 
-class WebdriverLogFacade(object):
-
-    last_timestamp = 0
-
-    def __init__(self, webdriver):
-        self._webdriver = webdriver
-
-    def get_log(self):
-        last_timestamp = self.last_timestamp
-        entries = self._webdriver.get_log("browser")
-        filtered = []
-
-        for entry in entries:
-            # check the logged timestamp against the
-            # stored timestamp
-            if entry["timestamp"] > self.last_timestamp:
-                filtered.append(entry)
-
-                # save the last timestamp only if newer
-                # in this set of logs
-                if entry["timestamp"] > last_timestamp:
-                    last_timestamp = entry["timestamp"]
-
-        # store the very last timestamp
-        self.last_timestamp = last_timestamp
-
-        return filtered
-
-
 class IntegrationTests(unittest.TestCase):
 
     @classmethod
@@ -64,7 +35,6 @@ class IntegrationTests(unittest.TestCase):
         )
         cls.percy_runner = percy.Runner(loader=loader)
         cls.percy_runner.initialize_build()
-        cls.log_facade = WebdriverLogFacade(cls.driver)
 
     @classmethod
     def tearDownClass(cls):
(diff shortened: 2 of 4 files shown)


src/components/Slider.react.js
```diff
@@ -26,6 +26,7 @@ export default class Slider extends Component {
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
+                style={{display: 'initial'}}
```
Marc-Andre-Rivet: overrides browser default styling `display: block` for default div behavior -- this allows the child to have the correct size

test/IntegrationTests.py
```diff
@@ -11,6 +11,36 @@
...
+
+                # save the last timestamp only if newer
+                # in this set of logs
+                if entry["timestamp"] > last_timestamp:
+                    last_timestamp = entry["timestamp"]
+
+        # store the very last timestamp
+        self.last_timestamp = last_timestamp
+
+        return filtered
```
Marc-Andre-Rivet: To be called once per test (at least once in the teardown phase) to get the test-related browser errors

test/IntegrationTests.py
```diff
@@ -20,18 +50,21 @@ def setUpClass(cls):
...
 
         options = Options()
+        capabilities = DesiredCapabilities.CHROME
+        capabilities['loggingPrefs'] = {'browser': 'SEVERE'}
 
         if 'DASH_TEST_CHROMEPATH' in os.environ:
             options.binary_location = os.environ['DASH_TEST_CHROMEPATH']
 
-        cls.driver = webdriver.Chrome(chrome_options=options)
+        cls.driver = webdriver.Chrome(options=options, desired_capabilities=capabilities)
```
Marc-Andre-Rivet: Settings to get browser errors

test/test_integration.py
```diff
@@ -217,6 +217,34 @@ def test_upload_gallery(self):
...
+        self.wait_for_element_by_css_selector('#vertical-slider')
+        self.snapshot('vertical slider')
+
+        v_slider = self.driver.find_element_by_css_selector(
+            '#vertical-slider div[role="slider"]'
+        )
+        v_slider.click()
+
+        for entry in self.log_facade.get_log():
+            raise Exception('browser error logged during test', entry)
```
Marc-Andre-Rivet: If there's a message here, something went wrong. Not doing this for all tests as lots of tests have issues at the moment..

Marc-Andre-Rivet: https://github.com/plotly/dash-core/issues/51

src/components/Slider.react.js
alexcjohnson: 🤔 IE11? https://developer.mozilla.org/en-US/docs/Web/CSS/initial#Browser_compatibility

A little fiddling indicates that for `display`, `initial` means `inline`? Can we use that?

test/IntegrationTests.py
alexcjohnson: I really like this solution - much cleaner than the `window.tests` hack, and (cc @valentijnnieman) it will catch errors during page load as well, presumably.

I'm curious though, why a separate class, rather than just a method of `IntegrationTests`?

test/IntegrationTests.py
```diff
@@ -20,18 +50,21 @@ def setUpClass(cls):
         super(IntegrationTests, cls).setUpClass()
 
         options = Options()
+        capabilities = DesiredCapabilities.CHROME
+        capabilities['loggingPrefs'] = {'browser': 'SEVERE'}
```
alexcjohnson: I tried in vain to find docs for this, did you? If this config does what's needed here that's fine for this PR, I'm just curious if we might like other settings in other cases, to get messages that don't quite rise to the level of SEVERE :rage4:

Marc-Andre-Rivet: https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities

Available values for most loggers are "OFF", "SEVERE", "WARNING", "INFO", "CONFIG", "FINE", "FINER", "FINEST", "ALL".

test/IntegrationTests.py
Marc-Andre-Rivet: No specific reason. Found it while looking for the capabilities finer points in Python and liked it. Did not stop to consider if it made more or less sense to have it standalone. I can imagine wanting to make this richer and keeping it in a class sort of marginally makes sense in that context.

src/components/Slider.react.js
Marc-Andre-Rivet: Oh.. did not realize IE11 did not support `initial`. Thanks for checking it out.

test/IntegrationTests.py
alexcjohnson: mmm yeah, I saw that one, but it also says

> See the documentation of each driver for what browser specific logging components are available.

Which I was hoping existed somewhere and would tell us something about what kind of messages are included in each level... oh well, trial and error I guess.

src/components/Slider.react.js
valentijnnieman: Shouldn't the display be inline-block anyway, like here? https://github.com/plotly/dash-core-components/blob/8ccbb40389e5f9eb714ca3b2b1fce87a77ab7fec/src/components/DatePickerRange.react.js#L130

test/IntegrationTests.py
valentijnnieman: Looks good :-) I think it would make more sense to me to have it as a method if `IntegrationTests` instead of a separate class - just to keep everything in one place.

test/test_integration.py
valentijnnieman: I'm wondering if this test shouldn't be in the `test_gallery` test, since that's where most of the snapshots of components are taken. Although, I would rather see the `test_gallery` test eventually be broken up into specific tests per component, so I guess this is a good start! :-)

src/components/Slider.react.js
Marc-Andre-Rivet: So.. realizing that maybe just doing ```style={vertical ? { height: '100%' } : {}}``` on the wrapper div might be just as effective.

test/test_integration.py
Marc-Andre-Rivet: I think the gallery snapshot approach is a bad idea in general. When I see a diff I want that diff to be isolated and linked to as little dependencies as possible. If the vertical slider breaks, it's obvious. And no wondering if there's multiple issues in the single diff.

alexcjohnson: Agreed - so many times I've had to look at a diff where something was added in the middle of a long page... and percy has no idea that it's all the same content just shifted down, so I have to try and do that diffing in my head. Much better to break these up into single-purpose images, and not worry about making a huge total number of images.

test/IntegrationTests.py
Marc-Andre-Rivet: Sure. I can move this into IntegrationTests

PR #482, opened 2019-03-08
Update dashCoreComponents package in R to 0.44.0

rpkyle:
In addition to updating `dash_core_components.min.js` and `dash_core_components.dev.js` to version 0.44.0, and including `plotly-1.45.0.min.js`, this PR also aims to incorporate support for loading_state in "transpiled" dashR components.

Some simplification of the component generation statements in `package.json` is proposed as well.


package.json
```diff
@@ -26,8 +26,8 @@
     "uninstall-local": "pip uninstall dash-core-components -y",
     "build:js": "webpack --mode production",
     "build:js-dev": "webpack --mode development",
-    "build:py": "node ./extract-meta src/components > dash_core_components/metadata.json && cp package.json dash_core_components && dash-generate-components ./src/components dash_core_components",
-    "build:r": "node ./extract-meta src/components > dash_core_components/metadata.json && cp package.json dash_core_components && dash-generate-components ./src/components dash_core_components --r-prefix 'core'",
+    "build:py": "dash-generate-components ./src/components dash_core_components",
+    "build:r": "dash-generate-components ./src/components dash_core_components --r-prefix 'dcc'",
```
alexcjohnson: Again, let's 🔪 `extract-meta`

man/dccChecklist.Rd
```diff
@@ -10,7 +10,7 @@ Checklist is a component that encapsulates several checkboxes. The values and la
 }
 
 \usage{
-coreChecklist(id=NULL, options=NULL, values=NULL, className=NULL, style=NULL, inputStyle=NULL, inputClassName=NULL, labelStyle=NULL, labelClassName=NULL, ...)
+dccChecklist(id=NULL, options=NULL, values=NULL, className=NULL, style=NULL, inputStyle=NULL, inputClassName=NULL, labelStyle=NULL, labelClassName=NULL, loading_state=NULL, ...)
```
alexcjohnson: These ones still have `...` at the end, but looks like the components themselves lost that a while back?

package.json
rpkyle: fixed in 2a7808473d90376f8487e18acb2ed12d429463c8

man/dccChecklist.Rd
rpkyle: This has been fixed in plotly/dash#638; when that PR is merged, `...` will only appear when wildcards are supported by the component.

rpkyle: The components being merged here were 'transpiled' using the correct code from the aforementioned PR ☝️, so the stray ellipsis has been :hocho:.

Issue #197, opened 2018-05-08
add support for `restyle` event, similar to `relayout` event


chriddyp:
The `restyle` event is triggered when the legend is clicked, which might be useful for preserving state between re-draws: https://community.plot.ly/t/preserve-trace-visibility-in-callback/5118/4

Vayel:
Also useful for `Parcoords` plots: https://community.plot.ly/t/displey-selected-data-in-parcoords/4055

PR #483, opened 2019-03-09
Modified Graph.react.js to include restyle event

mako-npm:
https://github.com/plotly/dash-core-components/issues/197
https://community.plot.ly/t/interactive-callback-on-legend-click/9236/6

Made changes to Graph.react.js to add support for restyle events, restyleData is now an available component_property for graph components in Dash. Toggling on/off legend items now triggers callbacks with the associated restyleData component property.


alexcjohnson:
Nice, thanks for the PR @mako-npm! A few things we'll need to finish this:

- [ ] Looks like the linter had a complaint, you can fix it with `npm run format`.
- [ ] I'd also like to add a test for `restyleData` - but as we don't have any graph event data tests yet, unless you're feeling adventurous I'll be happy to do that part next week.
- [ ] There are some percy test failures. They look like flaky tests rather than anything you did in this PR, but no time like the present to fix that! Again, I can take a look at this next week.

lo-ise:
Hey folks, huge thanks @mako-npm for this PR, I could really use this feature. I'm afraid I don't have the skills to help with the details of the PR, looks as though it didn't pass tests. Is there a plan to try and get this merged?

test/test_integration.py
```diff
@@ -640,7 +640,11 @@ def on_click(n_clicks):
                             'x': [1, 2, 3, 4],
                             'y': [4, 3, 2, 1]
                         }
-                    ]
+                    ],
+                    'layout': {
+                        'width': 700,
+                        'height': 450
+                    }
```
alexcjohnson: I'm not sure what to say about this one... I'm tempted to consider it a bug, but I don't know what we'd do to fix it. The issue - and I still don't understand why it's triggered intermittently here, but it's a real issue users will face - is that if you have an autosized graph that's initially drawn in a `display: none` container, it gets the default size. It keeps that default size when the container is displayed, because no redraw is triggered. But if the window is resized at all, THEN the graph gets sized correctly.

This seems to sometimes happen on CI - see eg https://percy.io/plotly/dash-core-components/builds/1576352?utm_campaign=plotly&utm_content=dash-core-components&utm_source=github_status - so I'm just locking down the size here.

test/test_integration.py
```diff
@@ -1196,8 +1200,8 @@ def _on_confirmed(submit_n_clicks, cancel_n_clicks,
                     return 'canceled'
 
         self.startServer(app)
-        self.snapshot(test_name + ' -> initial')
         button = self.wait_for_element_by_css_selector('#button')
+        self.snapshot(test_name + ' -> initial')
```
alexcjohnson: take the snapshot *after* we know the button exists

src/components/Graph.react.js
```diff
@@ -160,6 +160,14 @@ class PlotlyGraph extends Component {
                 }
             }
         });
+        gd.on('plotly_restyle', eventData => {
+            const restyleData = filterEventData(gd, eventData, 'restyle');
+            if (!isNil(restyleData)) {
+                if (setProps) {
```
T4rk1n: :hocho: Think we can remove these checks now ?

src/components/Graph.react.js
```diff
@@ -271,6 +279,12 @@ const graphPropTypes = {
      */
     relayoutData: PropTypes.object,
 
+    /**
+     * Data from latest restyle event which occurs
+     * when the user toggles a legend item
+     */
+    restyleData: PropTypes.object,
```
T4rk1n: Can we have a shape instead?

src/components/Graph.react.js
```diff
@@ -271,6 +279,12 @@ const graphPropTypes = {
      */
     relayoutData: PropTypes.object,
 
+    /**
+     * Data from latest restyle event which occurs
+     * when the user toggles a legend item
```
T4rk1n: It's not obvious to me if this prop is read-only or not. From reading the code I'd say changing it has no effect, so it's a read only and should be stated.

src/components/Graph.react.js
alexcjohnson: good call. In case @Marc-Andre-Rivet is already working on removing the others I'll just do this one for now.

src/components/Graph.react.js
alexcjohnson: Oh right, it’s actually a length-2 array of the args to `Plotly.restyle` - the first one an object of `<attribute.string>: <value>`, the second an integer or array of integers for the trace indices. So I’ll change it to array and describe this (yes, it’s also read-only) but it doesn’t look to me as though `PropTypes` allows us to do better other than via custom function, which wouldn’t do the user any good. Is that right or am I missing how to do that?

src/components/Graph.react.js
alexcjohnson: 🔪 in https://github.com/plotly/dash-core-components/pull/483/commits/8e7806b5135c653b6f3881a72487fc387fb8d3c6 - but as commented I only changed the new prop for now.

src/components/Graph.react.js
alexcjohnson: Changed to `array` and improved docs (including read-only on all the event data props) in https://github.com/plotly/dash-core-components/pull/483/commits/688a66f6c7ed59aea77321a37b5f6d7e180c7304

alexcjohnson:
Thanks @mako-npm for kicking this off! This will be part of our next release, which should come out in the next couple of days.

mako-npm:
@alexcjohnson Thanks for all the work on this, looking forward to seeing it!

PR #486, opened 2019-03-11
Issue 485 - Slider `className` / external styling regression

### LINKED ISSUE #485: Switch to 0.39 breaks Slider
Concerning this issue:
https://community.plot.ly/t/bug-slider-render-classname-on-wrong-element/20536

> As of a recent update I think, the className property is output to a different html on the Slider, resulting in the slider collapsing, while e.g. the RangeSlider renders completely fine.

> RangeSlider “col-10” is in the right place
> image

> Slider (surrounding div has no “col-10” as it is supposed to)
> 79dcf046202dd47a83de187814f3470d3587a9e4

It still happens. I can fix it simply by switching from 0.39 back to 0.38. Any hidden changes in the versioning 

It looks like this:
image
But should look like this:
image

Marc-Andre-Rivet:
Fixes #485 

- fix vertical styling for both slider and range slider
- fix loading state attribute for range slider
- added visual and basic functional test for both sliders
- added loading state attribute test for both sliders


src/components/Slider.react.js
```diff
@@ -18,14 +18,22 @@ export default class Slider extends Component {
...
+            vertical
+        } = this.props;
         const {value} = this.state;
         return (
             <div
                 id={id}
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
+                className={className}
```
Marc-Andre-Rivet: Apply class name to the wrapper. rc-slider applies very little styles to itself and relevant styles will be ok if applied to the parent (e.g. it doesn't explicitly set the color, font-size, etc.) unlike the dropdown.

alexcjohnson: Most of the other cases where we allow `className`, we also allow `style` - do we want to do that here?

alexcjohnson:
Can we 🔒 this with a test?

Does this fix (and #479 too I guess) need to be applied to `RangeSlider` too?

src/components/Slider.react.js
Marc-Andre-Rivet: No. This is because the wrapper was added with Loading Components and the wrapped slider did not support the style property. I don't mind adding support for styling here -- but will need to handle the existing  #481 style applied on the vertical slider.

Marc-Andre-Rivet:
This does not affect the range slider, as for some reason that slider (same 3rd party package) supported data-* props

src/components/Slider.react.js
Marc-Andre-Rivet: Fwiw, the rangeslider does not support `style` either.

alexcjohnson: Given that the main purpose of `className` is styling, it seems appropriate to add `style`. But as that's a feature rather than a bugfix, feel free to make an issue and defer for later if you don't want to do it in this PR.

alexcjohnson:
OK, I don't see what `data-*` has to do with these, but I'd feel better if we had these all in a test rather than just reasoning about it. Perhaps a snapshot with single and range sliders in both orientations and a little styling (via class, or inline if you add it here)

Marc-Andre-Rivet:
data-* lack of support was what caused the slider to be wrapped in the first place for the LC feature.

Yeah, writing some additional tests.

Marc-Andre-Rivet:
So.. digging a little bit more in all directions, turns out the range slider should have been wrapped too. LC attributes don't work there either. Not sure why the slider was tested/wrapped and not the range slider. This is the only 3rd party element that isn't wrapped, not sure how we came to the conclusion it was behaving correctly with LC..

test/test_integration.py
```diff
@@ -245,6 +312,112 @@ def test_vertical_slider(self):
         for entry in self.get_log():
             raise Exception('browser error logged during test', entry)
 
+    def test_loading_range_slider(self):
```
Marc-Andre-Rivet: For both slider and range slider, make sure that the horizontal / vertical version displays correctly using the previously used styling, that the slider nubs can be clicked without error and that loading state is applied correctly to the DOM element

test/test_integration.py
```diff
@@ -215,6 +215,69 @@ def test_upload_gallery(self):
...
+                value=5,
+            ),
+        ])
+
+        @app.callback(
+            Output('horizontal-slider', 'value'),
+            [Input('test-div', 'children')]
+        )
+        def delayed_value(children):
+            time.sleep(5)
```
alexcjohnson: I'd like to get away from this pattern of adding super long sleeps. Can we perhaps use a [`Condition`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Condition) or something, so you wait until the loading state is seen, then tell the callback it's OK to return?

test/test_integration.py
```diff
@@ -237,14 +241,16 @@ def test_loading_slider(self):
             [Input('test-div', 'children')]
         )
         def delayed_value(children):
-            time.sleep(5)
+            lock.acquire()
+            lock.release()
```
alexcjohnson: Very nice @Marc-Andre-Rivet !

@plotly/dash whenever you are tempted to add a `time.sleep` in tests, take a moment to think "is there a way to do it without a sleep?" A huge fraction of our test time is simply sleep statements, which are set super long in order to be robust.

A lot of the time we can solve this just by waiting for some condition (an element with the right contents, for example). This one is a bit trickier, we were using a sleep in the server process (where callbacks run) to ensure something had time to be observed in the browser-control process, but this commit shows a fairly simple pattern to do this with a `Lock` so that the server process stops waiting as soon as the browser has seen what it's looking for.

PR #490, opened 2019-03-18
[WIP] Dash component system support

T4rk1n:


dash_core_components/_checklist.py
```diff
@@ -0,0 +1,64 @@
...
+    inputStyle = ComponentProp('inputStyle', "{}", False)
+    inputClassName = ComponentProp('inputClassName', '', False)
+    labelStyle = ComponentProp('labelStyle', "{}", False)
+    labelClassName = ComponentProp('labelClassName', '', False)
+    loading_state = ComponentProp('loading_state', UNDEFINED, False)
+
+    def __init__(
+        self,
+        id=UNDEFINED,
+        options="[]",
```
alexcjohnson: @T4rk1n is it intentional that these empty objects are quoted (`"[]"`, `"{}"`) here and in `ComponentProp`?

T4rk1n: No, default props generation needs more testing.

dash_core_components/_graph.py
```diff
@@ -0,0 +1,141 @@
...
+            `true`(default={     frame: {
+            redraw: false,     },     transition: {
+            duration: 750,         ease: 'cubic-in-
+            out',     }, })
+        :param config: Plotly.js config options. See
+            https://plot.ly/javascript/configuration-options/
+            for more info.(default={     staticPlot: false,
+            editable: false,     edits: {
+            annotationPosition: false,         annotationTail:
+            false,         annotationText: false,
```
alexcjohnson: Hey re: https://github.com/plotly/dash-core-components/issues/514 I'm thinking of removing these defaults, just using `config: {}`, as well as writing some sort of test that the `PropTypes.shape` we give matches the available options in the current version of plotly.js. But in this PR the default seems to be the only info about what's allowed inside `config` that makes it into the Python docs. How do you think we should handle this?
- bring the shape into this docstring programmatically?
- require that this shape be described explicitly in the docstring for `graphPropTypes.config` (and then I'd write a test to verify that this is kept up-to-date)?
- something else?

T4rk1n: The default in the docstring come from `Graph.defaultProps`, if you change it to `{}` they will be removed.

> bring the shape into this docstring programmatically?

You cannot do that on at runtime with react-docgen as nothing is evaluated just parsed. It has to be in the source. Maybe a script that fetch and format and you insert the output in the docstring, I'd imagine you'd only have to do that when you update plotly.js.

> require that this shape be described explicitly in the docstring for graphPropTypes.config (and then I'd write a test to verify that this is kept up-to-date)?

I like that.

> something else?

Front-end validations and keep as is with the default `{}`?

T4rk1n: Come to think of it, the mypy validations for `shapes` is shallow and not key based, it might be worthwhile to format the PropType definition in the docstring.

alexcjohnson: OK cool - one way or another we can handle this in documentation. For now I'm going to take the defaults out for https://github.com/plotly/dash-core-components/issues/514

PR #496, opened 2019-03-23
[WIP] Renderer regression

Marc-Andre-Rivet:
May require a changelog entry if this and https://github.com/plotly/dash-renderer/pull/140 fixes the behavior encountered by community with dynamically loaded LCs. @valentijnnieman will want to go over that one with you.


.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ jobs:
                 name: Install dependencies (dash)
                 command: |
                     git clone git@github.com:plotly/dash.git
-                    git clone git@github.com:plotly/dash-renderer.git
+                    git clone -b renderer-regression git@github.com:plotly/dash-renderer.git
```
Marc-Andre-Rivet: To be removed once in master

src/components/Loading.react.js
```diff
@@ -37,9 +37,7 @@ export default class Loading extends Component {
             type,
         } = this.props;
 
-        const isLoading = getLoadingState(this);
-
-        if (isLoading) {
+        if (loading_state && loading_state.is_loading) {
```
Marc-Andre-Rivet: The nested loading state is already determined by the renderer. Only need to check the prop value

src/components/Loading.react.js
```diff
@@ -62,6 +61,8 @@ export default class Loading extends Component {
     }
 }
 
+Loading._dashprivate_deepstate = true;
```
Marc-Andre-Rivet: set the `deepstate` flag for nested loading state

src/components/Loading.react.js
```diff
@@ -126,23 +127,3 @@ Loading.propTypes = {
...
-    const children = element.props && element.props.children;
-    if (!children) {
-        return false;
-    }
-
-    return R.any(
-        child => child.type !== Loading && getLoadingState(child),
-        Array.isArray(children) ? children : [children]
-    );
-}
```
Marc-Andre-Rivet: This is resolved by the renderer, no longer needed

src/fragments/Loading/spinners/CircleSpinner.jsx
```diff
@@ -28,7 +28,7 @@ const CircleSpinner = ({
     return (
         <div style={style ? style : {}} className={spinnerClass}>
             {debugTitle}
-            <div className="dash-sk-circle">
+            <div className="dash-spinner dash-sk-circle">
```
Marc-Andre-Rivet: Extra class on the spinners, easier to target general spinner through css (and to automate tests)

src/fragments/Loading/spinners/GraphSpinner.jsx
```diff
@@ -19,18 +19,18 @@ const GraphSpinner = ({status, fullscreen, debug, className, style}) => {
         <div style={style ? style : {}} className={spinnerClass}>
             <div>
                 {debugTitle}
-                <div className="dash-spinner">
-                    <div className="dash-spinner__bottom">
+                <div className="dash-spinner dash-graph-spinner">
```
Marc-Andre-Rivet: dash-spinner and variants become dash-graph-spinner

test/test_integration.py
```diff
@@ -216,9 +216,306 @@ def test_upload_gallery(self):
...
+
+        @app.callback(
+            Output('div-1', 'children'),
+            [Input('root', 'n_clicks')]
+        )
+        def updateDiv(children):
+            with lock:
+                return 'content'
+
+        with lock:
```
Marc-Andre-Rivet: Using `with lock` instead of lock.acquire() / lock.release() as a test failure will throw an error and will prevent the lock from being released -- hence the test app will never terminate, blocking the tests indefinitely!

```
with lock:
    // Code
```

is equivalent to

```
try {
   lock.acquire()
    // Code
} finally {
    lock.release()
}

File: test/test_integration.py [hunk 8]
```diff
@@ -216,9 +216,306 @@ def test_upload_gallery(self):
 
         self.snapshot('test_upload_gallery')
 
+    def test_loading_component_initialization(self):
```
Marc-Andre-Rivet: 5 scenarios are tested here
- LC works on initial callback execution
- LC works on triggered callback
- multiple LCs have independent states
- nested LCs get pruned from parent's state
- dynamically loaded LCs work

test/test_integration.py
```diff
@@ -239,16 +536,15 @@ def test_loading_slider(self):
             [Input('test-div', 'children')]
         )
         def delayed_value(children):
-            lock.acquire()
-            lock.release()
-            return 5
+            with lock:
```
Marc-Andre-Rivet: Fixing tests previously written with lock.acquire() / lock.release()

test/unit/__snapshots__/Loading.test.js.snap
```diff
@@ -1,80 +1,9 @@
...
-                        transform: scaleY(0.4);
-                        -webkit-transform: scaleY(0.4);
-                        }  20% { 
-                        transform: scaleY(1.0);
-                        -webkit-transform: scaleY(1.0);
-                        }
-                    }
-            </style>"
-`;
+exports[`Loading checks all it's children for a loading_state: Loading spinner for children 1`] = `"<div>Child 1</div><div>Child 2</div><div loading_state=\\"[object Object]\\">Child 3</div>"`;
```
Marc-Andre-Rivet: new `dash-spinner` class changes the jest snapshots

test/test_integration.py
```diff
@@ -236,19 +534,29 @@ def test_loading_slider(self):
...
+            '#horizontal-slider:not([data-dash-is-loading="true"])'
         )
-        lock.release()
+
+        with lock:
+            self.driver.find_element_by_id('test-btn').click()
+
+            self.wait_for_element_by_css_selector(
+                '#horizontal-slider[data-dash-is-loading="true"]'
+            )
```
Marc-Andre-Rivet: Adding to the test -- we don't only check that the loading state is ok on initialization, also check that it's ok on user initiated actions

test/test_integration.py
```diff
@@ -332,19 +640,29 @@ def test_loading_range_slider(self):
...
+            '#horizontal-range-slider:not([data-dash-is-loading="true"])'
         )
-        lock.release()
+
+        with lock:
+            self.driver.find_element_by_id('test-btn').click()
+
+            self.wait_for_element_by_css_selector(
+                '#horizontal-range-slider[data-dash-is-loading="true"]'
+            )
```
Marc-Andre-Rivet: Init and user action cases, as above

Marc-Andre-Rivet:
Local testing: confirming that `2 graphs with different figures` test showing ```Loading...``` is a timing issue.

PR #498, opened 2019-03-25
:bug: fix loading_state props for input

byronz:
the `loading_state` should not be specified as props, as it already specified by data-* at top.


Marc-Andre-Rivet:
py37 failed on `test_graphs_with_different_figures` this is as in https://github.com/plotly/dash-core/issues/54 -- just rerun the tests...

PR #500, opened 2019-03-26
:bug: add implicitly_wait

byronz:
the added step  below was flakey because it has no implicitly_wait at web driver level, nor using the `wait_for` mechanism. this fix should solve this
`autoscale = self.driver.find_element_by_css_selector('#example-graph .ewdrag')`

we will run couple times to make sure it fixed
Fixes https://github.com/plotly/dash-core/issues/54

### DIFF
--- a/test/IntegrationTests.py
+++ b/test/IntegrationTests.py
@@ -105,6 +105,7 @@ class IntegrationTests(unittest.TestCase):
         time.sleep(2)
 
         # Visit the dash page
+        self.driver.implicitly_wait(2)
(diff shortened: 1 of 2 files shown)


byronz:
@Marc-Andre-Rivet  I think that test is stable now, also moved the snapshot timing so it can leverage the extra `wait_for`. the fix is ready for merge

Marc-Andre-Rivet:
@byronz cool, if you add `Fixes https://github.com/plotly/dash-core/issues/54` into the description, the issue will get closed when this PR gets merged.

PR #501, opened 2019-03-28
:hocho: remove vestigial fn from earlier :package: version

rpkyle:
#482 renamed core* to dcc*, coreLoading.R is not even exported, so it can be safely deleted.


nicolaskruchten:
💃

PR #502, opened 2019-03-28
 Bump plotly.js to version 1.45.3

wbrgss:
- Update plotly.js to latest patch versions
- plotly/plotly.js#3605 is important for DDK


wbrgss:
Just realized the ETA for `1.46.0` is ~April 1st...we may want to wait until this minor version to update here

alexcjohnson:
Yeah, let's wait for 1.46. I'd suggest making a patch release, but there are a couple of features already merged to master here, so we should wait a little longer and build this into the next dash-* minor release.

Issue #505, opened 2019-04-01
Input component property "maxlength" does not work


matthewchan15:
When using the `maxlength` property with the `input` component, nothing happens. The feature does not work. I've tested `maxlength` with both `string` and an `int` with no avail. I also tried setting `type=text` and that didn't work either.

This was tested with `dash-core-components==0.45.0`.

matthewchan15:
Test code:
```
import dash
from dash.dependencies import Input, Output, State
import dash_core_components as dcc
import dash_html_components as html


app = dash.Dash(__name__)

app.layout = html.Div(
    [
        dcc.Input(
            id="test",
            type='text',
            maxlength='10'
        )   
    ],
)


if __name__ == '__main__':
    app.server.run(debug=True)
```

celinehuang:
I have replicated the code in the latest version of Dash (`dash 1.1.1`) and this is no longer an issue @matthewchan15 could you close this issue since I don't have permission ?

Marc-Andre-Rivet:
Closing as fixed as per #523 above.

Issue #457, opened 2019-02-06
Set autocomplete to "off" on dcc.Input doesn't work


T4rk1n:
Should be `autoComplete` instead.

https://github.com/plotly/dash-core-components/blob/62e90491c67a79edcfb0bdbdf2f5fa134efd791e/src/components/Input.react.js#L161

https://community.plot.ly/t/how-to-disable-suggestions-on-input-component/18711

Lander37:
Not sure if this is the right place to ask, but will this be fixed in a future release? As it is still broken as of dcc v0.44. Or if there are any workarounds to disable autocomplete for now? Thanks!

thinkjrs:
@T4rk1n I've been using Dash for a while now and it's the right time for me to jump into dev here. 

I am happy to tackle this issue and should have this weekend to dedicate, pending I can get setup with the development environment--the readme setup/build instructions are crashing on my system (I commented on issue #459 seeking to resolve). 

Despite the setup issues, after looking through the linked code and generated .py modules, I'm assuming that other html attributes not camelCased need to be. Specifically:
- autocomplete
- autofocus
- inputmode
- maxlength
- minlength

What's strange is that the CHANGLOG.md indicates these were handled in release 0.37.0:

  [0.37.0] - 2018-11-04
  Fixed
  Some Input props weren't being picked up by React. Changed:
  autocomplete to autoComplete
  autofocus to autoFocus
  inputmode to inputMode
  maxlength to maxLength
  minlength to minLength

After digging into the release commits it looks like the unit test at `/test/unit/Input.test.js` is missing those above, though they were added to the test dataset.  

Going on the limited amount of digging I've done (and normally I'd just implement this and test first), am I safe to assume that the following will solve the issue (assuming all tests pass)?

1. Change the above attribute names to camelCased
2. Add unit tests for the above on the js side -> I'm assuming pytest/R/everything else is generated?

Thanks for the hand-holding through my first contrib :-)

tushar-semwal:
On setting autocomplete to 'off' for dcc.Input is still NOT working. Is this solved?
I am using dash = 0.39.0

ned2:
You're using a rather old version of Dash. Try updating to the latest,
would be the first thing to do.

PR #509, opened 2019-04-03
WIP - Issue 506 add prop type check

byronz:
would like to get some feedbacks before moving on

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,5 @@ venv/
 *.egg-info
 *.log
 .idea/
+.vscode/
 .DS_Store
--- a/dash_core_components/package.json
+++ b/dash_core_components/package.json
@@ -57,6 +57,7 @@
     "babel-eslint": "^10.0.1",
     "babel-jest": "^24.5.0",
     "babel-loader": "^8.0.5",
+    "check-prop-types": "^1.1.2",
     "component-playground": "^3.0.0",
     "copyfiles": "^2.0.0",
     "css-loader": "^1.0.1",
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -12,4 +12,5 @@ venv/
 *.egg-info
 *.log
 .idea/
+.vscode/
```
Marc-Andre-Rivet: Actually, depending on what is in that folder, sharing these `workspace` configurations for vscode could be a good idea. Already do that for the table https://github.com/plotly/dash-table/blob/master/.vscode/settings.json

dash_core_components/package.json
```diff
@@ -57,6 +57,7 @@
     "babel-eslint": "^10.0.1",
     "babel-jest": "^24.5.0",
     "babel-loader": "^8.0.5",
+    "check-prop-types": "^1.1.2",
```
valentijnnieman: What's the reason for using `check-prop-types` instead of `prop-types`?

byronz: the default prop-types check swallows the error, instead print a warn message in console.
the check-* one is basically a wrapper to throw the error back

byronz:
@valentijnnieman there are 3-4 break changes of the previous prop types you defined, would like to know if the change make sense to you. and yes they are breaking your unit tests too :)

src/components/Input.react.js
```diff
@@ -12,6 +13,7 @@ import {omit, isEmpty} from 'ramda';
 export default class Input extends Component {
     constructor(props) {
         super(props);
+        assertPropTypes(Input.propTypes, props, 'component prop', 'Input');
```
alexcjohnson: I wonder if, rather than putting this assertion in the constructor of each component, we could do it for every component we create, and only when devtools are enabled

https://github.com/plotly/dash-renderer/blob/b9254ede45f232590fe3e53b2bf0e86c529759ed/src/TreeContainer.js#L66-L75

alexcjohnson:
related: #457, #505

valentijnnieman:
> @valentijnnieman there are 3-4 break changes of the previous prop types you defined, would like to know if the changes make sense to you. and yes they are breaking your unit tests too :)

Yeah makes sense. I think most of what I did was rename those props, as they didn't work at all before. If they still don't work then I guess the breaking change is that they now work correctly instead of not at all, haha. Also do the unit tests "break" because of different prop names? If that's the case then simply renaming them in the tests should be fine too right?

src/components/Input.react.js
byronz: yep, this is still in my mind, but I found the attributes and props world is much more complicated than imagined.

PR #511, opened 2019-04-03
Issue 510 fix clearing date callback

derekclee:
Fixes a problem with `<DatePickerSingle />` where the callback `setProps` was not called when a date was cleared.


src/components/DatePickerSingle.react.js
```diff
@@ -67,8 +67,8 @@ export default class DatePickerSingle extends Component {
 
     onDateChange(date) {
         const {setProps} = this.props;
-        if (setProps && date !== null) {
-            setProps({date: date.format('YYYY-MM-DD')});
+        if (setProps) {
+            setProps({date: date ? date.format('YYYY-MM-DD') : null });
```
alexcjohnson: This looks good to me. I'm a bit concerned about backward compatibility though, as it's been this way for as long as these components have existed and users may depend on `null` never arriving in their callbacks. @chriddyp thoughts? Do we need to make this behavior opt-in?

alexcjohnson:
Thanks @derekclee! Really appreciate seeing the PR come with tests already in place 🏆 Aside from the question about backward compatibility that I'm not immediately sure how I feel about (if we were starting from scratch I'd much prefer to keep it simple like you have here), the only issue I see is linting - we use `prettier`, you can run it with `npm run format` to automagically fix the issues.

derekclee:
@alexcjohnson I ran `prettier` on the files and fixed up the commit. 

Our use case of the clearing is that we initially have a blank date filter on a visualization. The date may not always be present in the data thus having the filter optional is important. A user is able to filter by a date. If the date was not clearable, the user would have to do a full refresh of the application to get it back to the original state.  I can add an opt in prop if needed.   

On an unrelated note, I originally tried to use the `DatePickerRange` but that would no accept null starting dates. The fix for that is a lot harder as the error seems to be in `<DayPicker />` from `https://github.com/airbnb/react-dates` and this library is using an older version.

src/components/DatePickerSingle.react.js
alexcjohnson: OK, we discussed this internally, the consensus is we're happy to call the old behavior a bug, this PR a fix, so no need for opt-in 🎉

test/test_integration.py
```diff
@@ -1076,6 +1076,8 @@ def test_tabs_with_children_undefined(self):
 
         self.startServer(app=app)
 
+        self.wait_for_element_by_css_selector('#tabs-content')
```
alexcjohnson: There was a percy diff - the image produced in this PR prior to this commit was actually the right one, dunno how the incorrect one got approved... but to make that more robust I added this wait for the correct element to be on the page.

derekclee:
Thanks @alexcjohnson !

Issue #514, opened 2019-04-05
Graph config defaults don't match plotly.js


alexcjohnson:
`Graph` sets its own `config` defaults [here](https://github.com/plotly/dash-core-components/blob/b1932f9aaaea89f099c17be0ce8f1d9e0538379a/src/components/Graph.react.js#L586-L622) - but this has no mechanism to keep up to date with plotly.js, and in fact it's out of date already in at least one respect, `scrollZoom`, as of https://github.com/plotly/plotly.js/pull/3422

We should either remove those defaults entirely, or programmatically match them to plotly.js.

There's a related problem: if you override a *piece* of the config, the rest of the default gets dropped. So for example right now if you were to set something unrelated like `showLink: true`, the `scrollZoom: false` would be dropped - and set to the new plotly.js default!

cc @nicolaskruchten @michaelbabyn - discovered while researching https://github.com/plotly/plotly.js/issues/3738

nicolaskruchten:
Cc @wbrgss

alexcjohnson:
Did a little investigation of the current `Graph.defaultProps` values vs plotly.js:
- Only one value is provided but does NOT match plotly.js, the aforementioned `scrollZoom`.
- A whole bunch of entries aren't present at all:
  - plotlyServerURL
  - responsive
  - showSendToCloud
  - toImageButtonOptions
  - watermark
  - setBackground
  - logging
  - globalTransforms
  - locale
  - locales

For the moment I'll make a PR to remove the defaults and add documentation for the new entries. Later - unless it's quick and I can get it in for the upcoming release - I'll make another PR to test that these keys continue to match when we upgrade plotly.js.

alexcjohnson:
Actually I'm not going to include `setBackground`, `logging`, or `globalTransforms` - these either won't work or aren't serializable, if you wanted to use them you'd need to subclass `Graph`.

alexcjohnson:
Closing - #515 fixes the bug by removing the nested defaults, and includes a test that `propTypes` for `Graph` has the right keys from the plotly.js schema.

PR #515, opened 2019-04-09
Remove graph config innards from defaultProps

alexcjohnson:
Addresses *most* of #514 - doesn't include a test (yet?), but removes the config innards and adds the missing pieces to `propTypes`.


test/test_integration.py
```diff
@@ -1193,7 +1193,7 @@ def on_click(n_clicks):
     def check_plotlyjs(self):
         # find plotly.js files in the dist folder, check that there's only one
         all_dist = os.listdir(dcc.__path__[0])
-        js_re = '^plotly-(.*)\.min\.js$'
+        js_re = r'^plotly-(.*)\.min\.js$'
```
alexcjohnson: oddly I can't get this error to appear locally... 😐

test/test_integration.py
```diff
@@ -1184,6 +1185,54 @@ def on_click(n_clicks):
...
+        dist_version = re.match(js_re, plotlyjs_dist[0]).groups()[0]
+        self.assertEqual(page_version, dist_version)
+
+    def check_graph_config_shape(self):
+        config_schema = self.driver.execute_script(
+            'return Plotly.PlotSchema.get().config;'
+        )
+        with open(os.path.join(dcc.__path__[0], 'metadata.json')) as meta:
+            graph_meta = json.load(meta)['src/components/Graph.react.js']
+            config_prop_shape = graph_meta['props']['config']['type']['value']
```
alexcjohnson: kind of ugly to have to use `metadata.json` for this... I was hoping I could get this directly from the `Graph` component, but `PropTypes` seems to turn everything into a validation function, I don't see a way to get the input params back out.

test/test_integration.py
```diff
@@ -1184,6 +1185,54 @@ def on_click(n_clicks):
...
+
+        def crawl(schema, props):
+            for prop_name in props:
+                self.assertIn(prop_name, schema)
+
+            for item_name, item in schema.items():
+                if item_name in ignored_config:
+                    continue
+
+                self.assertIn(item_name, props)
```
alexcjohnson: Currently not testing anything about the props, just that the right keys exist. We could enhance this later to verify that their types match what's in the schema. Note that we could have put back in the deep defaults and used a similar framework to test that they match the schema, but I don't think that's a good idea. We'd still have the problem of a single override dropping the defaults - in principle picking up the right defaults again from plotly.js, but there could be subtle problems with this, or perhaps a user overrides the plotly.js version... I'd rather keep the looser coupling.

test/test_integration.py
```diff
@@ -1184,6 +1185,54 @@ def on_click(n_clicks):
...
+        self.check_graph_config_shape()
+        self.check_plotlyjs()
+
+    def check_plotlyjs(self):
+        # find plotly.js files in the dist folder, check that there's only one
+        all_dist = os.listdir(dcc.__path__[0])
+        js_re = r'^plotly-(.*)\.min\.js$'
+        plotlyjs_dist = [fn for fn in all_dist if re.match(js_re, fn)]
+
+        self.assertEqual(len(plotlyjs_dist), 1)
```
alexcjohnson: I don't know how we ended up with 2 copies of plotly.js on master, 1.44.1 and 1.45.0... the associated PRs all seem to swap one file for another, maybe a mis-merge somewhere. But anyway this test will prevent that in the future.

src/components/Graph.react.js
```diff
@@ -516,11 +533,43 @@ const graphPropTypes = {
...
+            scale: PropTypes.number,
+        }),
+
         /**
          * add the plotly logo on the end of the mode bar
          */
         displaylogo: PropTypes.bool,
 
+        /**
+         * add the plotly logo even with no modebar
```
Marc-Andre-Rivet:
```suggestion
         * Add the plotly logo even with no modebar
```

src/components/Graph.react.js
```diff
@@ -491,6 +501,13 @@ const graphPropTypes = {
          */
         displayModeBar: PropTypes.oneOf([true, false, 'hover']),
 
+        /**
+         * should we include a modebar button to send this data to a
```
Marc-Andre-Rivet:
```suggestion
         * Should we include a modebar button to send this data to a
```

src/components/Graph.react.js
```diff
@@ -374,6 +374,11 @@ const graphPropTypes = {
          */
         staticPlot: PropTypes.bool,
 
+        /**
+         * base URL for a Plotly cloud instance, if `showSendToCloud` is enabled
```
Marc-Andre-Rivet:
```suggestion
         * Base URL for a Plotly cloud instance, if `showSendToCloud` is enabled
```

src/components/Graph.react.js
alexcjohnson: Ah right, a lot of the older docstrings didn't bother capitalizing so I tried to follow that, but then some of the newer ones do... I should just go through and capitalize them all. Thanks for pointing this out.

PR #519, opened 2019-04-11
Issue 518 - Fix react dates regression in dash==0.41.0

### LINKED ISSUE #518: DatePickerSingle and DatePickerRange style regression in dash==0.41.0
Expecting picker to look like this:
image

Looks like this instead:
image

This is caused by the `.DateInput` not having `overflow: hidden;`

Marc-Andre-Rivet:
Fixes #518

### DIFF
--- a/src/components/css/react-dates@20.1.0-fix.css
+++ b/src/components/css/react-dates@20.1.0-fix.css
@@ -1,0 +1,3 @@
+.DateInput {
+    overflow: hidden;
+}
--- a/src/components/css/react-dates@20.1.0-fix.css
+++ b/src/components/css/react-dates@20.1.0-fix.css
@@ -1,3 +1,3 @@
-.DateInput {
-    overflow: hidden;
+.DateInput_input {
+    box-sizing: border-box;
 }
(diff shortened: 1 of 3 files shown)


src/components/css/react-dates@20.1.0-fix.css
```diff
@@ -0,0 +1,3 @@
+.DateInput {
+    overflow: hidden;
+}
```
Marc-Andre-Rivet: The react-dates component looks bad if the element is not very wide -- our previous usage and defaults tend to display incorrectly, proper overflow fixes this on our side for now.

alexcjohnson:
One weird effect shows up in percy - the datepicker arrow disappeared?
diff:
Screen Shot 2019-04-11 at 2 04 07 PM
new image:
Screen Shot 2019-04-11 at 2 04 16 PM
all the other visual diffs look excellent AFAICT

Marc-Andre-Rivet:
@alexcjohnson Thanks for catching that. The solution is the problem. Looking at this again.

Marc-Andre-Rivet:
Using `box-sizing` allows the `^` to be visible and the main elements to be bounded correctly / the border to appear.

PR #521, opened 2019-04-11
:wrench: change the test run order, so lint first

byronz:


byronz:
> Just so linting errors are displayed first in test failures?
> As long as we keep `-c` so everything else still runs even with a linting error, I don't care what order they run in

then shall we drop the `-c` ?  
that's no correct answer for this, I kinda know what's the pros and cons for both ways. 
but it would be good to notify earlier if there is anything than wait until the full run.

alexcjohnson:
Personally I’m strongly in favor of `-c`, I’d rather wait for all possible errors and fix them all at once than get the errors quicker but have to iterate.

Issue #467, opened 2019-02-23
Input timestamp return a string


romainmartinez:
Hello,

A minor problem appeared when using an `Input` component. The `n_submit_timestamp` property returns a string (_e.g. `'2019-02-23T16:39:12.076Z'`_) while other components such as Button returns an int (_e.g._ `1550939942322`).

I often use the following command to find out which button was last pressed and I can't use it with `Input`:

```Python
np.nanargmax(np.array([btn_1_n_clicks_timestamp, btn_2_n_n_clicks_timestamp, btn_3_n_n_clicks_timestamp], dtype=np.float)
```

Maybe you should standardize the two outputs.

alexcjohnson:
We have a better solution to this - which has already been merged and will be in the next release: callback context, that directly tells you which input(s) triggered the call [plotly/dash#608](https://github.com/plotly/dash/pull/608)

Having mismatched behavior for these is definitely a problem, but the solution - if callback context takes care of every case people were using `*_timestamp` for - may be to simply deprecate, and eventually remove, all the `*_timestamp` props.

alexcjohnson:
Dash 0.38 is released including `dash.callback_context`. I'll leave this issue open as a proxy for deciding whether to remove these props completely or fix them, once we get a bit more feedback on `callback_context`.

alexcjohnson:
The current proposal is to deprecate - and eventually remove - all `_timestamp` properties except for `modified_timestamp` in the `Store` component, which has the additional functionality of testing deep equality so we can avoid a repeat callback if the same data is stored twice.

romainmartinez:
Sounds great. In my use of dash, I only use `_timestamp` to know which is the last element used among an input list so I'm OK with to deprecate.

PR #523, opened 2019-04-12
Fix prop types

chriddyp:
in preparation for https://github.com/plotly/dash-renderer/pull/100


chriddyp:
a QA example: https://gist.github.com/chriddyp/93d5c12b88d7333f56457b867e94d7b6

src/components/Input.react.js
```diff
@@ -162,17 +165,26 @@ Input.propTypes = {
     /**
      * This attribute indicates whether the value of the control can be automatically completed by the browser.
      */
-    autocomplete: PropTypes.string,
+    autocomplete: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.bool
```
byronz: also the actual accepted type per spec is string, if we wanna accept boolean on Python/R side, they need to be coerced/casted to `on` and `off`

byronz:
@chriddyp  To keep DRY :camel: we might need a similar mechanism to share some of the global attributes definition like in HTML components.

src/components/RangeSlider.react.js
```diff
@@ -68,28 +68,13 @@ RangeSlider.propTypes = {
...
-            /**
-             * The style and label of the mark
-             */
-            PropTypes.shape({
-                style: PropTypes.object,
-                label: PropTypes.string,
-            }),
-        ]),
-    }),
+    marks: PropTypes.object,
```
alexcjohnson: Perhaps `objectOf` ala Table's `column_static_tooltip `? https://github.com/plotly/dash-table/blob/7a308b9ec62b1254db3bb1edcf931c1610c51af2/src/dash-table/dash/DataTable.js#L641

chriddyp: ah perfect, thanks!

src/components/Input.react.js
chriddyp: Ah ok. But these properties https://github.com/iandevlin/html-attributes/blob/master/boolean-attributes.json should accept True/False right?

chriddyp:
fixes #457
fixes 505

chriddyp:
> To keep DRY 🐫 we might need a similar mechanism to share some of the global attributes definition like in HTML components.

Yeah I agree. I'd like to punt this one to the future. For now, just want to make sure that we won't be breaking any valid code.

chriddyp:
Alright, I've addressed your comments @alexcjohnson @byronz

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Fixed
 - Fix style regression in DatePickerSingle and DatePickerRange [#518](https://github.com/plotly/dash-core-components/issues/518)
+- In `dcc.Input`, fixed several HTML properties that weren't properly camel cased and therefore were not actually applied in the DOM [#523](https://github.com/plotly/dash-core-components/pull/523/):
+    - `autocomplete` is now `autoComplete`
+    - `autofocus` is now `autoFocus`
+    - `inputmode` is now `inputMode`
+    - `maxlength` is now `maxLength`
+    - `minlength` is now `minLength`
+- Updated property definitions of various components in anticipation of upcoming component validation (https://github.com/plotly/dash-renderer/pull/100/). [#523](https://github.com/plotly/dash-core-components/pull/523/)
```
alexcjohnson: updated, or fixed? Perhaps "improved" since it's a little of each I guess.

src/components/Input.react.js
```diff
@@ -62,7 +62,7 @@ export default class Input extends Component {
                     if (setProps) {
                         const payload = {
                             n_blur: this.props.n_blur + 1,
-                            n_blur_timestamp: new Date(),
+                            n_blur_timestamp: new Date.now(),
```
alexcjohnson: This is a breaking change https://github.com/plotly/dash-core-components/issues/467 - yes, we should do it, but we should call it out.

src/components/Input.react.js
```diff
@@ -141,7 +141,10 @@ Input.propTypes = {
      * If true, changes to input will be sent back to the Dash server only on enter or when losing focus.
      * If it's false, it will sent the value back on every change.
      */
-    debounce: PropTypes.bool,
+    debounce: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.bool
+    ]),
```
alexcjohnson: What's string for here? Describe it in the docstring?

alexcjohnson: ^^ applies to a bunch of these where either string or bool now accepts either... it's not clear to me which strings do what, or if they all evaluate to booleans or do something else entirely.

chriddyp: this one was wrong actually - only `bool`. But for the other HTML boolean properties, HTML5 defines the only accepted values as the name of the attribute (case insensitive) or `""`: https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

That is, for `autoFocus`, the following assignments mean "turn on autofocus": `autoFocus=""`, `autoFocus="autoFocus"`, `autoFocus="autofocus"`, `autoFocus="AUTOFOCUS"`

And in order for it to be "turned off", it has to be omitted.

It seems like browsers are a little more loose and they evaluate these strings to be "If it's supplied, then it's applied" - it doesn't need to be the name of the attribute. This ends up being confusing because `checked="false"` is actually "turn on checked" because it's applied. MDN says:

> To be clear, the values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether. This restriction clears up some common misunderstandings: With checked="false" for example, the element’s checked attribute would be interpreted as true because the attribute is present.

So, we could either:
1. Just keep it as a string and allow users to navigate this spec. Move on for now and revist later.
2. Provide a preference but allow all strings: `autoFocus: oneOfType([oneOf(['autoFocus', 'autofocus', '']), propTypes.string])`
3. Be more opionated and only include the common forms like `'autoFocus': oneOf(['', 'autoFocus', 'autofocus', 'AUTOFOCUS'])` but omit the other permutations like `auToFoCus` that are technically valid.
3. Be even more opinionated and only include the most sane-looking version: `autoFocus='autoFocus'` and omit the other versions that are technically valid.

chriddyp: I think I prefer 2 or 3

alexcjohnson: So does `false` (not `'false'`) turn `autoFocus` off? Or is `None` -> `null` the only value that disables it?

If this check is only applied in dev mode, I'd vote to be more opinionated - (3) seems reasonable, plus `false` if that's allowed to disable. I'd omit `''` unless it's particularly common usage, as explicitly allowing a normally falsy value to mean true is bound to cause confusion. Either way we should document a recommended value, something like:
> autoFocus is enabled by `'autoFocus'`, but alternate capitalizations `'autofocus'` and `'AUTOFOCUS'` are also accepted.

chriddyp: I thought that `false` ended up getting rendered as `'false'` looks like I'm wrong about that:
image

(https://codepen.io/chriddyp/pen/MREmpK?editors=0010)

Similarly in our toolchain:
```python
import dash

app = dash.Dash(__name__)
app.layout = html.Div([
    dcc.Input(id='1', readOnly=True),
    dcc.Input(id='2', readOnly=False),
    dcc.Input(id='3', readOnly='readonly'),
    dcc.Input(id='4', readOnly='readOnly'),
    dcc.Input(id='5', readOnly='true'),
    dcc.Input(id='6', readOnly='false'),
    dcc.Input(id='7', readOnly=''),
    dcc.Input(id='8', readOnly='""'),
])

if __name__ == '__main__':
    app.run_server(debug=True)
```

image

`#7` is pretty confusing - to get `""` into the DOM, you have to pass in `'""'` not `''`


So I suppose we could allow `boolean` for these properties as well.

alexcjohnson: Oy, 7 is insane, definitely do not allow any version of that! So yeah, `true`, `false`, and a couple of capitalizations of `autoFocus`.

CHANGELOG.md
byronz: :+1: for improved, or we could have two categories in release log: 1 bug fix 2 feature enhancement

src/components/Input.react.js
byronz: a small section about the attributes in dash-docs might be necessary given all the experiments we have, as most of our audiences are not Frontend expert, we should align the preference in the demo code too

.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ jobs:
                 name: Install dependencies (dash)
                 command: |
                     git clone git@github.com:plotly/dash.git
-                    git clone git@github.com:plotly/dash-renderer.git
+                    git clone --branch chriddyp-dev-tools git@github.com:plotly/dash-renderer.git
```
chriddyp: I'll remove this before merging. Just making sure that the other repo's prop checking doesn't cause these tests to fail.

src/components/Input.react.js
```diff
@@ -162,19 +162,30 @@ Input.propTypes = {
     /**
      * This attribute indicates whether the value of the control can be automatically completed by the browser.
      */
-    autocomplete: PropTypes.string,
+    autoComplete: PropTypes.string,
```
byronz: yes, given the complexity of autoComplete, I think `string` is a good compromise for now

chriddyp:
Alright, this is passing with the `dev-tools` branch now 🎉 . 
- Had to make a fix to `ConfirmModal` so that it was idempotent, see https://github.com/plotly/dash-core-components/pull/523/commits/8b64f2d6c31723ec1ee5cbd53c8c5e65826dbf8a
- `.exact` didn't have a proper docstring. This docstring was pretty nice for the long list of properties in `Graph.config`, so I added support for `exact` in the component generation in https://github.com/plotly/dash/pull/692
- `.exact` also wasn't supported in the current version of `react-docgen` - upgrading to 3.0.0 fixed the issue.

chriddyp:
weird dateformat diff. `05/03/1997` vs `1997-05-03`
image

chriddyp:
that's the only error left

chriddyp:
Odd - the new screenshots match the date supplied exactly. We send `dt_input_1.send_keys("1997-05-03")` and it appears exactly as `"1997-05-03"` instead of getting reformatted as `05/03/1997`. 

Odd that this behaviour would change in this PR. Looking into it...

chriddyp:
OK - I've isolated the issue. There seems to be a problem with `dash-renderer`'s `setProps` not re-rendering the component with the correct prop types if an ID is not supplied. I'll look into dash-renderer next.

chriddyp:
Here's a reproducable example that shows the issue (try editing the dates). I'll work on a fix in `dash-renderer` now.
```python
from datetime import datetime as dt
import dash

app = dash.Dash(__name__)

app.layout = html.Div([
    html.Div([
        html.Label('With ID'),
        dcc.DatePickerSingle(id='date-picker-with-id'),
    ]),

    html.Div([
        html.Label('Without ID'),
        dcc.DatePickerSingle(),
    ])

])


if __name__ == '__main__':
    app.run_server(debug=True, dev_tools_hot_reload=False)
```

PR #531, opened 2019-04-18
WIP - R dbg

byronz:
This PR just merge from on-going fix-props-type, see those errors still exist.


chriddyp:
Looking into the date issue right now in https://github.com/plotly/dash-core-components/pull/523#issuecomment-484353338

byronz:
> I'm seeing Percy diffs related to `DatePickerRange`:
> 
> 
> 
> Specifically, the date format appears to me to be ISO-8601, whereas before it was `mm/dd/YYYY`. Is this intentional? I wasn't able to discern this from reviewing #519 or #518.

chris is debugging this now, as this also failed on fix-props-check branch. but the good thing is we know why your tests failed :)

PR #536, opened 2019-04-21
Clean up components with new setProps logic

chriddyp:
This PR:
- Makes several components stateless now that `setProps` is always defined 
- Unifies how certain components maintain their own state until a certain user action triggers a “report” to dash-renderer (e.g. `Slider` with `updatemode=‘mouseup’` or `Input` with `debounce`)
- Cleans up `dcc.Interval` & fixes what I'm considering to be a bug (but we can discuss) - https://github.com/plotly/dash-core-components/pull/536/commits/50275ea3110efe9f78e20fef29e1bff0b7e2d103
- Removes all `import * from ramda`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,10 +30,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
     ```
     0 seconds: interval started with `interval=5000`
     5 seconds: `n_intervals=1`
-    7 seconds: callback sets `disabled=True`
-    10 seconds: interval continues to run, but doesn't fire an update
+    7 seconds: callback sets `disabled=True` - interval stops
     13 seconds: callback sets `disabled=False` - clock resets
-    17 seconds: interval fires an update: `n_intervals=2`
+    18 seconds: interval fires an update: `n_intervals=2`
     ```
 
 ## [0.46.0] - 2019-04-10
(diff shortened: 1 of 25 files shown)


chriddyp:
For some more context 50275ea - here's an example where the existing behavior "feels glitchy" 
https://user-images.githubusercontent.com/1280389/56393731-4a26ee00-6203-11e9-966b-d467bc099ea4.gif

CHANGELOG.md
```diff
@@ -15,6 +15,26 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+    ```
+
+    Now, it will look like this:
+    ```
+    0 seconds: interval started with `interval=5000`
+    5 seconds: `n_intervals=1`
+    7 seconds: callback sets `disabled=True`
+    10 seconds: interval continues to run, but doesn't fire an update
+    13 seconds: callback sets `disabled=False` - clock resets
+    17 seconds: interval fires an update: `n_intervals=2`
```
alexcjohnson: Great catch! But I think you mean 18 seconds here?

alexcjohnson:
Looks great, thanks for doing this @chriddyp!

Re: the percy diffs: not sure what's up with the "Double click on legend" opacity change but that doesn't bother me. But we do have the date format issues still...

chriddyp:
Having a hard time figuring this image test out. The fix in 15a09b3 fixes this locally for me, and I created a new fresh environment entirely from scratch using the commands in `.circle/config.yml`. 

Locally, the state seems to be managed OK:

image

chriddyp:
@alexcjohnson - found the issue. Now that `debug=True` (so that our tests include component validation), we need the dev builds. `npm run build` in dash-renderer only builds the prod bundles, not the dev bundles 🙃 So, I added `&& npm run build:dev` in 05bc40d  and now everything looks 👌

alexcjohnson:
ping https://github.com/plotly/dash-core/issues/49

PR #537, opened 2019-04-22
New set props build

chriddyp:
Same as https://github.com/plotly/dash-core-components/pull/536, just testing if the builds in #536 are cached in some way that I don't understand.


chriddyp:
@alexcjohnson - found the issue. Now that `debug=True`, we need the dev builds. `npm run build` in `dash-renderer` only builds the prod bundles, not the dev bundles 🙃  So, I added `&& npm run build:dev` in 82fe3bc and now everything looks 👌

PR #540, opened 2019-04-29
Add frames to Graph

nicolaskruchten:
`frames` are part of the Plotly.js figure spec, so `dcc.Graph()` should accept them.

### DIFF
--- a/src/components/Graph.react.js
+++ b/src/components/Graph.react.js
@@ -104,15 +104,18 @@ class PlotlyGraph extends Component {
         ) {
             return Plotly.animate(id, figure, animation_options);
         }
-        return Plotly.react(id, figure.data, clone(figure.layout), config).then(
-            () => {
-                if (!this._hasPlotted) {
-                    this.bindEvents();
-                    Plotly.Plots.resize(document.getElementById(id));
-                    this._hasPlotted = true;
-                }
+        return Plotly.react(id, {
+            data: figure.data,
+            layout: clone(figure.layout),
+            frames: figure.frames,
+            config: config,
+        }).then(() => {
+            if (!this._hasPlotted) {
+                this.bindEvents();
+                Plotly.Plots.resize(document.getElementById(id));
+                this._hasPlotted = true;
             }
-        );
+        });
     }
 
     extend(props) {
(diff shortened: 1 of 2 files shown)


src/components/Graph.react.js
```diff
@@ -622,7 +622,7 @@ const graphDefaultProps = {
     relayoutData: null,
     extendData: null,
     restyleData: null,
-    figure: {data: [], layout: {}},
```
alexcjohnson: FWIW it seems to me nested defaults like this are a bit dangerous, as providing some other part of the `figure` prop will cause the rest of the defaults to be dropped - first mentioned in https://github.com/plotly/dash-core-components/issues/514. It's OK here because these defaults behave the same as the attributes being missing entirely, but by the same token is there actually a benefit to including any of them? Could we just say `figure: {}`?

nicolaskruchten: I don't mind either way, I was just following the current pattern :)

Can we get this one merged in and follow up in a different issue/PR?

nicolaskruchten:
I'm not sure what Percy is up to here... @alexcjohnson what's your recommendation? rerun? approve?

alexcjohnson:
We've seen that percy diff before, have to track it down at some point but it's OK for now.

Does this PR work though? Doesn't look like frames are propagated to `Plotly.react`

https://github.com/plotly/dash-core-components/blob/bab0d5281c9e387a8642160ffde0680aa6a72c7d/src/components/Graph.react.js#L101

in fact we have to use the other signature `Plotly.react(gd, {data, layout, frames, config})` in order to use frames, right?

nicolaskruchten:
Good point, I hadn't actually dug in that far :)

nicolaskruchten:
Do you think I should clone `frames` or not?

alexcjohnson:
Shouldn’t need cloning, no.

nicolaskruchten:
OK so this one actually works now @alexcjohnson  :)

nicolaskruchten:
I've approved that same Percy diff, so this one should be good to review.

nicolaskruchten:
Do you want a changelog entry here also?

nicolaskruchten:
OK I added one.

src/components/Graph.react.js
```diff
@@ -322,12 +325,12 @@ const graphPropTypes = {
      * Plotly `figure` object. See schema:
      * https://plot.ly/javascript/reference
      * Only supports `data` array and `layout` object.
-     * `config` is set separately by the `config` property,
-     * and `frames` is not supported.
```
alexcjohnson: You took out the note that `frames` is not supported... but it still says 
> Only supports `data` array and `layout` object.

nicolaskruchten: fixed.

nicolaskruchten:
🤝 deal.

PR #542, opened 2019-04-29
Deleting unused proptype ConfirmDialog.key

nicolaskruchten:
This doesn't seem to be read or written anywhere


nicolaskruchten:
So there's a test for this but it's still not clear what if anything it does.

nicolaskruchten:
Would we be worried about some apps built that use this? Should I just add a "Deprecated, doesn't do anything" description instead?

PR #543, opened 2019-04-29
add in all missing descriptions

nicolaskruchten:
If we merge in https://github.com/plotly/dash-core-components/pull/542 then this PR resolves all "missing description" warnings. If we don't merge that one we need to write a description for this mysterious `key` :)

See also https://github.com/plotly/dash-core/issues/52

### DIFF
--- a/src/components/DatePickerRange.react.js
+++ b/src/components/DatePickerRange.react.js
@@ -178,6 +178,11 @@ export default class DatePickerRange extends Component {
 }
 
 DatePickerRange.propTypes = {
+    /**
+     * The ID of this component, used to identify dash components
+     * in callbacks. The ID needs to be unique across all of the
+     * components in an app.
+     */
     id: PropTypes.string,
 
     /**
@@ -187,7 +192,16 @@ DatePickerRange.propTypes = {
      */
     start_date: PropTypes.string,
 
+    /**
+     * The HTML element ID of the start date input field.
+     * Not used by Dash, only by CSS.
+     */
     start_date_id: PropTypes.string,
+
+    /**
+     * The HTML element ID of the end date input field.
+     * Not used by Dash, only by CSS.
+     */
     end_date_id: PropTypes.string,
 
     /**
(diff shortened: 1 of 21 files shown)


nicolaskruchten:
Note that I wrote very few of these: they're mostly copy-pasta from other places where similar props were used :)

src/components/DatePickerRange.react.js
```diff
@@ -187,7 +192,14 @@ DatePickerRange.propTypes = {
...
     start_date: PropTypes.string,
 
+    /**
+     * The ID of the start date input field
+     */
     start_date_id: PropTypes.string,
+
+    /**
+     * The ID of the end date input field
+     */
```
alexcjohnson: I might elaborate a bit, something like:

> The HTML element ID of the end date input field. Not used by dash, only by CSS.

src/components/Graph.react.js
```diff
@@ -75,6 +75,12 @@ function generateId() {
     );
 }
 
+/**
+ * Graph cand be used to render any plotly.js-powered data visualization.
```
alexcjohnson: can~d~

src/components/Input.react.js
```diff
@@ -178,6 +178,9 @@ Input.propTypes = {
         PropTypes.bool,
     ]),
 
+    /**
+     * Provides a hint as to the type of data that might be entered by the user while editing the element or its contents.
```
alexcjohnson: so it's clear why:

> Provides a hint *to the browser* as to the type...

src/components/Link.react.js
```diff
@@ -27,6 +27,9 @@ function CustomEvent(event, params) {
 }
 CustomEvent.prototype = window.Event.prototype;
 
+/**
+ * Link allows you to create a clickable link within a multi-page app
+ */
```
alexcjohnson: Do we want to say "for links outside this app, use `html.A` instead"? Or is `Link` good in both cases?

nicolaskruchten:
@alexcjohnson addressed all your comments.

PR #550, opened 2019-05-09
Ronniebugia fixing checklist

ronniebugia:
There was an issue regarding the Checklist.react.js component. There was no default value specified for the ``values`` attribute. If you have a dash app with a ```dcc.Checklist()``` and did not specify a value for the ```values``` attribute, you would have gotten an error that says ```error loading layout``` .


Marc-Andre-Rivet:
@ronniebugia Thanks for finding this. We'll need a few more things to make this PR complete:

1. Backpedal the fix, add a test that fails without it (commit + CI build) followed by the fix and the test now passing (commit + CI build) -- this test should be in `test_integration.py` and be an isolated test (e.g. `test_vertical_range_slider`) with a snapshot and a browser log check at the end

2. Add a `CHANGELOG` unreleased `###Fixed` entry that refers to this PR

alexcjohnson:
Closing - the default was added while renaming `values` to `value` in #558

PR #555, opened 2019-05-13
Build & slew of minor improvements

Marc-Andre-Rivet:
- the `dev` build was useful before we used external source maps on the prod build -- prod and dev are now redundant as the source map is always available
- use `package-info.json` in the dash_core_components/ dist folder to allow normal npm publish to work (like in table)
- whitelisting npm publish content through `files` in package.json instead of blacklisting through .gitgnore / .npmignore (like in table)
- get version information from `package-info.json` and `package.json` (setup.py) instead of `version.py` (like in table)
- resolve `setup.py` content from `package.json` (like in table)
- clean up the build commands & always build r and python components
- removing old artefacts & dev artefacts

Tagging @rpkyle as this may impact the R generation in ways I don't understand :)


alexcjohnson:
Do we use the copied `package[-info].json` for anything besides the version info? If so, can we just stop copying it, and instead have the component generation step create a `version.py` matching the `package.json` version?

For now since we already have this pattern in table, fine to expand it. But perhaps with `dash-component-system` we simplify this to an auto-generated `version.py`.

Marc-Andre-Rivet:
Agreed. I kept it because my understanding was that R uses it. @rpkyle?

alexcjohnson:
>  I kept it because my understanding was that R uses it.

AFAICT R only uses the base `package.json`

https://github.com/plotly/dash/blob/ec1576c3652cbb90ffb064abe2753a17e13f3c51/dash/development/component_generator.py#L97-L102

Marc-Andre-Rivet:
@alexcjohnson Actual back on your comment
> Do we use the copied package[-info].json for anything besides the version info?

We do not, the only other usage is done with the root `package.json` [here](https://github.com/plotly/dash-core-components/pull/555/files#diff-2eeaed663bd0d25b7e608891384b7298R12) -- will update to generate `version.py` and update the table as well.

PR #558, opened 2019-05-30
Checklist: rename `values` prop to `value`

alexcjohnson:
This is one of the breaking changes planned for the Dash 1.0 release.

Also a couple of little prop type & doc fixes for the `value` prop in `Dropdown` and `RadioItems`.

### DIFF
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -838,7 +838,7 @@ class Tests(IntegrationTests):
                     {'label': 'San Francisco', 'value': 'SF'},
                     {'label': u'北京', 'value': u'北京'}
                 ],
-                values=['MTL', 'SF']
+                value=['MTL', 'SF']
             ),
 
             html.Label('Text Input'),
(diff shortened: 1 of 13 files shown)


test/test_integration.py
```diff
@@ -838,7 +838,7 @@ def test_gallery(self):
                     {'label': 'San Francisco', 'value': 'SF'},
                     {'label': u'北京', 'value': u'北京'}
```
byronz: we should leave an eastern egg here with {'label': u'北京', 'value': u'帝都'}

alexcjohnson: I liked your original comment - it would be an eastern egg indeed 😉 feel free to add it!

byronz: but you might not get the egg? or you level of mandarin is beyond my imagination :)

alexcjohnson: My mandarin is only as good as google translate's. But I have no problem with spreading the fun around to more languages - we have plenty of little English jokes scattered around our tests!

PR #563, opened 2019-06-07
Don't attempt to resize unmounted graphs

### LINKED ISSUE #534: dcc.Graph attempts to resize umounted Graphs
Results in a console error: `Uncaught Error: DOM element provided is null or undefined`

Demo of behaviour:

dom-element-provided

<details>
  <summary>
  Dash app used to demo this bug
  </summary>

- Using **Chromium Version 73.0.3683.86 (Official Build) Built on Ubuntu , running on Ubuntu 18.04 (64-bit)**
- Using **Dash 0.41.0**

``` python
import dash

app = dash.Dash(__name__)

    html.Div([
        # without the fix in this PR, this graph will disappear after clicking
        # graphs back and forth ~8 times
        dcc.Graph(
            id='example-graph-ex',
            figure={
                'data': [
                    {'x': [1, 2, 3], 'y': [4, 1, 2],
                        'type': 'scattergl', 'name': 'SF'},
                    {'x': [1, 2, 3], 'y': [2, 4, 5],
                     'type': 'scattergl', 'name': u'Montréal'},
                ]
            }
        )
    ]),
    dcc.Tabs(id="tabs", children=[
        dcc.Tab(label='Tab one', children=[
            html.Div([
                dcc.Graph(
                    id='example-graph',
                    figure={
                        'data': [
                            {'x': [1, 2, 3], 'y': [4, 1, 2],
                                'type': 'scattergl', 'name': 'SF'},
                            {'x': [1, 2, 3], 'y': [2, 4, 5],
                             'type': 'scattergl', 'name': u'Montréal'},
                        ]
                    }
                )
            ])
        ]),
        dcc.Tab(label='Tab two', children=[
                dcc.Graph(
                    id='example-graph-1',
                    figure={
                        'data': [
                            {'x': [1, 2, 3], 'y': [1, 4, 1],
                                'type': 'scattergl', 'name': 'SF'},
                            {'x': [1, 2, 3], 'y': [1, 2, 3],
                             'type': 'scattergl', 'name': u'Montréal'},
                        ]
                    }
                )
        ]),
        dcc.Tab(label='Tab three', children=[
                dcc.Graph(
                    id='example-graph-2',
                    figure={
                        'data': [
                            {'x': [1, 2, 3], 'y': [2, 4, 3],
                                'type': 'scattergl', 'name': 'SF'},
                            {'x': [1, 2, 3], 'y': [5, 4, 3],
                             'type': 'scattergl', 'name': u'Montréal'},
                        ]
                    }
                )
        ]),
    ])
])


if __name__ == '__main__':
    app.run_server(debug=True)
```
</details>

Simple fix: check if the element with the given `id` exists before [calling `Plotly.Plots.resize()`](https://github.com/plotly/dash-core-components/blob/master/src/components/Graph.react.js#L204)

wbrgss:
Simple check for a graph element before a resize attempt. Fixes #534.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Fixed
 - Fixed `min_date_allowed` and `max_date_allowed` bug in `DatePickerRange` [#551]https://github.com/plotly/dash-core-components/issues/551)
+- Fixed unwanted resize() calls on unmounted `Graph`s [#534](https://github.com/plotly/dash-core-components/issues/534)
 
 ### Changed
 - Changed `dcc.Checklist` prop `values` to `value`, to match all the other input components [#558](https://github.com/plotly/dash-core-components/pull/558). Also improved prop types for `Dropdown` and `RadioItems` `value` props to consistently accept both strings and numbers.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Fixed
 - Fixed `min_date_allowed` and `max_date_allowed` bug in `DatePickerRange` [#551]https://github.com/plotly/dash-core-components/issues/551)
-- Fixed unwanted resize() calls on unmounted `Graph`s [#534](https://github.com/plotly/dash-core-components/issues/534)
+- Fixed unwanted `resize()` calls on unmounted `Graph`s [#534](https://github.com/plotly/dash-core-components/issues/534)
 
 ### Changed
 - Changed `dcc.Checklist` prop `values` to `value`, to match all the other input components [#558](https://github.com/plotly/dash-core-components/pull/558). Also improved prop types for `Dropdown` and `RadioItems` `value` props to consistently accept both strings and numbers.
--- a/src/components/Graph.react.js
+++ b/src/components/Graph.react.js
@@ -91,6 +91,7 @@ class PlotlyGraph extends Component {
         super(props);
         this.bindEvents = this.bindEvents.bind(this);
         this._hasPlotted = false;
+        this.graphResize = this.graphResize.bind(this);
     }
 
     plot(props) {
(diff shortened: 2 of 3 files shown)


alexcjohnson:
We probably should remove the resize listener in `componentWillUnmount` as well. I'm imagining a case where you unmount and remount a plot many times, and we'll end up with a million resize listeners. `Plotly.Plots.resize` *should* be smart enough to not actually redraw the plot more than once, but there's still some overhead (and I wouldn't be surprised if there's a case where it *would* do many redraws).

Do we need this same conditional around the `resize` and `bindEvents` in `plot` too?

https://github.com/plotly/dash-core-components/blob/6d945c41b4c7e2c1d596769ffdbbb49bf6bdd123/src/components/Graph.react.js#L114-L115

Much less likely, but in principle the plot element could be removed before it finishes drawing and enters that `.then`

wbrgss:
> probably should remove the resize listener in componentWillUnmount as well.

In order to do this I moved the resize handler from its anon function into a named function bound in the constructor.

alexcjohnson:
Looks great, thanks! Given that this is a bugfix, would you mind locking it down with a test?

Also changelog

I'm 99% convinced the way you did the `removeEventListener` is right - was worried that we were referencing a class attr instead of an instance attr but I guess `this.graphResize = this.graphResize.bind(this);` ensures it's an instance attr and we're doing the right thing.

wbrgss:
I added an integration test.

Before the changes in this PR (failing test):
``` bash
======================================================================
ERROR: test_unmounted_graph_resize (test.test_integration.Tests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/plotly/Workspace/dash-core-components/test/test_integration.py", line 138, in test_unmounted_graph_resize
    raise Exception('browser error logged during test', entry)
Exception: ('browser error logged during test', {'level': 'SEVERE', 'message': 'http://localhost:8050/_dash-component-suites/dash_core_components/plotly-1.47.0.min.js?v=0.48.0&m=1555705071 6:2043297 Uncaught Error: DOM element provided is null or undefined', 'source': 'javascript', 'timestamp': 1560296987995})

----------------------------------------------------------------------
Ran 1 test in 5.934s
```
After:
```
.
----------------------------------------------------------------------
Ran 1 test in 5.954s

OK
```
(no exceptions raised)

PR #564, opened 2019-06-08
Slider tooltips

wbrgss:
This PR enables tooltips over `dcc.Slider` and `dcc.RangeSlider` handles, either on hover or always. Motivated by a customer request via @cldougl, but it's been [desired](https://stackoverflow.com/questions/54229804/how-to-add-a-label-over-slider-selector) in our open-source community as well. 

Adds a new property `tooltip` that passes in parameters from `tipProps`, an object native to the `rc-slider` `createSliderWithTooltip` API. `tipProps` isn't very well documented but AFAIK it essentially mirrors the [rc-tooltip API](https://github.com/react-component/tooltip#api). See example of `tipProps` usage in React [here](https://codesandbox.io/s/zn908z99yx).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
-- Ability to add tooltips to `Slider` and `RangeSlider`, which can be visible always or on hover. Tooltips also take a position argument. [#564]
+- Ability to add tooltips to `Slider` and `RangeSlider`, which can be visible always or on hover. Tooltips also take a position argument. [#564](https://github.com/plotly/dash-core-components/pull/564)
 
 ### Fixed
 - Fixed `min_date_allowed` and `max_date_allowed` bug in `DatePickerRange` [#551]https://github.com/plotly/dash-core-components/issues/551)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- Ability to add tooltips to `Slider` and `RangeSlider`, which can be visible always or on hover. Tooltips also take a position argument. [#564]
+
 ### Fixed
 - Fixed `min_date_allowed` and `max_date_allowed` bug in `DatePickerRange` [#551]https://github.com/plotly/dash-core-components/issues/551)
 - Fixed unwanted `resize()` calls on unmounted `Graph`s [#534](https://github.com/plotly/dash-core-components/issues/534)
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "moment": "^2.20.1",
     "prop-types": "^15.6.0",
     "ramda": "^0.24.1",
-    "rc-slider": "^8.3.1",
+    "rc-slider": "^8.6.11",
     "react-addons-shallow-compare": "^15.6.0",
     "react-dates": "^20.1.0",
     "react-docgen": "^3.0.0",
--- a/src/components/RangeSlider.react.js
+++ b/src/components/RangeSlider.react.js
@@ -159,7 +159,6 @@ RangeSlider.propTypes = {
     pushable: PropTypes.oneOfType([PropTypes.bool, PropTypes.number]),
 
     tooltip: PropTypes.exact({
-
        /**
         * Determines whether tooltips should always be visible
         * (as opposed to the default, visible on hover)
--- a/src/components/RangeSlider.react.js
+++ b/src/components/RangeSlider.react.js
@@ -12,8 +12,8 @@ export default class RangeSlider extends Component {
         super(props);
         this.propsToState = this.propsToState.bind(this);
         this.DashSlider = props.tooltip
-          ? createSliderWithTooltip(Range)
-          : Range;
+            ? createSliderWithTooltip(Range)
+            : Range;
     }
 
     propsToState(newProps) {
@@ -163,7 +163,7 @@ RangeSlider.propTypes = {
          * (as opposed to the default, visible on hover)
          */
         PropTypes.exact({
-            visible: PropTypes.oneOf(['visible', 'hover']).isRequired
+            visible: PropTypes.bool.isRequired,
         }),
 
         /**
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "moment": "^2.20.1",
     "prop-types": "^15.6.0",
     "ramda": "^0.24.1",
-    "rc-slider": "^8.3.1",
+    "rc-slider": "^8.6.11",
```
wbrgss: Necessary minor upgrade that adds the `tipProps` API (undocumented; last update to the changelog / [HISTORY.MD](https://github.com/schrodinger/rc-slider/blob/master/HISTORY.md) is from 2016).

src/components/RangeSlider.react.js
```diff
@@ -152,6 +157,49 @@ RangeSlider.propTypes = {
      */
     pushable: PropTypes.oneOfType([PropTypes.bool, PropTypes.number]),
 
+    tooltip: PropTypes.oneOfType([
```
wbrgss: This is a pattern that @Marc-Andre-Rivet suggested. Example usage [here](https://github.com/plotly/dash-table/blob/master/src/dash-table/dash/DataTable.js#L453) in `DataTable`. However, it's not working for me, allowing e.g.

``` python
    dcc.RangeSlider(
        id='my-range-slider',
        min=0,
        max=20,
        step=0.5,
        value=[5,10],
        tooltip={
            'foo': 'bar' # I expect this not to be allowed with the strict type-checking below
        }
    ),
```

package.json
alexcjohnson: Seems like its repo has moved to https://github.com/react-component/slider - `HISTORY.MD` is still pretty sparse there, but `tipProps` appears to have been added back in 8.1.something. Anyway, I'm happy to have this up to date!

src/components/RangeSlider.react.js
```diff
@@ -11,6 +11,9 @@ export default class RangeSlider extends Component {
     constructor(props) {
         super(props);
         this.propsToState = this.propsToState.bind(this);
+        this.DashSlider = props.tooltip
+            ? createSliderWithTooltip(Range)
+            : Range;
```
alexcjohnson: This will break if someone starts out with no tooltips and adds them later.

src/components/RangeSlider.react.js
alexcjohnson: Huh, not sure what's going on there, looks to me like that should work.

Note though that your last one, `PropTypes.exact({visible, position})`, actually encompasses both of the other two as well as `{}` since neither is required. Unless you specifically want to prevent `tooltip={}` from creating a tooltip (and to my mind that seems like a fine way for users to say "give me a default tooltip") I'd just use that last form by itself, no `oneOfType`.

src/components/Slider.react.js
```diff
@@ -133,6 +138,49 @@ Slider.propTypes = {
...
+                'bottom',
+                'topLeft',
+                'topRight',
+                'bottomLeft',
+                'bottomRight',
+            ]).isRequired,
+        }),
+
+        PropTypes.exact({
+            visible: PropTypes.oneOf(['visible', 'hover']),
```
alexcjohnson: elsewhere `visible` is a `bool` - which is right?

wbrgss: Good catch, it should be a bool in both `Slider` and `RangeSlider`.

src/components/RangeSlider.react.js
wbrgss: OK — I added this back to the render method; that should be fine right?

src/components/RangeSlider.react.js
wbrgss: > Unless you specifically want to prevent tooltip={} from creating a tooltip

IIRC @Marc-Andre-Rivet  _did_ want to prevent this, but I could have misunderstood him. Regardless, I'm inclined to agree `tooltip={}` should be allowed. `propTypes.exact` seems to actually allow other types as well (not just objects), which is confusing to me, but that means any truthy value will result in a default tooltip.

As an aside, I'd like to be more informed about stricter type-checking with `propTypes` and the `oneOfType` / `exact` pattern — do you understand why the above doesn't work @Marc-Andre-Rivet ?).

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- Ability to add tooltips to `Slider` and `RangeSlider`, which can be visible always or on hover. Tooltips also take a position argument. [#564]
```
alexcjohnson: This doesn't work quite like in conversations - you need to put the full link. (I notice there's a missing left paren below, would you mind fixing that too?)

src/components/Slider.react.js
```diff
@@ -133,6 +139,31 @@ Slider.propTypes = {
      */
     max: PropTypes.number,
 
+    tooltip: PropTypes.exact({
+       /**
+        * Determines whether tooltips should always be visible
+        * (as opposed to the default, visible on hover)
+        */
+        visible: PropTypes.bool,
```
Marc-Andre-Rivet: `visible` seems like a fuzzy name for hover vs. always visible

wbrgss: Yeah, I agree. I left it like that because that's the [API underneath](https://github.com/react-component/tooltip#props) (see "tooltip"). But considering our discussion on not needing a 1-1 API match for DCC, if you'd rather have an alias i.e. `visible` = `always_visible` I'll add it.

alexcjohnson: I like `always_visible: bool` actually - you'd only use it when it's `true` since `false` is the default. And even if you did say `always_visible: false`, it's pretty clear that that's not the same thing as "never visible", which is the implication of `visible: false`.

wbrgss: OK, I've made the change.

alexcjohnson: LGTM - @Marc-Andre-Rivet are you comfortable with this?

src/components/RangeSlider.react.js
wbrgss: On second thought, putting this in the `render()` method will break dragging behavior (unfortunately my screen recorder can't record a cursor; I'm dragging after hovering on the handle): 

drag

I'm not sure I actually understand the problem with this in the constructor. If I have a hot-reload app, and I add/remove the `tooltip` prop in `Range`/`RangeSlider`, Dash will handle it fine. The only thing that will be reset is any value that has been changed via user interaction (i.e. dragging/clicking the handle).

<details>
  <summary>This is a simple app I'm using to test</summary>

``` python
import dash

app = dash.Dash(__name__)
server = app.server

app.layout = html.Div([
    html.H1("Tooltips"),
    html.Br(),
    html.H1("Slider, always visible"),
    dcc.Slider(
        id='my-slider',
        min=0,
        max=20,
        step=0.5,
        value=10,
        tooltip={'always_visible': True}
    ),

    html.Br(),

    html.H1("RangeSlider, on hover"),
    dcc.RangeSlider(
        id='my-range-slider',
        min=0,
        max=20,
        step=0.5,
        value=[5,10],
        tooltip={'placement': 'bottom'}
    ),
])

if __name__ == '__main__':
  app.run_server(debug=True)
```
</details>

src/components/RangeSlider.react.js
```diff
@@ -152,6 +171,31 @@ RangeSlider.propTypes = {
...
+         */
+        always_visible: PropTypes.bool,
+
+        /**
+         * Determines the placement of tooltips
+         * See https://github.com/react-component/tooltip#api
+         * top/bottom{*} sets the _origin_ of the tooltip, so e.g. `topLeft` will
+         * in reality appear to be on the top right of the handle
+         */
+        placement: PropTypes.oneOf([
```
wbrgss: I made a mistake before setting the key as `position`, it should be `placement`

src/components/RangeSlider.react.js
alexcjohnson: The problem is if `tooltip` is the output of a callback

wbrgss: Ah right, of course. I've added the same conditional to `componentWillReceiveProps` and `tooltip` callabacks + normal drag behaviour appear to be working correctly now.

src/components/Slider.react.js
Marc-Andre-Rivet: Yup. All good.

PR #566, opened 2019-06-12
Update R package to 0.48.0 

rpkyle:
This PR proposes to update `dashCoreComponents` to 0.48.0. This package version enforces a dependency on `dash` rather than `dashR`.


rpkyle:
> 💃 as with [plotly/dash-html-components#117](https://github.com/plotly/dash-html-components/pull/117) just a non-blocking comment about the double-trailing-newlines - which I notice the `.Rd` files have too.

fixed in ffd6c34

PR #569, opened 2019-06-14
Markdown dedent, and remove containerProps

### LINKED ISSUE #567: For 1.0, dedent in Markdown
Per @chriddyp, the Markdown component should auto-dedent in JS.

See also #250

alexcjohnson:
Closes #567 

### Breaking change for Dash 1.0

Added a `dedent` prop to the `Markdown` component, set to `true` by default. Behavior is intended to match [`textwrap.dedent`](https://docs.python.org/2/library/textwrap.html#textwrap.dedent) - in particular:
- Empty lines are ignored
- Lines with only whitespace are emptied, then ignored
- All other lines: we find the longest leading whitespace string they all share, and remove it from them all.
- Spaces and tabs are both considered whitespace, but are not considered equal.

Also:
- Removed the `containerProps` prop - based on the refactor @wbrgss did in #562 it appears to be useless, as we have the necessary `id`, `className`, and `style` on the container already.
- Stopped passing on extra props to the underlying `react-markdown` component. All props are already used explicitly, this seems like it can only cause problems, it's probably a relic of some older usage.


src/components/Markdown.react.js
```diff
@@ -11,6 +11,11 @@ import './css/highlight.css';
  * [react-markdown](https://rexxars.github.io/react-markdown/) under the hood.
  */
 class DashMarkdown extends Component {
+    constructor(props) {
+        super(props);
+        this.highlightCode = this.highlightCode.bind(this);
```
alexcjohnson: apparently not needed given our current usage, but makes me nervous to have it there unbound...

test/unit/Markdown.test.js
```diff
@@ -0,0 +1,77 @@
...
+    test('same chars are removed from multiple lines, ignoring blanks', () => {
+        ['', '    ', '\t', '\t\t', '          ', '\t    \t'].forEach(pre => {
+            expect(
+                md.dedent(
+                    pre +
+                        'a\n' +
+                        pre +
+                        '  b\r' +
+                        pre +
+                        'c\r\n' +
```
alexcjohnson: Prettier certainly did not improve clarity here... and its indentation choice is peculiar for this case. Oh well...

test/unit/Markdown.test.js
```diff
@@ -0,0 +1,77 @@
...
+
+    test('mismatched chars are not removed', () => {
+        expect(md.dedent('    \ta\n\t    b')).toEqual('    \ta\n\t    b');
+    });
+
+    test('the dedent prop controls behavior', () => {
+        const text = '    a\n    b';
+        const mdDedented = render(<Markdown children={text} />);
+        expect(mdDedented.find('code').length).toEqual(0);
+        expect(mdDedented.find('p').length).toEqual(1);
```
alexcjohnson: dedented: makes everything into a `<p>` element.

test/unit/Markdown.test.js
```diff
@@ -0,0 +1,77 @@
...
+
+        const mdRaw = render(<Markdown children={text} dedent={false} />);
+        expect(mdRaw.find('code').length).toEqual(1);
+        expect(mdRaw.find('p').length).toEqual(0);
```
alexcjohnson: dedent turned off: the leading spaced mean it gets turned into a code block.

alexcjohnson:
@wbrgss Two things I missed when reviewing recent PRs:
- most of integration tests got commented out! That's why Percy is complaining it got no snapshots 🤦‍♂ fixed in https://github.com/plotly/dash-core-components/pull/569/commits/d207da36f9e6ec1d95f27d0b0511b089eb1b6c5c
- changelog for the syntax highlighter removal PR, along with this PR https://github.com/plotly/dash-core-components/pull/569/commits/df5192d1f4f9fd3557d4c17100eff6153f63b8c4

alexcjohnson:
Hmm, the percy diffs, now that they're back, may point to a problem, inline `code` is not actually inline...

test/unit/Markdown.test.js
byronz: cannot partially disable just for this chunk?

CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+- `Markdown` components support code highlighting - no need to switch to `SyntaxHighlighter`, which has been removed. Use triple backticks, with the opening backticks followed by the language name or abbreviation. [#562](https://github.com/plotly/dash-core-components/pull/562) Supported languages:
```
wbrgss: Maybe add that `Markdown` will attempt to auto-detect the language (via `highlight.js`) without specifying the language name

test/test_integration.py
```diff
@@ -71,2341 +71,2346 @@ def snapshot(self, name):
...
-    #        date_tokens = set(start_date.get_attribute('value').split('/'))
-    #        date_tokens.update(end_date.get_attribute('value').split('/'))
-    #
-    #        self.assertEqual(
-    #            set(itertools.chain(*[
-    #                _.split('-')
-    #                for _ in self.driver.find_element_by_css_selector(
-    #                    '#date-picker-range-output').text.split(' - ')])),
-    #            date_tokens,
-    #            "date should match the callback ou
```
wbrgss: :see_no_evil: :+1:

src/components/Markdown.react.js
```diff
@@ -25,14 +31,48 @@ class DashMarkdown extends Component {
...
+                if (!commonPrefix) {
+                    break;
+                }
+            }
+        }
+
+        const commonLen = commonPrefix ? commonPrefix.length : 0;
+        return lines
+            .map(line => {
+                return line.match(/\S/) ? line.substr(commonLen) : '';
```
wbrgss: :+1:

CHANGELOG.md
wbrgss: Also maybe mention the `theme` prop, but this could be in the docs for `dcc.Markdown` as well/instead

PR #570, opened 2019-06-15
Migrate integration

byronz:
This PR targets migrating test cases of `upload` and `store` components

it's not a simple convert of syntax, but includes the concept on how to structure the test code. Some test case have been rewritten like  `test_upft001_test_upload_with_different_file_types`. Some test cases are newly added as they were missing in the previous scope, e.g. `test_stda003_large_data_size`

few short terms migrating topics to follow:
1. cover markdown with all languages, the content itself is already heavily covered the dash-docs snapshots
2. rewrite the gallery all-in-one tests as a simpler smoke test. a very light app contains all dcc components in a dynamic fashion


.circleci/config.yml
```diff
@@ -26,8 +26,8 @@ jobs:
             - run:
                 name: Install dependencies (dash)
                 command: |
-                    git clone --depth 1 git@github.com:plotly/dash.git dash-main
-                    git clone --depth 1 git@github.com:plotly/dash-html-components.git
+                    git clone --depth 1 https://github.com/plotly/dash.git dash-main
+                    git clone --depth 1 https://github.com/plotly/dash-html-components.git
```
alexcjohnson: Why is https better than ssh?

byronz: if you use `circleci cli`, ssh will always ask to confirm `yes/no`, as the main purpose is to get the source code, so https is slightly more automatic than ssh

dev-requirements.txt
```diff
@@ -1,5 +1,4 @@
-percy
-selenium
 pandas
 xlrd
 flake8
+mimesis;python_version>="3.6"
```
byronz: there are two major data faker: mimesis, faker

mimesis only support 3.6+, but much faster for large dummy data generation.

tests/integration/store/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+from dash.dependencies import Input, Output, State
+from dash.exceptions import PreventUpdate
+import dash_core_components as dcc
+import dash_html_components as html
+
+UUID = "store-test-{}".format(uuid.uuid4().hex)
+
+
+@pytest.fixture(scope="module")
+def store_app():
```
byronz: it's another way to define the app, not inside tests but provide the app as another fixture. 

pros: share the same app among tests, make the test terser, even without import.
cons: you might need to open two windows to have a better understanding of the test.

tests/integration/store/test_component_props.py
```diff
@@ -0,0 +1,69 @@
...
+    dash_duo.find_element("#clear-btn").click()
+
+    assert (
+        not dash_duo.find_element("#output").text
+        and not dash_duo.get_local_storage()
+        and not dash_duo.get_session_storage()
+    ), "set clear_data=True should clear all data in three storage types"
+
+
+def test_stcp200_modified_ts(store_app, dash_duo):
```
byronz: stcp200, stdl001 were part of one gigantic test, I split them into smaller tests with specific focus

tests/integration/store/test_component_props.py
```diff
@@ -0,0 +1,69 @@
...
+import time
+from pytest import approx
+import dash
+from dash.dependencies import Input, Output, State
+from dash.exceptions import PreventUpdate
+import dash_core_components as dcc
+import dash_html_components as html
+
+
+def test_stcp100_clear_data_on_all_types(store_app, dash_duo):
```
byronz: the three digit here plays another role as sub category,  1xx will be test on `clear_data` property
and 2xx will on `modified_timestamps`

tests/integration/store/test_store_data.py
```diff
@@ -0,0 +1,161 @@
...
+    # it would of crashed the app before adding the recursive check.
+
+    list_btn.click()
+    dash_duo.wait_for_text_to_equal("#output", json.dumps(nested_list))
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 6),
+    reason="tests requires depedency only available in 3.6+",
+)
```
alexcjohnson: makes a lot of sense. Pretty soon it will be time to move Percy over to py3 I'd think, so we don't have to choose between a modern helper package and taking snapshots...

PR #574, opened 2019-06-21
Build Dash 1.0.0 version of dash-core-components for R

rpkyle:
This PR proposes to update the dashCoreComponents package to v1.0.0.

This is a breaking change in Dash for R, and this library will also require v1.0.0 of `dash` due to externalized property types ([#767](https://github.com/plotly/dash/pull/767/)).


alexcjohnson:
I've mostly seen `.yml` but this is a funny read https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml?rq=1

not sure why the tests keep failing, looks unrelated to the changes. Might be something flaky in this afternoon's update... if it doesn't pass after a few more retries I can look at it Monday.

Issue #513, opened 2019-04-05
dcc.Input with type="number" unable to clear


oliverbrace:
When one has dcc.Input with type="number"  it is not possible to clear what is inside the input like with dcc.Input with type="text". I assume this is because deleting it so nothing is contained sets it to "" which is not in the list of allowed values for numbers. 
However this results in placeholder text being irritativable in "input type number" once text has been inserted. 

I also have input boxes, which when blank, uses a default value within the code. Currently I have to create a button and a callback which clears the input box, however .

HTML input type number does allow for blank values (and the letter e), however something is then changing it to be 0. 

I feel that the functionality should be the same as hmtl.input and allow for "" to be an accepted value.

alexcjohnson:
related: #455

Issue #455, opened 2019-02-03
validation of dcc.Input with type="number" makes entering negative numbers difficult


ned2:
When a `dcc.Input` component has `type="number"`, it's difficult to get the field to accept negative numbers due to the client-side validation. A `-` character is not permitted unless it precedes a number, which means you have to to enter the digits of your desired number first, before moving the cursor to the start of the field and entering the `-` character.

When the last character in the field is deleted the field's contents are also immediately set to `0`, which I find incurs some amount of cognitive friction when editing the field's contents. I'm guessing both these properties of the validation are intended to always ensure the current value of the field is a valid number, however this comes at the cost of usability I think. I also wonder how important this invariant is to maintain given that an invalid (ie non-numeric) value can be set as an initial value in the layout and therefore can be sent through to the callback.

Marc-Andre-Rivet:
@ned2 Thanks for pointing this issue out. Agree that usability is taking a toll here and needs to be improved -- something as simple as behaving like a normal ```<input type='number' />``` would be better.

Marc-Andre-Rivet:
Discussion about the feature's details in the PR, starting around:
https://github.com/plotly/dash-core-components/pull/580#pullrequestreview-260869321

PR #580, opened 2019-07-10
Rework on Input 

byronz:
this fixes both #455 and #513,  the root cause of all the inconsistent behavior is the `Number()` casting.


byronz:
feedback is welcome before I'm adding a few test cases for `type=number`

alexcjohnson:
Agreed - the prop value should be a number. Interestingly, playing with it now in Chrome, the browser does character filtering to the set `1234567890eE+-.` but no more sanity checks than that (you can write `+-+-eEeEe-+-+` if you so desire...) other than a built-in tooltip that shows up if the number is invalid:

Screen Shot 2019-07-12 at 12 25 02 PM

I'd propose we allow the user to type whatever they want, and have separate focused and blurred behavior:
- While focused: don't rerender under any circumstances, and only set the prop if the value was successfully cast to a number.
- On blur: keep current behavior - cast to a number, throw the input away if casting fails.

So here are two cases to consider
- User types junk like `eEeEe` -> it stays there while focused, the prop is unchanged from whatever value it had previously. On blur we rerender, and either put back the previous value or clear the field (configurable like table edits? For now just do whatever matches the current behavior most closely)
 only set the prop if the value was successfully cast to a number.
- User types a nonstandard variant of a number, like `1e+1` -> it stays like that while focused, while the prop is set to (number) `10`. On blur we rerender as `10`.

Seem reasonable?

byronz:
demo

Marc-Andre-Rivet:
Looking at the code, this component is a bit weird at the moment.

There's 3 scenarios where we `set` value:
- onChange: Only if `debounce=false`, this does a min/max check
- onBlur / onEnter: Only if `debounce=true`, this does NOT do a min/max check

🐛 Separate bug to fix the min/max check -- whether we fix it at the same time or not is up to us.

> While focused: don't rerender under any circumstances, and only set the prop if the value was successfully cast to a number.
On blur: keep current behavior - cast to a number, throw the input away if casting fails.

- Whatever happens on blur should happen on `Enter` too
- Clearing the input if the cast fails seems counter-intuitive to me. This is not how the input component behaves by default. I'd try to stay as close to the normal behavior as possible
- I'd suggest that setting value should ALWAYS happen for blur/enter whether debounce is true/false.
If the cast fails we send back `None`
- Maybe we could add some usability / visual indicator when the value is invalid instead of erasing it: some red border, etc. (this is technically an enhancement and out of scope)

byronz:
> * Clearing the input if the cast fails seems counter-intuitive to me. This is not how the input component behaves by default. I'd try to stay as close to normal behavior as possible

yes, I don't feel the clearing input while `onblur` or `Enter keypress` when cast failed is a good solution, something else like a tooltip or red box might be more intuitive. 

> * I'd suggest that setting value should ALWAYS happen for blur/enter whether debounce is true/false.
>   If the cast fails we send back `None`

`None` is not clear for me @Marc-Andre-Rivet  can you be more specific?

Marc-Andre-Rivet:
@byronz By `None` I mean what the DashPy user would get. In JS, I'm suggesting we set the value to `null`.

byronz:
# debounce => False
debounce_false
# debounce => True
debounce_true

@alexcjohnson  @Marc-Andre-Rivet  please review the improved error handling behavior with numbers

take note that browser have different native behavior for `input[number]` mode:

-  Chrome is the only one does the key filtering, i.e. only accepts valid number range, but has no number validation
- Firefox  accepts any keypress but does a number validation with a css effect of `box-shadow: red` 
- Safari does neither key filtering nor number validation

Marc-Andre-Rivet:
The behavior seems inconsistent between Firefox and (Chrome + Safari) when entering an invalid value — in Firefox the value stays in the input while in the other two the value becomes blank. In all cases the default behavior of the html <input> field is to retain the value.

I think this is a side-effect of the implementation causing `undefined` to be passed to the `input` element / handled differently across browsers.

Marc-Andre-Rivet:
@byronz From our previous discussion and using the code you gave me, here's the behavior I'm seeing in Chrome (top) and Firefox (bottom). Notice that Chrome clears the input while Firefox does not.

I am using `tab` and `shift+tab` to move focus around and trigger the blur.

```
import dash

app = dash.Dash(__name__)
app.layout = html.Div(
    [
        dcc.Input(
            id="dfalse",
            type="number",
            debounce=False,
            placeholder="debounce=>False",
        ),
        html.Div(id='out_false', children=''),
        html.Hr(id='bl'),
        dcc.Input(
            id="dtrue",
            type='number',
            debounce=True,
            placeholder="debounce=>True",
        ),
        html.Div(id='out_true', children=''),
    ]
)

@app.callback(
    [Output('out_false', 'children'), Output('out_true', 'children')],
    [Input('dfalse', 'value'), Input('dtrue', 'value')])
def render(fval, tval):
    return fval, tval


if __name__ == "__main__":
    app.run_server(debug=True)
```

Chrome-input-undef

Firefox-input-undef

byronz:
this is another approach by hijacking the number type inside the `Input` component and applying similar validation to have consistent behavior. 
the previous solution will end up having consistent frontend behavior, but breaking the backend contract when returning empty string,   i.e. it gives '' instead of none

textmode_2

byronz:
as discussed with @Marc-Andre-Rivet  we needs an ad-hoc notes in dcc.Input documentation. for edge case like 0.0 => 0, 1.0 => 1 in number mode. this is a limitation of js/json convert, and the returned value on python side might need an extra guard like `float()` casting.

Marc-Andre-Rivet:
@byronz The last implementation does not behave correctly either. Entering an incorrect value on the `debounce=True` field makes its text disappear.

Marc-Andre-Rivet:
Behavior seems fine now!

src/components/Input.react.js
```diff
@@ -9,76 +16,73 @@ import {omit, isEmpty} from 'ramda';
...
+        this.setInputValue(
+            R.isNil(valueAsNumber) ? value : valueAsNumber,
+            this.props.value
+        );
     }
 
     render() {
-        const {setProps, type, min, max, debounce, loading_state} = this.props;
-        const value = this.state.value;
+        if (this.props.type === 'number') {
```
Marc-Andre-Rivet: Here you can get rid of the double implementation by doing

```
        let props = this.props.type.number === 'number' ?
            {} :
            { value: this.props.value };

return (<input {...props} ... />);
```

src/components/css/input.css
```diff
@@ -0,0 +1,3 @@
+.invalid {
```
Marc-Andre-Rivet: I think this style is not used anymore. We're not giving a visual indicator that the value is not valid.

tests/integration/input/test_number_input.py
```diff
@@ -0,0 +1,63 @@
...
+        for debounce in ("false", "true"):
+
+            elem = dash_duo.find_element("#input_{}".format(debounce))
+            assert not elem.get_attribute("value")
+
+            elem.send_keys(invalid_number)
+            elem.send_keys(Keys.TAB)  # onblur
+
+            # the number input element is special now, selenium doesn't work
+            dash_duo.percy_snapshot("inni001 - invalid number input onblur")
```
Marc-Andre-Rivet: Unable to look at the Percy results atm but from the style above, assuming these invalid values do not highlight the input red.

byronz: needs a fix on dash.testing side for the percy issue, will work on that. 

the current behavior is close to native one, you get red on firefox though

src/components/css/input.css
byronz: right

PR #589, opened 2019-07-30
Plotlyjs 1.49.0 upgrade

### LINKED ISSUE #577: Bump Plotly.js to 1.49.0
Once https://github.com/plotly/plotly.js/milestone/45 is released, bump in DCC.

shammamah:
Closes #577


alexcjohnson:
(oh looks like there's a new config option that needs propagating to the component props... anyway the 💃 stands once you get tests to pass)

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.1.0] -
```
Marc-Andre-Rivet:
```suggestion
## [Unreleased] -
```

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
...
+### Updated
+- Upgraded plotly.js to 1.49.0 [#589](https://github.com/plotly/dash-core-components/pull/589)
+  - [Feature release 1.49.0](https://github.com/plotly/plotly.js/releases/tag/v1.49.0) which contains:
+
```
Marc-Andre-Rivet: @alexcjohnson Could you fill this in with the most significant items from the new release?

src/components/Graph.react.js
```diff
@@ -474,6 +474,11 @@ const graphPropTypes = {
             'reset+autosize',
         ]),
 
+	/**
+	 * Delay for registering a double-click event in ms
+	 */
+	doubleClickDelay: PropTypes.number,
```
shammamah: @Marc-Andre-Rivet The prop in plotly.js [has min/max values defined](https://github.com/pynklu/plotly.js/blob/9dcbc716c81ed995020f31a2622e73cf67bccb46/src/plot_api/plot_config.js#L178), but I don't think there's a way to do that here. Should I add the min/max values to the docstring?

Marc-Andre-Rivet: You're right. Proptypes support runtime validation but our use of react-docgen does not. The docstring is best. :)

CHANGELOG.md
alexcjohnson: OK, added. I'm not sure I need to always be the one to do this though, I might have a bit more context about this work but even I'm mostly just summarizing the biggest additions from the release notes.

rpkyle:
> @shammamah Nice catch cleaning the old ones out of `inst/deps` before any more pile up! 💃
> 
> @rpkyle Let's push this through now for the current update to get properly applied, but should we be emptying out that dir during package generation, before refilling it with files found from the Python side? Is there ever a case we need to leave something in that dir that's not found in Python? Please make an issue in the dash repo to follow up on this. Then hopefully future plotly.js updates here can leave `inst/deps` alone, and it will be updated during the release process.

https://github.com/plotly/dash/issues/847

Issue #594, opened 2019-08-06
DatePickerRange : Date is not cleared with clearable=True


Sandy-T:
dash 1.1.0
dash-core-components 1.1.0

byronz:
reproduced with the following code, the root cause was the `x` button only clears the state, but not the props. 

```python
import dash
from dash.dependencies import Input, Output
import dash_table
import dash_core_components as dcc
import dash_html_components as html


app = dash.Dash(__name__)

app.layout = html.Div([dcc.DatePickerRange(id="date", clearable=True)])

if __name__ == "__main__":
    app.run_server(debug=True)
```

Issue #593, opened 2019-08-02
fix one flakey test


byronz:
https://circleci.com/gh/plotly/dash-core-components/5668

this test case is flakey, should fixed with new structure
```
―――――――――――――――――――― Tests.test_datepickerrange_updatemodes ――――――――――――――――――――

self = <tests.test_integration.Tests testMethod=test_datepickerrange_updatemodes>

    def test_datepickerrange_updatemodes(self):
        app = dash.Dash(__name__)
    
        app.layout = html.Div([
            dcc.DatePickerRange(
                id='date-picker-range',
                start_date_id='startDate',
                end_date_id='endDate',
                start_date_placeholder_text='Select a start date!',
                end_date_placeholder_text='Select an end date!',
                updatemode='bothdates'
            ),
            html.Div(id='date-picker-range-output')
        ])
    
        @app.callback(
            dash.dependencies.Output('date-picker-range-output', 'children'),
            [dash.dependencies.Input('date-picker-range', 'start_date'),
             dash.dependencies.Input('date-picker-range', 'end_date')])
        def update_output(start_date, end_date):
            return '{} - {}'.format(start_date, end_date)
    
        self.startServer(app=app)
    
        start_date = self.wait_for_element_by_css_selector('#startDate')
        start_date.click()
    
        end_date = self.wait_for_element_by_css_selector('#endDate')
        end_date.click()
    
        self.wait_for_text_to_equal('#date-picker-range-output', 'None - None')
    
        # using mouse click with fixed day range, this can be improved
        # once we start refactoring the test structure
        start_date.click()
    
        sday = self.driver.find_element_by_xpath("//td[text()='1' and @tabindex='0']")
        sday.click()
        self.wait_for_text_to_equal('#date-picker-range-output', 'None - None')
    
        eday = self.driver.find_elements_by_xpath("//td[text()='28']")[1]
>       eday.click()

tests/test_integration.py:1230: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
venv/lib/python2.7/site-packages/selenium/webdriver/remote/webelement.py:80: in click
    self._execute(Command.CLICK_ELEMENT)
venv/lib/python2.7/site-packages/selenium/webdriver/remote/webelement.py:633: in _execute
    return self._parent.execute(command, params)
[...]

PR #599, opened 2019-08-07
Fix clearable

byronz:
this PR fixes both #593 and #594

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,6 +73,7 @@ workflows:
   version: 2
   build:
     jobs:
+      - percy-finalize
       - 'python-2.7'
       - "percy-finalize":
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,10 +1,13 @@
 version: 2
 jobs:
-  "percy-finalize":
+  "percy-finalize": &percy
       docker:
       - image: percyio/agent
       steps:
       - run: percy finalize --all
+
+  "percy-reset":
+        <<: *percy
 
   'python-2.7': &test-template
       docker:
@@ -73,7 +76,7 @@ workflows:
   version: 2
   build:
     jobs:
-      - percy-finalize
+      - percy-reset
       - 'python-2.7'
       - "percy-finalize":
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,10 @@
 version: 2
 jobs:
-  "percy-finalize": &percy
+  "percy-finalize":
       docker:
       - image: percyio/agent
       steps:
       - run: percy finalize --all
-
-  "percy-reset":
-        <<: *percy
 
   'python-2.7': &test-template
       docker:
@@ -15,6 +12,7 @@ jobs:
             environment:
                 PYTHON_VERSION: py27
                 PERCY_PARALLEL_TOTAL: '-1'
+          - image: percyio/agent
 
       steps:
           - checkout
@@ -56,6 +54,11 @@ jobs:
                     python --version
                     npm run test
 
+          - run:
+                name: Percy TearDown
+                command: percy --finalize -all
+                when: on_fail
+
   'python-3.6':
       <<: *test-template
       docker:
@@ -76,8 +79,7 @@ workflows:
   version: 2
   build:
     jobs:
-      - percy-reset
-      - 'python-2.7'
+      - python-2.7
       - "percy-finalize":
           requires:
             - 'python-2.7'
--- a/src/components/DatePickerRange.react.js
+++ b/src/components/DatePickerRange.react.js
@@ -76,8 +76,8 @@ export default class DatePickerRange extends Component {
             clearable &&
             !start_date &&
             !end_date &&
-            oldMomentDates.start_date &&
-            oldMomentDates.end_date
+            (oldMomentDates.start_date !== start_date ||
+                oldMomentDates.end_date !== end_date)
         ) {
             setProps({start_date: null, end_date: null});
         }
--- a/src/components/DatePickerRange.react.js
+++ b/src/components/DatePickerRange.react.js
@@ -72,7 +72,13 @@ export default class DatePickerRange extends Component {
             }
         }
 
-        if (clearable && !start_date && !end_date) {
+        if (
+            clearable &&
+            !start_date &&
+            !end_date &&
+            oldMomentDates.start_date &&
+            oldMomentDates.end_date
+        ) {
             setProps({start_date: null, end_date: null});
         }
     }
--- a/src/components/DatePickerRange.react.js
+++ b/src/components/DatePickerRange.react.js
@@ -46,7 +46,7 @@ export default class DatePickerRange extends Component {
     }
 
     onDatesChange({startDate: start_date, endDate: end_date}) {
-        const {setProps, updatemode} = this.props;
+        const {setProps, updatemode, clearable} = this.props;
 
         const oldMomentDates = convertToMoment(this.state, [
             'start_date',
@@ -70,6 +70,10 @@ export default class DatePickerRange extends Component {
                     end_date: end_date.format('YYYY-MM-DD'),
                 });
             }
+        }
+
+        if (clearable && !start_date && !end_date) {
+            setProps({start_date: null, end_date: null});
         }
     }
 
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -53,6 +54,11 @@ jobs:
                     python --version
                     npm run test
 
+          - run:
+                name: Percy TearDown
```
byronz: there are numerous requests on circleci community about conditional `Jobs` upon previous failure. right now, the behavior is when the previous job fails, the percy finalize job cannot be executed.

tests/integration/calendar/test_calendar_props.py
```diff
@@ -0,0 +1,110 @@
...
+import dash_core_components as dcc
+import dash_html_components as html
+import dash
+from dash.dependencies import Input, Output
+import dash.testing.wait as wait
+
+DAY_SELECTOR = 'div[data-visible="true"] td.CalendarDay'
+
+
+@pytest.mark.DCC594
```
byronz: this is an idea I have in mind to associate the test case with issue number, could be helpful for test selection by associating the test case with the issue

src/components/DatePickerRange.react.js
```diff
@@ -71,6 +71,10 @@ export default class DatePickerRange extends Component {
                 });
             }
         }
+
```
byronz: @Marc-Andre-Rivet  needs your golden eye to check the fix

src/components/DatePickerRange.react.js
```diff
@@ -71,6 +71,10 @@ export default class DatePickerRange extends Component {
...
+        if (clearable && !start_date && !end_date) {
+            setProps({start_date: null, end_date: null});
+        }
```
Marc-Andre-Rivet: This fix correctly handles the `x` action case - only additional fix is that it might setting one of the date props superfluously if already / still `None`.

Leaving the scope of this specific issue / fix and looking into the logic of this component as a whole. Multiple scenarios seem wrongly implemented and I think we should use this opportunity to address them instead of just fixing this one case.

For example, setting `end_date` first with `updatemode=bothdates` will trigger setProps, setting `start_date` afterwards will not, even though both values are now set -- there's an assumption here that start is always set before end...; manually resetting a value with `clearable=True` is possible if only one value is set but not if both are, etc.

@alexcjohnson My feeling is we should pay the price and fix these components in depth as we work on them. I'm almost positive many other comps will exhibit deep logic flaws as we start going through them. Would rather fix them for real than make 'cosmetic' fixes that addresses whatever specific issue was logged.

[Warning: Wall of text...]

Four props are involved in the update logic: `start_date`, `end_date`, `clearable`, `updatemode`.

The possible operation / cases the user might trigger are:
With `clearable=True|False`
- set start_date w/ end_date unset
- set end_date w/ start_date unset
- set start_date w/ end_date set
- set end_date w/ start_date set

With `clearable=True`
- manually unset start_date w/ end_date unset
- manually unset end_date w/ start_date unset
- manually unset start_date w/ end_date set
- manually unset end_date w/ start_date set
- clear w/ start_date set, end_date unset
- clear w/ start_date unset, end_date set
- clear w/ start_date set, end_date set

All combinations and expected behaviors, hopefully without mistakes:

`clearable=False` + `updatemode=singledate`
- set start_date w/ end_date unset -> setProps called w/ start_date only
- set end_date w/ start_date unset -> setProps called w/ end_date only
- set start_date w/ end_date set -> setProps called w/ start_date only
- set end_date w/ start_date set -> setProps called w/ end_date only
- manually unset start_date w/ end_date unset -> does not unset
- manually unset end_date w/ start_date unset -> does not unset
- manually unset start_date w/ end_date set -> does not unset
- manually unset end_date w/ start_date set -> does not unset
- clear w/ start_date set, end_date unset -> not available
- clear w/ start_date unset, end_date set -> not available
- clear w/ start_date set, end_date set -> not available

`clearable=False` + `updatemode=bothdates`
- set start_date w/ end_date unset -> state updated, no setProps call
- set end_date w/ start_date unset -> state updated, no setProps call
- set start_date w/ end_date set -> setProps called w/ start_date and end_date
- set end_date w/ start_date set -> setProps called w/ start_date and end_date
- manually unset start_date w/ end_date unset -> does not unset
- manually unset end_date w/ start_date unset -> does not unset
- manually unset start_date w/ end_date set -> does not unset
- manually unset end_date w/ start_date set -> does not unset
- clear w/ start_date set, end_date unset -> not available
- clear w/ start_date unset, end_date set -> not available
- clear w/ start_date set, end_date set -> not available

`clearable=True` + `updatemode=singledate`
- set start_date w/ end_date unset -> setProps called w/ start_date only
- set end_date w/ start_date unset -> setProps called w/ end_date only
- set start_date w/ end_date set -> setProps called w/ start_date only
- set end_date w/ start_date set -> setProps called w/ end_date only
- manually unset start_date w/ end_date unset -> unsets, setProps called with start_date only
- manually unset end_date w/ start_date unset -> unsets, setProps called with end_date only
- manually unset start_date w/ end_date set -> unsets, setProps called with start_date only
- manually unset end_date w/ start_date set -> unsets, setProps called with end_date only
- clear w/ start_date set, end_date unset -> unsets, setProps called with start_date
- clear w/ start_date unset, end_date set -> unsets, setProps called with end_date
- clear w/ start_date set, end_date set -> unsets, setProps called with start_date and end_date

`clearable=True` + `updatemode=bothdates`
- set start_date w/ end_date unset -> state updated, no setProps call
- set end_date w/ start_date unset -> state updated, no setProps call
- set start_date w/ end_date set -> setProps called w/ start_date and end_date
- set end_date w/ start_date set -> setProps called w/ start_date and end_date
- manually unset start_date w/ end_date unset -> unsets, setProps called with start_date
- manually unset end_date w/ start_date unset -> unsets, setProps called with end_date
- manually unset start_date w/ end_date set -> unsets, no setProps call
- manually unset end_date w/ start_date set -> unsets, no setProps call
- clear w/ start_date set, end_date unset -> unsets, setProps called with start_date
- clear w/ start_date unset, end_date set -> unsets, setProps called with end_date
- clear w/ start_date set, end_date set -> unsets, setProps called with end_date and start_date

NB. The above considers that `bothdates` triggers setProps call if either both values are defined or both vaules are None.

byronz: I OK to increase the scope, but prefer to close this PR and start a `wall` issue

Marc-Andre-Rivet: Alright. In that case just the first part of the comment, before the divider, applies.

Marc-Andre-Rivet: Please open a `wall` issue for the follow up :)

src/components/DatePickerRange.react.js
```diff
@@ -71,6 +71,16 @@ export default class DatePickerRange extends Component {
...
+
+        if (
+            clearable &&
+            !start_date &&
+            !end_date &&
+            oldMomentDates.start_date &&
+            oldMomentDates.end_date
+        ) {
+            setProps({start_date: null, end_date: null});
+        }
```
Marc-Andre-Rivet: Rather than preventing the update if one of start/end was not previously set, should instead check and add each individually if the condition is met.

e.g. this code is not optimal but gives the general idea
```
if (clearable && (!start_date || !end_date)) {
    const props = { };
    if (!start_date && oldMomentDates.start_date) {
        props.start_date = null;
    }
    // idem for end_date

    setProps(props);
}
```

src/components/DatePickerRange.react.js
```diff
@@ -71,6 +71,16 @@ export default class DatePickerRange extends Component {
                 });
             }
         }
+
+        if (
+            clearable &&
+            !start_date &&
+            !end_date &&
+            (oldMomentDates.start_date !== start_date ||
+                oldMomentDates.end_date !== end_date)
```
byronz: @Marc-Andre-Rivet  when you click the `x`, both start and end will be set to null, the condition can be either one date changed .

tests/integration/calendar/test_calendar_props.py
alexcjohnson: @byronz we can keep this if it's useful, but we should do something to clear the warning it currently gives:

> /Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/_pytest/mark/structures.py:324: PytestUnknownMarkWarning:
> Unknown pytest.mark.DCC594 - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html

PR #602, opened 2019-08-10
:alembic: first try

byronz:
apologies for the messy commit history, I can refactor it in another PR if needed. 

I don't know why, but the spin-up environment in DCC is slower in Dash, it might take 30s-1minutes versus 10-15s in Dash. this has an impact on the speedup as now we have more separate jobs, but in the longer term, we will still benefit from the change with more integration tests to come. Looks like most of the time, they cannot hit a docker cache layer, thus requires an extra pull down from remote. 

the percy finalize is still a problem, the ideal way is to use the current finalize job as configured in the yaml file. But circleci has a limitation that if test job failed, the finalize job cannot be executed as it requires on the test job to be finished. 

- [ ] We need to change the github repo setting to reselect the mandatory CI jobs.


.circleci/config.yml
```diff
@@ -144,13 +144,15 @@ jobs:
       - image: circleci/python:3.6-stretch-node-browsers
         environment:
           PYTHON_VERSION: py36
+          PERCY_ENABLE: '0'
```
alexcjohnson: Screen Shot 2019-08-12 at 8 38 53 PM

I guess also take `percy-finalize` out of the 2.7 and 3.6 workflows?

PR #610, opened 2019-08-22
Update and fix docstrings.

shammamah:
Make docstrings for `dcc.Graph` compliant with the guidelines in https://github.com/plotly/dash/issues/853. 

Additionally fix some items in the CircleCI config.

### DIFF
--- a/src/components/Graph.react.js
+++ b/src/components/Graph.react.js
@@ -403,7 +403,7 @@ const graphPropTypes = {
          */
         edits: PropTypes.exact({
             /**
-             * annotationPosition: the main anchor of the annotation, which is the
+             * The main anchor of the annotation, which is the
              * text (if no arrow) or the arrow (which drags the whole thing leaving
              * the arrow length & direction unchanged)
              */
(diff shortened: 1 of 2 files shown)


src/components/Graph.react.js
```diff
@@ -540,12 +540,12 @@ const graphPropTypes = {
          * Remove mode bar button by name.
          * All modebar button names at https://github.com/plotly/plotly.js/blob/master/src/components/modebar/buttons.js
          * Common names include:
-         *  - sendDataToCloud
-         * - (2D): zoom2d, pan2d, select2d, lasso2d, zoomIn2d, zoomOut2d, autoScale2d, resetScale2d
-         * - (Cartesian): hoverClosestCartesian, hoverCompareCartesian
```
alexcjohnson: `:` or no `:`? still a couple of them below.

Also https://github.com/plotly/dash/issues/853 says "no unindented bullet lists (-)" - does that mean indented bullet lists are still OK, or was that just me hypothesizing about something we could choose to support? It's ok in this case, I guess since there's nothing in the docstring after the list, but if another case were to have additional text I'd have pushed to keep the bullets.

shammamah: Good catch - those `:` should also be removed now. I think indented bullet lists are OK; IIRC the reason we wanted to remove *unindented* bullet lists is that they made the separation between props more difficult to identify. If lists are indented, it still maintains that separation.

PR #616, opened 2019-08-26
Async support - Lazy dcc.Graph and lazy plotlyjs

Marc-Andre-Rivet:
Companion PR for https://github.com/plotly/dash-table/pull/554

---

- [x] Use `plotly-dist` for `dash_core_components/plotly-x.xx.x.js` instead of having two copies to keep in sync
- [x] Run `dcc.Graph` tests in both `eager` and `lazy` mode
- [x] Test `window.Plotly` loading / behavior for lazy / eager / explicitly loaded Plotlyjs through asset

Note: `plotly.js-dist` only provides a non-minified version of `plotly.js`, switched over to using `plotly.js` directly and resolving the desired js with `resolve.alias` in webpack..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Added
+- [#616](https://github.com/plotly/dash-core-components/pull/616) Async Graph and Plotly.js
+
 ## [1.3.1] - 2019-10-17
 ### Updated
 - Upgraded plotly.js to 1.50.1 [#681](https://github.com/plotly/dash-core-components/issues/681)
@@ -13,7 +17,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [1.3.0] - 2019-10-08
 ### Added
 - Added `search_value` prop to `Dropdown`, for server-side options loading/filtering. [#660](https://github.com/plotly/dash-core-components/pull/660)
-- [#616](https://github.com/plotly/dash-core-components/pull/616) Async Graph and Plotly.js
 
 ### Updated
 - Upgraded plotly.js to 1.50.0 [#675](https://github.com/plotly/dash-core-components/pull/675)


src/components/Graph.react.js
```diff
@@ -73,208 +54,92 @@ const filterEventData = (gd, eventData, event) => {
...
-            animate &&
-            this._hasPlotted &&
-            figure.data.length === gd.data.length
-        ) {
-            return Plotly.animate(gd, figure, animation_options);
+    componentDidMount() {
+        if (this.props.extendData) {
+            this.setState({
+                extendData: [this.props.extendData],
+            });
```
Marc-Andre-Rivet: `extendData` does not respect the usual pattern of props for Dash - as the wrapper waits for the real component to be loaded, it accumulates updates to extendData, the real component is now capable of processing an array of `extendData` instead of an object. The array is cleared upon processing.

src/components/Graph.react.js
```diff
@@ -73,208 +54,92 @@ const filterEventData = (gd, eventData, event) => {
...
-    shouldComponentUpdate(nextProps) {
+    render() {
         return (
-            this.props.id !== nextProps.id ||
-            JSON.stringify(this.props.style) !== JSON.stringify(nextProps.style)
+            <ControlledPlotlyGraph
+                {...{
+                    ...this.props,
+                    extendData: this.state.extendData,
+                    clearExtendData: this.clearExtendData,
```
Marc-Andre-Rivet: When the real component is done processing `extendData` entries, it calls `clearExtendData`

src/components/Graph.react.js
```diff
@@ -73,208 +54,92 @@ const filterEventData = (gd, eventData, event) => {
...
-        }
-    }
-
-    componentDidUpdate(prevProps) {
-        if (prevProps.id !== this.props.id) {
-            this.plot(this.props);
-        }
-    }
+const LazyPlotlyGraph = isLazyComponentReady(PlotlyGraph, () =>
+    loader.plotly().then(() => loader.graph())
```
Marc-Andre-Rivet: Load plotlyjs first (if necessary) then the graph component. `isLazyComponentReady` decorates the wrapper component in order for the renderer to know when it's ready.

`isLazyComponentReady` is currently duplicated in dcc and the table.

src/fragments/Graph.lazy.react.js
```diff
@@ -0,0 +1,304 @@
...
+                function generateIndices(data) {
+                    return Array.from(Array(getFirstProp(data).length).keys());
+                }
+                traceIndices = generateIndices(updateData);
+            }
+
+            const gd = this.gd.current;
+            return Plotly.extendTraces(gd, updateData, traceIndices, maxPoints);
+        });
+        clearExtendData();
```
Marc-Andre-Rivet: Like the old extend code but handles an array / clears

tests/integration/misc/test_platter.py
```diff
@@ -10,6 +10,10 @@ def test_mspl001_dcc_components_platter(platter_app, dash_duo):
     dash_duo.start_server(platter_app)
 
     dash_duo.wait_for_element("#waitfor")
+
+    # wait for Graph to be ready
+    dash_duo.wait_for_element("#graph .main-svg")
```
Marc-Andre-Rivet: Need to wait for the graphs to be loaded -- waiting for the svg

tests/unit/Tabs.test.js
```diff
@@ -102,16 +102,24 @@ describe('Tabs correctly appends classes', () => {
...
+            )
         );
     });
     test('props.parent_className=>', () => {
-        expect(tabs.prop('class')).toEqual('tab-parent ' + testParentClass);
+        expect(tabs.prop('class')).toEqual(
+            expect.stringMatching(
+                new RegExp(`jsx-\\d+\\stab-parent\\s${testParentClass}`)
+            )
+        );
```
Marc-Andre-Rivet: Not sure why this shows up in the diffs..

generator/plotly.js
```diff
@@ -0,0 +1,14 @@
...
+const plotlyJsVersion = (package.dependencies && package.dependencies['plotly.js']) ||
+    (package.devDependencies && package.devDependencies['plotly.js']);
+
+fs.copyFile('node_modules/plotly.js/dist/plotly.min.js', `dash_core_components/plotly-${plotlyJsVersion}.min.js`, err => {
+    if (err) {
+        throw err;
+    }
+
+    console.log('copied plotly.js', plotlyJsVersion);
+});
```
Marc-Andre-Rivet: Copy plotly.min.js as part of the build's generator step.

src/components/Graph.react.js
```diff
@@ -73,208 +29,92 @@ const filterEventData = (gd, eventData, event) => {
...
-        const gd = this.gd.current;
-        return Plotly.extendTraces(gd, updateData, traceIndices, maxPoints);
-    }
+        if (
+            nextProps.extendData &&
+            this.props.extendData !== nextProps.extendData
+        ) {
+            extendData.push(nextProps.extendData);
+        } else {
+            extendData = EMPTY_EXTEND_DATA;
```
alexcjohnson: setting `extendData` to a module constant, then later pushing things into said module constant? Don't we need either a new empty array, or copy into a new longer array?

Marc-Andre-Rivet: No. Before it's used, the array is copied from the state with `slice`, the state array always being treated as immutable.
https://github.com/plotly/dash-core-components/pull/616/files/efe1e766b4c0ad4b23d04f319649f480d7bf11cf#diff-b2faf9645a4fcca0badb59ef2fba58a3R55

alexcjohnson: Ah got it - thanks.

generator/plotly.js
alexcjohnson: Remember to update https://github.com/plotly/dash-core-components/blob/dev/CONTRIBUTING.md#updating-plotlyjs - it's already out of date but will be waaaaay simpler after this change 😍

Marc-Andre-Rivet: Updating. It's not really all that much simpler though, as it's only updating one step in the whole generation process. https://github.com/plotly/dash-core-components/pull/616/commits/6c00ed28de5ff2e50d0bd2a6942939a5fea43c96

tests/integration/graph/test_graph_basics.py
```diff
@@ -11,8 +11,9 @@
 import dash.testing.wait as wait
 
 
-def test_grbs001_graph_without_ids(dash_dcc):
-    app = dash.Dash(__name__)
+@pytest.mark.parametrize("is_eager", [True, False])
```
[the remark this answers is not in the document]
Marc-Andre-Rivet: After a few tweaks and figuring out the equivalent functions vs. `self.`, this seems to be working fine :)

generator/plotly.js
alexcjohnson: ah true... though partly that's because the previous version was missing a step - clearing out the old plotly.js version manually - that was a source of errors in the past. Further simplification will come when we auto-generate `__init__.py` in its entirety.

Marc-Andre-Rivet:
@alexcjohnson Last concern / mystery solved for this PR..

https://github.com/plotly/dash-core-components/pull/616/files#r327786637

Amongst the mass of merges / tweaks for various unrelated changes that had to be done, I forgot I did this change for the tab tests.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [1.3.0] - 2019-10-08
 ### Added
 - Added `search_value` prop to `Dropdown`, for server-side options loading/filtering. [#660](https://github.com/plotly/dash-core-components/pull/660)
+- [#616](https://github.com/plotly/dash-core-components/pull/616) Async Graph and Plotly.js
```
alexcjohnson: `Unreleased`

PR #634, opened 2019-09-02
Math in dcc.Markdown

anders-kiaer:
Adding this pull request to show how the `dcc.Markdown` component can be extended to also support rendering of LaTeX math equations using [react-katex](https://www.npmjs.com/package/react-katex) based on [KaTeX](https://katex.org/), for those users who need this feature. Feel free to close if not aligned with `dcc` development plans.

In order to explain/describe some given data set visualized in a Dash app, it is sometimes useful to be able to include mathematical equations and symbols. Instead of making a new Dash component, this PR extends the `dcc.Markdown` component such that math easily can be added in between markdown text. E.g. you can do:

```python
import dash

app = dash.Dash(__name__)

app.layout = dcc.Markdown(r'''

Below are the different Maxwell equations listed,

$$
\begin{aligned}
\nabla \cdot \mathbf{E} &= \frac {\rho} {\varepsilon_0} \\
\nabla \cdot \mathbf{B} &= 0 \\
\nabla \times \mathbf{E} &= -\frac{\partial \mathbf{B}} {\partial t} \\
\nabla \times \mathbf{B} &= \mu_0\left(\mathbf{J} + \varepsilon_0 \frac{\partial \mathbf{E}} {\partial t} \right)
\end{aligned}
$$

which form the foundation of classical electromagnetism. 

Using inline math you can describe the different parameters that goes into
the Maxwell's equations, e.g. $\varepsilon_0$ is permittivity of free space.
''')

if __name__ == '__main__':
    app.run_server(debug=True)
```
will result in

image

Built files are (currently) not included in the PR (in order to only show the source code changes needed). I.e. you will need to run `npm install`, `npm run build` and `python setup.py install` first.

**Related:**
- https://community.plot.ly/t/mathjax-in-dash/6653/18
- https://github.com/plotly/dash-core-components/pull/194
- https://github.com/plotly/dash/issues/242

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -92,6 +92,14 @@ module.exports = (env, argv) => {
                         },
                     ],
                 },
+                {
+                    test: /\.(png|woff|woff2|eot|ttf|svg)(\?v=[0-9]\.[0-9]\.[0-9])?$/,
+                    use:[
+                        {
+                            loader: 'url-loader'
+                        }
+                    ]
+                }
             ],
         },
         resolve: {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -92,14 +92,6 @@ module.exports = (env, argv) => {
                         },
                     ],
                 },
-                {
-                    test: /\.(png|woff|woff2|eot|ttf|svg)(\?v=[0-9]\.[0-9]\.[0-9])?$/,
-                    use:[
-                        {
-                            loader: 'url-loader'
-                        }
-                    ]
-                }
             ],
         },
         resolve: {
--- a/demo/Demo.react.js
+++ b/demo/Demo.react.js
@@ -58,17 +58,6 @@ Some code
 \\\`\\\`\\\`
 
 Inline \\\`code\\\` example.
-
-***
-
-Some block mathematics:
-
-$$
-e^{ix} = \\\\cos x + i\\\\sin x.
-$$
-
-Some inline mathematics $\\\\alpha=2$ and $\\\\beta=3$.
-
 \`
 
 ReactDOM.render(<Markdown>{markdown}</Markdown>, mountNode);
--- a/demo/Demo.react.js
+++ b/demo/Demo.react.js
@@ -58,6 +58,17 @@ Some code
 \\\`\\\`\\\`
 
 Inline \\\`code\\\` example.
+
+***
+
+Some block mathematics:
+
+$$
+e^{ix} = \\\\cos x + i\\\\sin x.
+$$
+
+Some inline mathematics $\\\\alpha=2$ and $\\\\beta=3$.
+
 \`
 
 ReactDOM.render(<Markdown>{markdown}</Markdown>, mountNode);
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "color": "^3.1.0",
     "fast-isnumeric": "^1.1.3",
     "highlight.js": "^9.17.1",
+    "mathjax": "^3.0.0",
     "moment": "^2.20.1",
     "plotly.js": "1.52.2",
     "prop-types": "^15.6.0",
@@ -50,6 +51,7 @@
     "react-resize-detector": "^4.2.1",
     "react-select-fast-filter-options": "^0.2.3",
     "react-virtualized-select": "^3.1.3",
+    "remark-math": "^1.0.6",
     "uniqid": "^5.0.3"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -46,10 +46,12 @@
     "react-dates": "^20.1.0",
     "react-docgen": "^3.0.0",
     "react-dropzone": "^4.1.2",
+    "react-katex": "^2.0.2",
     "react-markdown": "^4.0.6",
     "react-resize-detector": "^4.2.1",
     "react-select-fast-filter-options": "^0.2.3",
     "react-virtualized-select": "^3.1.3",
+    "remark-math": "^1.0.6",
     "uniqid": "^5.0.3"
   },
   "devDependencies": {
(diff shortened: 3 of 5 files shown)


byronz:
@anders-kiaer  thanks for this PR, I think `markdown` is a logical place for maths equations. 
I will leave @wbrgss  and @Marc-Andre-Rivet  to comment on the possible increment of bundle size (compressed):  react-katex is 64.4kb and remark-math is about 1kb

We would also need an integration test to cover this feature using `dash.testing`. you can put the test inside a `markdown` folder under `tests/integration`.  A percy snapshot might be sufficient with a more comprehensive example.

webpack.config.js
```diff
@@ -75,7 +75,15 @@ module.exports = (env, argv) => {
                         },
                     ],
                 },
+                {
+                    test: /\.(png|woff|woff2|eot|ttf|svg)(\?v=[0-9]\.[0-9]\.[0-9])?$/,
+                    use:[
+                        {
+                            loader: 'url-loader'
+                        }
```
alexcjohnson: Presumably this is for the Katex fonts... where is this documented? what does this do to the bundle size?

anders-kiaer: Correct @alexcjohnson 👍 I would guess this is not the optimal / final way of including KaTeX fonts wrt. bundle size and loading speed, if KaTeX support in `dcc.Markdown` is something worth adding. Maybe keep KaTeX as an external thing 📦 (like [`highlight.js` is already](https://github.com/plotly/dash-core-components/blob/7bdeec55ea8734cc08fcc21086acad797b617238/dash_core_components/__init__.py#L50)) and load [KaTeX globally](https://katex.org/docs/browser.html#loading-as-global)?

_Another thing to consider if including math in the markdown component:_ Should it have an argument stating if math is to be rendered or not (such that it is not a breaking change if e.g. the user writes `$some text which should not be inline math$`)? Or maybe a better alternative: do not use `$...$` for inline math, but rather e.g. the [MathJax default `\(...\)`](http://docs.mathjax.org/en/latest/basic/mathematics.html?highlight=inline#tex-and-latex-input).

alexcjohnson: True, it wouldn't be hard to end up with two `$` on a line together accidentally. Thanks for bringing this up! So yeah, let's make this opt-in. Perhaps also with an option to choose new delimiters, though that's not so important if math rendering is opt-in.

Global loading and fonts frankly sounds like a headache - particularly as we need to maintain the ability to run airgapped. Perhaps the best would be to hold this PR until we finish sorting out lazy loading #616, then make sure KaTeX is only loaded when rendering math-enabled markdown? Thoughts @Marc-Andre-Rivet?

alexcjohnson:
Ooh there's now a [MathJax v3](http://docs.mathjax.org/en/latest/upgrading/whats-new-3.0.html) that claims a big speed boost, synchronous rendering, and modern build options. I wonder if that changes our thinking about how to support math in markdown?

I love KaTeX but it doesn't have an SVG renderer so can't be used with plotly.js. We've thought about some ridiculously hacky ways to convert it to SVG (it needs to be real SVG, not a foreign object) but I'm skeptical we could really do this sufficiently robustly.

So plotly.js is going to stay on MathJax for the foreseeable future, and it'll be a project in its own right to upgrade that to v3, but if we use MathJax here we'd at least be on a path toward a single math display package for all of Dash.

Thoughts?

renatobellotti:
This seems to be finished. Why is it not merged?

renatobellotti:
Thanks for your effort!

anders-kiaer:
There is now one commit using KaTeX, and one using MathJax. I must say the user experience of using MathJax **v3** is much better compared with MathJax v2. The [MathJax documentation on direct usage](https://github.com/mathjax/MathJax-demos-node/tree/master/direct#non-component-based-examples) of the `es5` modules is still not very extensive I would say, so it is a bit difficult to know how to best use MathJax within a JavaScript library, together with `import`. There is some information in this issue comment: https://github.com/mathjax/MathJax/issues/2194#issuecomment-530625793

For now at least, this WIP PR uses a ready-made MathJax component `tex2svg`. The component seems to register itself to `window` on `import`.

A test example below, where user input (including math) is rendered live and dynamically (which can be tested after `npm install --ignore-scripts && npm run build && pip install -e .`), for those that want to try it out:
```python
import dash

SOME_MARKDOWN = r"""
Below are the different Maxwell equations listed,

$$
\begin{aligned}
\nabla \cdot \mathbf{E} &= \frac{\rho}{\varepsilon_0} \\
\nabla \cdot \mathbf{B} &= 0 \\
\nabla \times \mathbf{E} &= -\frac{\partial \mathbf{B}} {\partial t} \\
\nabla \times \mathbf{B} &= \mu_0\left(\mathbf{J} + \varepsilon_0 \frac{\partial \mathbf{E}} {\partial t} \right)
\end{aligned}
$$

which form the foundation of classical electromagnetism. 

Using inline math you can describe the different parameters that goes into
the Maxwell's equations, e.g. $\varepsilon_0$ is permittivity of free space.
"""

app = dash.Dash(__name__)

app.layout = html.Div(
    [
        dcc.Textarea(
            id="textarea", value=SOME_MARKDOWN, style={"width": "100%", "height": "50vh"}
        ),
        dcc.Markdown(id="user-provided-output"),
    ]
)


@app.callback(
    Output(component_id="user-provided-output", component_property="children"),
    [Input(component_id="textarea", component_property="value")],
)
def update_output_div(input_value):
    return input_value


if __name__ == "__main__":
    app.run_server(debug=True)
```

**Remaining tasks if this is to be taken further:**
- [ ] Move import/load of MathJax to `LazyLoader`.
- [ ] Configure MathJax to not typeset any existing math on page when loaded.
- [ ] Use MathJax event system to ensure that we do not try to render before MathJax is loaded and ready.
- [ ] [Chain MathJax typeset promises](http://docs.mathjax.org/en/latest/web/typeset.html#handling-asynchronous-typesetting)?
- [ ] Investigate if compiling/creating own MathJax component is better?
- [ ] Pass symbols indicating block/inline math from the Dash Markdown component (as discussed earlier, we might now want `$inline_math$` to be the default).
- [ ] Flag on Dash Markdown component to turn rendering of mathematics on/off.

anders-kiaer:
Superseded by https://github.com/plotly/dash/pull/1949. 🎉

Issue #640, opened 2019-09-09
tooltip={'always_visible': True} doesn't work with slider


jonyow:
I get an error when I use the tooltip={'always_visible': True} argument when creating a dash slider in python.

The error only appears when I move the slider.

When I remove the tooltip argument, the code runs fine (but I don't get the tooltip that I need).

```
# Here's a minimal example that reproduces the error

import dash
import dash_html_components as html
import dash_core_components as dcc

app = dash.Dash(__name__) 
app.layout = html.Div([
    dcc.Slider(
        id='my-slider',
        min=0,
        max=20,
        step=0.5,
        value=5,
        tooltip={'always_visible': True}
    ),
    html.Div(id='output-container-slider')
])

@app.callback(
    dash.dependencies.Output('output-container-slider', 'children'),
    [dash.dependencies.Input('my-slider', 'value')])
def update_output(value):
    return 'You have selected "{}"'.format(value)

if __name__ == '__main__':
    app.run_server(debug=True)
```

I would expect the tooltip to show the current value of the slider as it moves and the tooltip to persist even I move the mouse away.

Instead, I get this error:

```
Failed component prop type: Invalid component prop `tooltip` key `visible` supplied to Slider.
Bad object: {
  "visible": true
}
Valid keys: [
  "always_visible",
  "placement"
]

(This error originated from the built-in JavaScript code that runs Dash apps. Click to see the full stack trace or open your browser's console.)
Error: Failed component prop type: Invalid component prop `tooltip` key `visible` supplied to Slider.

Bad object: {
  "visible": true
}

Valid keys: [
  "always_visible",
  "placement"
]

    at propTypeErrorHandler (http://127.0.0.1:8050/_dash-component-suites/dash_renderer/dash_renderer.dev.js?v=1.0.1&m=1568039086:44125:11)

    at CheckedComponent (http://127.0.0.1:8050/_dash-component-suites/dash_renderer/dash_renderer.dev.js?v=1.0.1&m=1568039086:40498:9)

    at renderWithHooks (http://127.0.0.1:8050/_dash-component-suites/dash_renderer/react-dom@16.8.6.js?v=1.0.1&m=1568039086:13073:18)

    at mountIndeterminateComponent (http://127.0.0.1:8050/_dash-component-suites/dash_renderer/react-dom@16.8.6.js?v=1.0.1&m=1568039086:15155:13)

[...]

PR #645, opened 2019-09-16
Fix dash-core-compoents Slider and RangeSlider tooltip

AboorvaDevarajan:
I faced the same issue as #640, with this patch I was able to get it working, and it could be a potential fix.


alexcjohnson:
Thanks @AboorvaDevarajan -but I think the problem is much simpler: we just need to reverse the order of arguments to `Object.assign`, as is it's overwriting the original prop.

AboorvaDevarajan:
@alexcjohnson Yes, that should be simpler. I have updated the PR.

AboorvaDevarajan:
With reference to #648 I will close this. Thanks.

Issue #584, opened 2019-07-17
new Graph props for current state of figure


alexcjohnson:
Currently the way to respond to zoom events, legend toggle events, etc. is to listen to `relayoutData` and `restyleData`. This gets super complicated though since it only contains the latest changes, which may be a partial zoom (single axis, single-ended) or completely unrelated to the information you actually care about. So to make a really robust app, the dev needs to manually accumulate these changes in a `Store` or something.

It would be way better to just expose the current state of the figure as its own prop. Then, to avoid infinite loops, the normal usage pattern would be to use `relayoutData` as an `Input` just to trigger the callback, and use `liveFigure` or whatever we call it as `State`.

But of course we don't want the *entire* figure, we need to at least exclude data arrays as they may be huge and slow things down.

The `uirevision` system collects [attribute patterns](https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_api.js#L2425-L2465) that can be set by the UI - we could presumably piggyback off that, though that's reaching fairly deep into plotly.js unless we build this functionality directly into plotly.js. Alternatively we could clone the whole figure minus data arrays - or even just clip all arrays, though there are a few that we need to keep in their entirety (`hiddenlabels`, `selectedpoints`...)

chriddyp:
Would this change the way `selectedData` would work too? Since `selectedpoints` is part of the figure now?

chriddyp:
Re selections - This API would be quite nice for writing apps with parcats selection. `restyleData` only contains the last selection, rather than the current state of the selected axes:
![image](https://user-images.githubusercontent.com/1280389/74680033-8281ba80-518d-11ea-91d5-999d359cd34b.png)

emmanuelle:
What should be the content of this prop? The whole layout, with a few exceptions (`layout.mapbox.layers` can be huge for example. Any others?). `selectedData` can be large too but it's quite useful (see the parcats example above) so probaby we should keep it? @nicolaskruchten

alexcjohnson:
The key criterion, IMO, is that anything that can potentially be modified via GUI actions must be included, whether it actually was modified or not. So that would absolutely include `selectedpoints` despite its potentially large size.

Beyond that, we could imagine sending back either:
- everything but the trace data arrays and a few large layout items like mapbox layers
- only a whitelisted set of GUI-editable attributes
- something in between that's easier to generate?

[...]

PR #646, opened 2019-09-16
add persistence to relevant components

alexcjohnson:
Gives the following components persistence support (see plotly/dash#903):
`Checklist`, `DatePickerRange`, `DatePickerSingle`, `Dropdown`, `Input`, `RadioItems`, `RangeSlider`, `Slider`, `Tabs`, `Textarea`
All but the date pickers persist only `value`. `DatePickerRange` is the only one that persists more than one prop (`start_date` and `end_date`). We could imagine building a facility into persistence to omit `persisted_props` in this case, since it's pretty much useless to the user... for now I just mentioned this in the docstring.

These components do not have persistence support:
`ConfirmDialog[Provider]`, `Graph`, `Interval`, `Link`, `Loading`, `Location`, `LogoutButton`, `Markdown`, `Store`, `Tab`, `Upload`

I think this is all straightforward, the only one I think might at some point want to get persistence support is `Graph` (selections, perhaps zoom or camera angle?), but this is not going to be simple to add, so I suggest omitting this until there's a clear need.


alexcjohnson:
> just needs checks to be fixed

- [x] https://github.com/plotly/dash-core-components/pull/646/commits/c725103bde1b2220a9bc3c1f0e25b316ff742362

> and a changelog entry

- [x] https://github.com/plotly/dash-core-components/pull/646/commits/4e8aa3276324de0201790e2c05d456c9932381f7

bcliang:
This is a pretty significant improvement for users who have lots of input fields! I've been manually using browser sessionStorage to persist input values. 

Any comments on impl. for `Graph` ?

alexcjohnson:
@bcliang What's "impl."? Implications? Implementation? 😈 

We didn't include `Graph` for now, despite the fact that this feature was somewhat modeled off `uirevision` from plotly.js, since this is going to be rather more involved.

We might be able to do it in this repo after #584 makes a prop reporting user interactions back; but if we want to support the full complexity that `uirevision` allows, it might be better to implement it on the plotly.js side, perhaps with a config attribute that stores these edits in local or session storage rather than in regular memory (attached to the DOM element).

It might be best to make a new issue for this - here, in this repo for now - and explain the use cases you have in mind. Then we can decide how best to handle it.

PR #647, opened 2019-09-17
Update ramda to 0.26.1 and convert deprecated contains -> includes

alexcjohnson:
A quick little bit of maintenance... @wbrgss you can remove some edits in your graph fork after this 😌


wbrgss:
Thanks for doing this @alexcjohnson; I started on it but ran into npm/npm#17749 — but if build files are not being committed I guess it doesn't matter anymore

PR #648, opened 2019-09-17
fix #640 - tooltip.always_visible for slider and rangeslider

### LINKED ISSUE #640: tooltip={'always_visible': True} doesn't work with slider

alexcjohnson:
Thanks @AboorvaDevarajan for kicking this off - I also realized we don't really want to use `Object.assign` anyway, we're still trying to maintain IE11 compatibility. I switched it to `assoc` from `ramda`, and added a test for each component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed `Slider` and `RangeSlider` components with `tooltip.always_visible` [#640](https://github.com/plotly/dash-core-components/issues/640)
+
 - Fixed an infinite loop problem when `Graph` is wrapped by `Loading` component [#608](https://github.com/plotly/dash-core-components/issues/608)
 
 ## [1.1.2] - 2019-08-27
--- a/src/components/RangeSlider.react.js
+++ b/src/components/RangeSlider.react.js
@@ -1,6 +1,6 @@
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
-import {omit} from 'ramda';
+import {assoc, omit} from 'ramda';
 import {Range, createSliderWithTooltip} from 'rc-slider';
 
 /**
@@ -52,9 +52,7 @@ export default class RangeSlider extends Component {
              * the rc-tooltip API uses `visible`, but `always_visible is more semantic
              * assigns the new (renamed) key to the old key and deletes the old key
              */
-            tipProps = Object.assign(tooltip, {
-                visible: tooltip.always_visible,
-            });
+            tipProps = assoc('visible', tooltip.always_visible, tooltip);
             delete tipProps.always_visible;
         } else {
             tipProps = tooltip;
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -1,7 +1,7 @@
 import React, {Component} from 'react';
 import ReactSlider, {createSliderWithTooltip} from 'rc-slider';
 import PropTypes from 'prop-types';
-import {omit} from 'ramda';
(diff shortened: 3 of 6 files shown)


alexcjohnson:
@byronz there's still something weird going on with cached builds; first commit `test-37` failed being unable to find `persistence` as a prop (so it's got a build left over from before that prop existed, but running against tests added with that prop??), but `test-27` and `test-36` passed. Second time around it was `test-36` that failed, the others passed, but it passed again on rerun.

PR #657, opened 2019-09-24
655 datepickersingle

### LINKED ISSUE #655: add test coverage for DatePickerSingle
after investigation, issue #652 reported in Dash 1.3.0 regression was not detected by CI jobs. the root cause is that we don't have any integration test to cover this component. 

we should cover with the app provided in #652 and also align with the existing tests for `DatePickerRange`

byronz:
this closes #655


tests/integration/utils.py
```diff
@@ -1,4 +1,38 @@
 from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+
+DATE_PICKER_DAY_SELECTOR = 'div[data-visible="true"] td.CalendarDay'
+
+
+def choose_calendar_date(dd, comp_id, index=0, day="", outside_month=False):
```
byronz: @alexcjohnson  I saw your comment about the scope of the calendar API should stay in DCC, but I would prefer to move this in `dash.testing` as it's a very common action we might use elsewhere and requires some tricks to make it work more stable.  another reason I don't like the current utils.py solution is that it requires the extra `__init__.py` be present in the test folders, which makes the test code not clean in a sense that you can run the tests only in the root folder for example.

alexcjohnson: Agreed that adding `__init__.py` is unpleasant, as is the need to explicitly pass in `dash_duo` to the function. But I think most of what happens here is quite specific to DCC and becomes substantially harder to maintain - 2 PRs with coupled versions if the datepicker changes, for example, and if anyone else DOES depend on this API we'd have to find a way to maintain compatibility.

Is there a way to bake this into a DCC-specific version of `dash_duo`, like via `conftest` or something?

> requires some tricks to make it work more stable

If there are *pieces* of this that are transferrable to other scenarios independent of the specific element structure, selectors, etc of DCC, I'll be more than happy to include those in `dash.testing`.

byronz: yes, we can basically have a dcc fixture like we did in dash-docs, and these `utils` can go to a mixin class like https://github.com/plotly/dash/blob/dev/dash/testing/dash_page.py,  and the fixture will be defined in a `conftest.py` file. this solution would allow the selector to be stayed in dcc, and the abstraction can ship to dash repo.

tests/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+
+        # start server with app and pass Dash arguments
+        self.server(app, **kwargs)
+
+        # set the default server_url, it implicitly call wait_for_page
+        self.server_url = self.server.url
+
+
+@pytest.fixture
+def dash_dcc(request, dash_thread_server, tmpdir):
```
byronz: @alexcjohnson  all the dcc utils will be defined in `dash_core_components_page.py`

tests/integration/misc/test_persistence.py
```diff
@@ -137,44 +118,44 @@ def make_output(*args):
...
+    dash_dcc.wait_for_text_to_equal("#settings", json.dumps(initial_settings))
 
-    dash_duo.find_element("#checklist label:last-child input").click()  # 🚀
+    dash_dcc.find_element("#checklist label:last-child input").click()  # 🚀
 
-    dash_duo.find_element("#start_date").click()
-    click_date(dash_duo, "#datepickerrange", 0, 6)  # 2019-05-04
-    dash_duo.find_element("#end_date").click()
-    click_date(dash_duo, "#datepickerrange", 2, 2)  # 2019-05-14
+    dash_dcc.select_calendar_date(
```
byronz: use the new API which specifies the day instead of the position tuple

tests/dash_core_components_page.py
```diff
@@ -0,0 +1,64 @@
+import logging
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+
+logger = logging.getLogger(__name__)
+
+
+class DashCoreComponentsMixin(object):
+    def select_calendar_date(
```
byronz: @alexcjohnson  the selector is now a property, would this be generic enough to be ported?

alexcjohnson: Personally I'd prefer to still leave it in this repo. If we find another use for it outside DCC, *and* it still works in the form here, it should be easy to move over at that point; but it feels quite likely to me that it's still tightly coupled to this particular datepicker implementation - the start/end stuff, the `__outside`...

Regardless, looks great this way :)

tests/integration/calendar/test_calendar_props.py
```diff
@@ -7,40 +7,55 @@
 from dash.dependencies import Input, Output
 import dash.testing.wait as wait
 
-DAY_SELECTOR = 'div[data-visible="true"] td.CalendarDay'
+DATE_PICKER_DAY_SELECTOR = 'div[data-visible="true"] td.CalendarDay'
```
alexcjohnson: `dcc.date_picker_day_locator`?

byronz: good catch

package.json
```diff
@@ -18,7 +18,7 @@
     "test": "run-s -c lint lint:py format:test test-unit test:legacy test:intg test:pyimport",
     "test:legacy": "pytest tests/test_integration*.py",
     "test:intg": "pytest --nopercyfinalize --headless tests/integration",
-    "test:pyimport": "python -m unittest tests/test_dash_import.py",
+    "test:pyimport": "echo $(python -V 2>&1) | grep 3. && python -m unittest tests/test_dash_import.py",
```
alexcjohnson: Worries me a little that this is completely silent about skipping this test - can we have it show a message about what it did at least? something like `|| echo "skipping test_dash_import in Python 2.x"`?

alexcjohnson: oh you moved this into the `circle.yml` - but I think the same comment applies there.

byronz: yes, the NPM version does not work. I remember this test was ignored for quite a long time, and I think the tests in Python3 should be enough to cover this https://github.com/plotly/dash-core-components/blob/474f196d285e83cbe750bbbab998b220a0d61ee2/dash_core_components/__init__.py#L18-L23

PR #659, opened 2019-09-25
Remove artifacts from `dev`

Marc-Andre-Rivet:
Same as dash-table https://github.com/plotly/dash-table/pull/603


.circleci/config.yml
```diff
@@ -1,240 +1,240 @@
 version: 2
 jobs:
-  "percy-finalize":
-    docker:
-      - image: percyio/agent
-    steps:
-      - run: percy finalize --all
+    'percy-finalize':
```
byronz: does this file just get `prettier`? hard to tell from the diff the real change

Marc-Andre-Rivet: fixed

.circleci/config.yml
```diff
@@ -77,10 +77,10 @@ jobs:
       - run:
           name: 🏗️ build dash
           command: |
-            . venv/bin/activate && pip install --upgrade -e . --quiet && mkdir packages
+            . venv/bin/activate && mkdir packages
             # build main dash
             git clone --depth 1 https://github.com/plotly/dash.git dash-main
-            cd dash-main && pip install -e .[dev,testing] && python setup.py sdist && mv dist/* ../packages/
+            cd dash-main && pip install -e .[dev,testing] --quiet && python setup.py sdist && mv dist/* ../packages/
```
byronz: I think our intention is to mute the progress bar,  ` --progress-bar off`
at least for now, we need to check the pip install logs from time to time

Marc-Andre-Rivet: https://github.com/plotly/dash-core-components/pull/659/commits/56093b14ebd5ca7170da85b61fa3cf781e686c8c

PR #660, opened 2019-09-26
Add search_value property to Dropdown

inytar:
This PR adds the `search_value` property to the `Dropdown` component. Using the `search_value` you can do server side filtering of the `options` as shown in the following example:

```python
import dash

options=[
    {'label': 'New York City', 'value': 'NYC'},
    {'label': 'Montreal', 'value': 'MTL'},
    {'label': 'San Francisco', 'value': 'SF'}
]

app.layout = dcc.Dropdown(id='my-dynamic-dropdown')


@app.callback(
    dash.dependencies.Output('my-dynamic-dropdown', 'options'),
    [dash.dependencies.Input('my-dynamic-dropdown', 'search_value')],
)
def update_options(search_value):
    return [o for o in options if search_value in o['label']]


if __name__ == '__main__':
    app.run_server(debug=True)
```

This example will also be added to [dash-cocs](https://github.com/plotly/dash-docs).

An example of a use-case of this feature is described in [this  forum topic](https://community.plot.ly/t/dynamic-options-for-drop-down/7550).


inytar:
@alexcjohnson thanks for the review. I'll remove the artifacts.
About the test, I started writing it and then got stuck, I'm not really sure how I would test this. I'll look in the integration directory, see if I can find an example.

alexcjohnson:
A good example is [`test_state_and_inputs` in the `Input` tests](https://github.com/plotly/dash-core-components/blob/474f196d285e83cbe750bbbab998b220a0d61ee2/tests/integration/input/test_input_and_state.py#L10) - it shows how to start an app and send keypresses to the right component. Then you'll just have to figure out a selector that returns the option list and do something like:
```py
wait.until(lambda: len(dash_duo.find_elements('<option selector>')) == <expected number>)
assert dash_duo.find_element('<option selector>').text == <expected option>
```

tests/integration/dropdown/test_search_value.py
```diff
@@ -0,0 +1,36 @@
...
+        Output("output", "children"), inputs=[Input("dropdown", "search_value")]
+    )
+    def update_output(search_value):
+        call_count.value += 1
+        return 'search_value="{}"'.format(search_value)
+
+    dash_duo.start_server(app)
+
+    # Get the inner input used for search value.
+    input_ = dash_duo.find_element("#dropdown").find_element_by_tag_name("input")
```
byronz:
```suggestion
input_ = dash_duo.find_element("#dropdown input")
```

tests/integration/dropdown/test_search_value.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # callback gets called with initial input
+    wait.until(lambda: call_count.value == 1, timeout=1)
+
+    assert dash_duo.wait_for_text_to_equal("#output", 'search_value="something"')
```
byronz: all the `wait_for_*` APIs are some sort of assertions, so the `assert` here is not needed

tests/integration/dropdown/test_search_value.py
```diff
@@ -0,0 +1,36 @@
...
+
+    input_.send_keys("x")
+    wait.until(lambda: call_count.value == 2, timeout=1)
+    assert dash_duo.wait_for_text_to_equal("#output", 'search_value="x"')
```
byronz: same as above

tests/integration/dropdown/test_search_value.py
```diff
@@ -0,0 +1,36 @@
+from multiprocessing import Value
+import dash
+from dash.dependencies import Input, Output, State
+import dash.testing.wait as wait
+import dash_core_components as dcc
+import dash_html_components as html
+
+
+def test_search_value(dash_duo):
```
byronz:
```suggestion
def test_ddsv001_search_value(dash_duo):
```

tests/integration/dropdown/test_search_value.py
alexcjohnson: 😈  who's the @byronz who wrote this line (in the test I pointed @inytar to as an example) that does exactly the same https://github.com/plotly/dash-core-components/blame/474f196d285e83cbe750bbbab998b220a0d61ee2/tests/integration/input/test_input_and_state.py#L38

byronz: 🦈  
that's why I didn't `suggest` but just points out it with a guilty conscience

byronz: must be 🍻

tests/integration/dropdown/test_search_value.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+    call_count = Value("i", 0)
+
+    @app.callback(
+        Output("output", "children"), inputs=[Input("dropdown", "search_value")]
+    )
+    def update_output(search_value):
+        call_count.value += 1
+        return 'search_value="{}"'.format(search_value)
```
alexcjohnson: Ah sorry, I missed this earlier - I think the test should use the same structure as the example app you gave, which is updating `options` from `search_value` in the same component. The reason this is important is (a) that's the primary use case for this prop, and (b) it will cause the dropdown to rerender, and we need to ensure that when it does the correct options are visible. I can certainly imagine errors in the component where on render it loses focus, or for some other reason it stops displaying the option list.

(BTW nota big deal but you can probably get rid of all the `call_count` stuff, that doesn't really matter for this test, we only care about what's going on within the browser)

inytar:
@alexcjohnson I miss read your first comment and thought you just wanted a basic integration test. I've kept this test because I think it's good to keep a basic test that is easier to debug/understand if something fails. I've also added a test that updates the options dynamically from the server, I hope this is what you meant.

tests/integration/dropdown/test_dynamic_options.py
```diff
@@ -0,0 +1,54 @@
+import dash
+from dash.dependencies import Input, Output, State
+from dash.exceptions import PreventUpdate
+import dash.testing.wait as wait
+import dash_core_components as dcc
+
+
+def test_ddsv001_dynamic_options(dash_duo):
```
byronz: sorry I was not clear about the suggestion. the TCID added in the previous test should be unique. more details refer to Notes #2 in example. https://dash.plot.ly/testing

this one should be sth like `dddo001`

tests/integration/dropdown/test_dynamic_options.py
```diff
@@ -0,0 +1,54 @@
...
+    input_ = dropdown.find_element_by_css_selector("input")
+
+    # Focus on the input to open the options menu
+    input_.send_keys("x")
+
+    #No options to be found with `x` in them, should show the empty message.
+    dash_duo.wait_for_text_to_equal(".Select-noresults", "No results found")
+
+    input_.clear()
+    input_.send_keys("o")
```
byronz: should we try some more edge cases with values like punctuation, Unicode etc?

alexcjohnson: I think this is fine for now - if problems come up we can test them in a bugfix PR. @inytar this test looks great, thanks!

alexcjohnson:
Oh, the linter had some complaints as well:
```
tests/integration/dropdown/test_dynamic_options.py:9:12: E225 missing whitespace around operator
tests/integration/dropdown/test_dynamic_options.py:36:5: E265 block comment should start with '# '
```
It looks like js linting also complained but that somehow doesn't cause the test run to fail... so don't worry about that, we'll get that separately.

tests/integration/dropdown/test_dynamic_options.py
inytar: I can do that in a next PR (in our use-case we have emojis working without any problems).

inytar:
Thanks for looking at this and the fast reviews! I enjoyed the process and hope I get to work with plotly dash more in the future.

inytar:
Btw, [this](https://github.com/plotly/dash-docs/pull/656) is the related example for the documentation. I'd be very happy if this can also be added.

alexcjohnson:
oops I forgot to ask for a changelog entry - added here: https://github.com/plotly/dash-core-components/commit/d750da6c41263eebc1ae33cca625c552fa07a3d4

PR #662, opened 2019-09-26
:alembic: test exit status

byronz:
add `set -eo pipefail` in steps where multiples tests are done to avoid the false signal of green light based only on last test result.


alexcjohnson:
Strange, I wonder where these subpixel Chinese character Percy diffs keep coming from...
Screen Shot 2019-09-26 at 9 26 26 PM

byronz:
That's simply a Percy bug. Had trouble with Chinese characters

PR #687, opened 2019-10-28
Use start date, if one is supplied, for initial_visible_month.

### LINKED ISSUE #679: DateRangePicker: unable to select when initial_visible_month is not given
With 
```
from datetime import datetime as dt

dcc.DatePickerRange(
    id='date-picker-range',
    start_date = dt(2017, 1, 31),
    end_date = dt(2018, 3, 21),
    min_date_allowed = dt(2017, 1, 31),
    max_date_allowed = dt(2018, 3, 21),
)
```
the component is not responsive (nothing can be selected), whereas with 
```
from datetime import datetime as dt

dcc.DatePickerRange(
    id='date-picker-range',
    start_date = dt(2017, 1, 31),
    end_date = dt(2018, 3, 21),
    min_date_allowed = dt(2017, 1, 31),
    max_date_allowed = dt(2018, 3, 21),
    initial_visible_month=dt(2017, 1, 31)
)
```
it's OK. If the `start_date` is given but not `initial_visible_month`, then `initial_visible_month` could default to `start_date`.

shammamah:
Closes #679.


src/utils/convertToMoment.js
```diff
@@ -11,7 +11,9 @@ export default (newProps, momentProps) => {
             dest[key] = null;
 
             if (key === 'initial_visible_month') {
-                dest[key] = moment();
+                dest[key] = newProps['start_date'] || newProps['min_date_allowed'] ?
```
Marc-Andre-Rivet: From the PR:
> the component is not responsive (nothing can be selected)

The selection is not disable but as the initial month used to open the menu is after the `end_date` the dates are disabled -- moving back to a month before the `end_date` enables the menu. That said, this is a technicality, the behavior is unexpected and annoying especially if the date range is far in the future / past.

With the change here, the default behavior can also be annoying if `end_date` or `max_date_allowed` is set because they are not being used to decide the initial month displayed.

I think we could go further and also use `end_date` and `max_date_allowed` in determining the `initial_visible_month` if `start_date` and `min_date_allowed` are not available.

With
```
from datetime import datetime as dt

import dash

app = dash.Dash('app',
    suppress_callback_exceptions=True)

app.layout = dcc.DatePickerRange(
    id='date-picker-range',
    # start_date = dt(2017, 1, 31),
    end_date = dt(2018, 3, 21),
    # min_date_allowed = dt(2017, 1, 31),
    max_date_allowed = dt(2018, 3, 21),
)

if __name__ == "__main__":
    app.run_server(debug=True)
```

image

shammamah: So far I have `start_date` -> `min_date_allowed`, and it makes the most sense to me to then look for `end_date` and `max_date_allowed` in that order if neither of those values are supplied. Do you think this is the best order to use?

Marc-Andre-Rivet: Yup. Makes sense to me :) low current -> lowest -> high current -> highest

src/utils/convertToMoment.js
```diff
@@ -11,7 +11,11 @@ export default (newProps, momentProps) => {
             dest[key] = null;
 
             if (key === 'initial_visible_month') {
-                dest[key] = moment();
+                dest[key] = newProps['start_date'] || newProps['min_date_allowed']
+		    || newProps['end_date'] || newProps['max_date_allowed'] ?
+		    moment(newProps['start_date'] || newProps['min_date_allowed']
+			  || newProps['end_date'] || newProps['max_date_allowed']) :
+		    moment();
```
Marc-Andre-Rivet: Since `moment()` and `moment(undefined)` are equivalent (but not `moment(null)` 🙄), this can be simplified to: 

```
dest[key] = moment(newProps['start_date'] || newProps['min_date_allowed'] || newProps['end_date'] || newProps['max_date_allowed'] || undefined);
```

Marc-Andre-Rivet:
@shammamah One small suggested code change and `npm run format` to be run for the linter to stop complaining :)

PR #692, opened 2019-11-01
Async components

Marc-Andre-Rivet:
Follow up to https://github.com/plotly/dash-core-components/pull/616

Brings async to components:
- datepicker single/range
- dropdown
- markdown
- upload

Slider/RangeSlider behave incorrectly when asynced -- they seem to execute just fine but they display nothing. Needs to be investigated further, and there is diminishing returns.

Brings `dcc.min.js` entry point down from ~1.09MB to ~374kB.

None of these components have side-effects, so they do not use the async decorator.

- [x] changelog
- [x] fix percy diffs for datepickers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Added
+- [#692](https://github.com/plotly/dash-core-components/pull/692) Async DatePickerSingle, DatePickerRange, Dropdown, Markdown, Upload components
+
 ## [1.4.0] - 2019-10-29
 ### Added
 - [#616](https://github.com/plotly/dash-core-components/pull/616) Async Graph and Plotly.js
--- a/src/fragments/DatePickerRange.react.js
+++ b/src/fragments/DatePickerRange.react.js
@@ -19,14 +19,14 @@ export default class DatePickerRange extends Component {
         };
     }
 
-    propsToState(newProps) {
+    propsToState(newProps, force = false) {
         const state = {};
 
-        if (newProps.start_date !== this.props.start_date) {
+        if (force || newProps.start_date !== this.props.start_date) {
             state.start_date = newProps.start_date;
         }
 
-        if (newProps.end_date !== this.props.end_date) {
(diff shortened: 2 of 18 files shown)


src/fragments/DatePickerRange.react.js
```diff
@@ -0,0 +1,194 @@
...
+            this.setState(state);
+        }
+    }
+
+    componentWillReceiveProps(newProps) {
+        this.propsToState(newProps);
+    }
+
+    componentWillMount() {
+        this.propsToState(this.props, true);
```
Marc-Andre-Rivet: This is the most significant code change. Since the real/fragment component is wrapped, its props lifecycle is slightly different, `propsToState` is updated to only update the state when the value changes or if forced (on mount)

alexcjohnson: I'm missing something - this seems like it would have been a good idea before, but somehow more than a good idea it's actually *necessary* now? What's the bad thing that happens if we don't do this?

alexcjohnson (approved):
https://github.com/plotly/dash-core-components/pull/692#discussion_r342170034 is just for information, no change needed. This looks great. 💃

src/fragments/DatePickerRange.react.js
Marc-Andre-Rivet: props override state, always - the component essentially doesn't work at all anymore

Marc-Andre-Rivet:
Will wait on https://github.com/plotly/dash-core-components/pull/693 for the test fix as this alone will trigger an incorrect baseline b/c of November calendar 👽

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- [#692](https://github.com/plotly/dash-core-components/pull/692) Async DatePickerSingle, DatePickerRange, Dropdown, Markdown, Upload components
```
Rigobertomarro, line 7: Ñ se corta mi aser

Rigobertomarro, line 7: Quiero aprender

Rigobertomarro (changes requested):

PR #695, opened 2019-11-07
Omit marks that are outside of range specified by min and max.

### LINKED ISSUE #517: RangeSlider: out-of-range labels
if you put labels outside the slider range, they are still drawn, but they shouldn’t be
see here FYI https://github.com/plotly/dashR/issues/80#issue-430575656

### Example:
```
dccRangeSlider(
    marks=as.list(setNames(c(paste("Label", -5:6)), -5:6)),
    min=-5,
    max=6,
    value=list(-3, 4)
)
```

Screen Shot 2019-04-10 at 10 28 23 AM

shammamah:
Closes #517.


shammamah:
@alexcjohnson Right now I'm just omitting all labels that fall outside of the range specified for the slider, but I also agree with this comment you made:
> I feel like there's a use case for "here are all the possible labels" and sometimes restricting the range to a subset of that without needing to also filter the labels.

Do you feel it would be a good idea to include some new property to "disable" the options for the ones outside of a specific range (but still show them)?

alexcjohnson:
> Do you feel it would be a good idea to include some new property to "disable" the options for the ones outside of a specific range (but still show them)?

No, I think what you did here is exactly what we want, no new options needed 🎉

src/components/RangeSlider.react.js
```diff
@@ -58,6 +58,17 @@ export default class RangeSlider extends Component {
...
+        const omittedMarks = this.props.marks
+            ? Object.keys(this.props.marks).filter(value => {
+                  return value < this.props.min || value > this.props.max;
+              })
+            : [];
+
+        const truncatedMarks = omit(
+            omittedMarks.map(mark => parseInt(mark, 10)),
+            this.props.marks
+        );
```
alexcjohnson: `parseInt` seems like a problem here - these might not be integers. But anyway looks like you can do this in one step using https://ramdajs.com/docs/#pickBy

src/components/RangeSlider.react.js
```diff
@@ -82,8 +93,15 @@ export default class RangeSlider extends Component {
...
+                    marks={truncatedMarks}
                     {...omit(
-                        ['className', 'value', 'setProps', 'updatemode'],
+                        [
+                            'className',
+                            'value',
+                            'setProps',
+                            'marks',
+                            'updatemode',
+                        ],
```
alexcjohnson: Not for now, but just because I'm thinking about it: I'd like us to consider moving away from `omit` for forwarding props to 3rd-party components, and use `pick` instead. IMO `omit` makes it too hard to follow which props are being passed on, too easy to forget to remove newly added props, and makes us lazy about accepting someone else's choice of names and values even if they don't make sense in our case. (cc @Marc-Andre-Rivet )

tests/integration/sliders/test_sliders.py
```diff
@@ -58,3 +58,19 @@ def update_output(rng):
...
+
+
+def test_slsl003_out_of_range_marks_rangeslider(dash_dcc):
+
+    app = dash.Dash(__name__)
+    app.layout = html.Div([
+        dcc.RangeSlider(
+            min=0,
+            max=5,
+            marks={i: 'Label {}'.format(i) for i in range(10)}
```
alexcjohnson: include one before `min` too?
💯 for including marks exactly at `min` and `max, to verify that both are included 🎉

alexcjohnson:
Two more questions:
- `Slider` too, in addition to `RangeSlider`?
- Is there anything we can do about the labels overflowing the bounding box of the `<div>` we draw the slider in?

shammamah:
>`Slider` too, in addition to `RangeSlider`?
Yes, this makes sense. 

>Is there anything we can do about the labels overflowing the bounding box of the <div> we draw the slider in?
I'll take a look at the CSS!

shammamah:
@alexcjohnson Made some changes to the padding and now it looks good! 

Before: 
Screen Shot 2019-11-12 at 2 32 49 PM


After:
Screen Shot 2019-11-12 at 2 32 57 PM

shammamah:
Some other changes made (cc @wbrgss since this affects how the slider CSS looks -- would love any suggestions you have): 

- The slider would just have a height of `0px` if `vertical` was set to `True`, unless a height was specified in a stylesheet. This has now changed, as I've added a `verticalHeight` prop that defaults to `400px`. (This won't affect any existing apps, since this prop will be overridden by a custom stylesheet.) 
- Whitespace is now preserved in the mark text, to allow for users to enter line breaks in the labels if they start to overlap each other.
- The padding is now optional and based on tooltips. 
    - If the tooltip is always visible and on the top in a horizontal slider, the padding at the top is `25px`; otherwise it is `0px`. 
    - If the tooltip is always visible and on the left in a vertical slider, the padding for the left side is `25px`; otherwise it is `0px`. 
    - Padding for the bottom and the right are always `25px`, since this is where the labels are.

wbrgss:
@shammamah These look like good changes to me. In DDK, we're "[documenting around](https://dash-gallery.plotly.host/dash-design-kit/pro-tips)" (private docs, see "dcc.Slider best practices" section) some of these issues, including vertical height:

> - When setting the marks parameter for a dcc.Slider, ensure that min <= marks[0] and max >= marks[-1] and that your range of marks matches the min/max range. Otherwise, your marks may overflow their container.
> - Make sure your marks are spaced correctly such that they don't overlap each other.
> - On vertical sliders (dcc.Slider(vertical=True, ...)), you will probably have to use the height parameter on your control to ensure your marks and labels have enough room. In other controls and components, the height parameter should usually be avoided...

>  I've added a verticalHeight prop that defaults to 400px
This seems like a good idea to me, and so does 
> Padding for the bottom and the right are always 25px, since this is where the labels are.
I think that will be adequate to prevent this situation:

image

wbrgss:
@shammamah What about labels that overlap horizontally, because there are too many of them to fit?
image

If this PR doesn't handle it, handling this robustly, by e.g. staggering the labels, might be complicated so probably out of scope for this PR. But if you think it's a good idea I can make an issue for it. Maybe we should at least document it to warn people if they are making "responsive" sliders that might have squishy labels if their screen width is too small.

shammamah:
@wbrgss The best way to handle this now will have to do with the change I made with c8070d4 -- users will have to supply their own linebreaks. I think the ability to rely on Dash to add in the line breaks in a "smart" fashion (i.e., automatically) might be more suited to DDK. What do you think?

src/components/RangeSlider.react.js
Marc-Andre-Rivet: @alexcjohnson Agreed for DCC! There's also a bit of this happening in the table fragments / d3-format wrappers that should be looked into. Due to its generated nature I'd say whether this makes sense on HTML components would have to be evaluated independently. Can you open an issue for follow up and tag it for `Dash v1.x` milestone? Thanks.

src/components/RangeSlider.react.js
```diff
@@ -58,14 +59,51 @@ export default class RangeSlider extends Component {
             tipProps = tooltip;
         }
 
+        const truncatedMarks = this.props.marks
+            ? pickBy(
+                  (k, mark) => mark >= this.props.min && mark <= this.props.max,
+                  this.props.marks
+              )
+            : this.props.marks;
```
Marc-Andre-Rivet: A matter of personal preference, but you can also make this more compact with
```
const truncatedMarks = this.props.marks && pickBy(...);
```

JS does not return a boolean for `&&`, `||`, etc. but rather the last evaluated value.

src/components/RangeSlider.react.js
```diff
@@ -58,14 +59,51 @@ export default class RangeSlider extends Component {
...
+                    'topRight',
+                    'bottomRight',
+                ]))
+        ) {
+            style = merge(style, {paddingLeft: '0px'});
+        }
+
+        if (vertical) {
+            style = merge(style, {height: verticalHeight + 'px'});
+        }
```
Marc-Andre-Rivet: * Could be simplified with shared vertical/!vertical checks.
* Minor point, could be memoized `(style, vertical, tooltip) -> style` to prevent style re-evaluation below

src/components/Slider.react.js
```diff
@@ -58,14 +59,51 @@ export default class Slider extends Component {
...
+                    'topRight',
+                    'bottomRight',
+                ]))
+        ) {
+            style = merge(style, {paddingLeft: '0px'});
+        }
+
+        if (vertical) {
+            style = merge(style, {height: verticalHeight + 'px'});
+        }
```
Marc-Andre-Rivet: Same as above

src/utils/computeSliderStyle.js
```diff
@@ -0,0 +1,35 @@
+import {merge, memoizeWith, identity, contains} from 'ramda';
+
+export default () => {
+    return memoizeWith(identity, (vertical, verticalHeight, tooltip) => {
+        let style = {
+            padding: '25px',
+        };
+
+        if (vertical) {
+            style = merge(style, {height: verticalHeight + 'px'});
```
Marc-Andre-Rivet: `style` is created inside the function scope, it's fine to consider it as mutable inside the function

PR #699, opened 2019-11-11
Update CSS for DatePickerRange vertical arrows.

### LINKED ISSUE #553: DatePicker vertical component out of bound
Screen Shot 2019-05-10 at 11 05 37 AM

It seems like the vertical selection of the date picker component is somewhat out of bound [(link)](https://dash.plot.ly/dash-core-components/datepickerrange). I think it's because of the "bottom:0" in the class DayPickerNavigation__verticalDefault. I fixed it manually by adjusting the bottom and the margin but not sure if it's supposed to be like this . 
Screen Shot 2019-05-10 at 11 14 40 AM

shammamah:
Closes #553.

## Before 
image

## After
image


alexcjohnson:
Can you include before & after screenshots?

shammamah:
@alexcjohnson Added to the initial comment for this PR. I realized while taking the screenshots that this is only for the default `day_size` (`39`) of the `DatePickerRange` component. The SVG up and down arrows have a hard-coded height, so I'm not sure it's possible to make it depend on the value for `day_size`. I'll have to check.

src/components/css/react-dates@20.1.0-fix.css
```diff
@@ -1,3 +1,13 @@
...
     box-sizing: border-box;
-}
\ No newline at end of file
+}
+
+.DayPickerNavigation__verticalDefault {
+    text-align: center;
+    height: initial;
+    padding: 10px 0px;
+}
```
Marc-Andre-Rivet: Adding `box-sizing: border-box;` will take the border width into account when calculating the `50%` size for the buttons, resulting in arrows being side by side.

image

Marc-Andre-Rivet:
This component's styling is a mess 😨 Let's just make sure the arrows make sense for now. Please open a follow up issue in `Dash v1.x` to investigate why our styling is off vs. what we're seeing in the airbnb example page.

PR #706, opened 2019-11-21
dcc.Graph: Improve responsiveness, expose resize status, allow wrapper extension

### LINKED ISSUE #605: [BUG] Graph(style='height=100%') fails when reducing window size on Firefox
**PIP**

```
dash (1.0.2)
dash-bootstrap-components (0.7.0)
dash-core-components (1.0.0)
dash-daq (0.1.7)
dash-html-components (1.0.0)
dash-renderer (1.0.0)
dash-table (4.0.2)
```
**Describe the bug**

Writing: `Graph(..., style=dict(height='100%'))`

**Expected behavior**

The graph should fill the full height of the `div`, which will obey normal CSS rules, that is, span all the available space in the layout. It should not cause any overflow, and respond to window resize (also on Firefox ;-).

**Actual behavior**

* A scrollbar appears for a small amount of overflow (both Chrome and Firefox). Since the amount is small, it is possible to work around this with `overflow=hidden`, but it may cause clipping (e.g. if using long labels for the X axis, etc.).

* Reducing window size does not cause graph resize (Firefox only). That is strange, one would expect that any resize event is handled the same way, regardless if the new size is higher or lower than the original size.

**Edit**

The following example _almost_ works, except for the two issues above. This threw me off, originally I thought that the feature does not work at all. I edited the issue title and content accordingly.

Marc-Andre-Rivet:
Fixes #605 

**Context**
(1) Graph resize
Plotly.js provides graph resize on `window.resize` event but does not provide graph resize on parent element resize. This is consistent with other similar libraries.

Graph resize is affected by many props on the graph figure and config, namely:
- `figure.layout.autosize`, `figure.layout.height`, `figure.layout.width`
- `config.responsive`

The overall behavior is complex and can be summarized by these checks:
https://github.com/plotly/dash-core-components/blob/605-graph-resize/tests/integration/graph/test_graph_responsive.py#L101
https://github.com/plotly/dash-core-components/blob/605-graph-resize/tests/integration/graph/test_graph_responsive.py#L107

(2) Extendable figure/config processing in wrapped component
In certain use cases, the dcc.Graph is wrapped inside another component and provides a service rather than being the desired component itself. In these situations the wrapper component may want to override the graph's figure/config processing with logic of its own.

(3) Graph resizing / loading / asyncing... not ready in general
In certain use cases, the application requires the dcc.Graph to be fully loaded, renderered, and resized before some other operation is made (e.g. QA / testing).

---

**Proposed Changes**
1. Extend the resize behavior of the Plotly.js graph in Dash so as to apply it on parent resize that is not triggered by a window.resize event
2. Allow users of Dash to override the behavior of the graph by passing a new prop `responsive` that will override the graph responsiveness. `responsive` allows 3 values: True, False, 'auto' (default) that will (True) force the graph to be responsive, (False) force the graph to be non-responsive, ('auto') use the graph's default behavior **and** make the graph responsive on parent resize if applicable
3. Add two `function` props: `_dashprivate_transformFigure` and `_dashprivate_transformConfig` that will override default processing of `figure` and `config` (https://github.com/plotly/dash-core-components/blob/605-graph-resize/src/fragments/Graph.react.js#L142)
4. The graph will re-evaluate the figure/config when `_dashprivate_transformFigure` or `_dashprivate_transformConfig` is updated (e.g. generate a new function when the params change)
5. The component is responsible for making sure that the functions at 3 and 4 respect the immutability of figure and config. These values should be cloned and the modified clone returned by the function.
6. Expose the graph's state by adding/removing a `dash-graph--pending` class to the wrapper div of the component

---

** Changed unrelated to responsiveness **
1. While updating the Plotly.js version, used the opportunity to clean up the build process in regardsto the plotly.min.js dependency

---

**Side-effects**
1. The addition of the resize observer div requires the graph div to be wrapped up with the resize observer div inside of another div for styling reasons. To preserve backward compatibility with previous selectors, the wrapper div will in addition of having the class `dash-graph`, have `js-plotly-plot`, the class added by Plotly.js on graph divs.

---

**Dependencies and other considerations**
1. Requires https://github.com/plotly/plotly.js/pull/4392 to be released
2. It was discussed and decided outside the scope of this PR discussion that graph resizing on parent element resize should not be handled by Plotly.js -- related historical items https://github.com/plotly/plotly.js/issues/3984, https://github.com/plotly/plotly.js/pull/2974
3. Testing for this feature is done by testing all combinations of (responsive, autoresize, height/width, dash responsive override) and making sure that the initial and resized graph have the expected size
4. The solution must work in IE11

---

**Todo**
- [x] Unify graph state for (1) async, (2) 1st render, (3) resize -- for example `dash-graph-pending` flag / class -- need to validate the class gets added prior to the `resize` starting, there might be a need for deferring `resize` inside a `setTimeout(..., 0)`
- [x] IE11 support? Seems to be 👌, changing `flex` into hard-coded heights and percentages, the graphs are responsive and all.
- [x] docs (https://github.com/plotly/dash-docs/issues/720) -- tagged for `v1.8` also

---

Use variation of https://github.com/plotly/dash-core-components/blob/605-graph-resize/tests/integration/graph/test_graph_responsive.py#L23 to manually test scenarios.


src/components/Graph.react.js
```diff
@@ -108,6 +108,20 @@ PlotlyGraph.propTypes = {
...
+     * `config.responsive` to True, `figure.layout.autosize` to True and unsetting
+     * `figure.layout.height` and `figure.layout.width`.
+     * If False (default), the Plotly.js plot may exhibit certain responsive
+     * behaviors on its own but nothing is done in Dash to help this behavior.
+     * This is the legacy behavior of the Graph component.
+     * If 'auto', the Graph will determine if the Plotly.js plot can be made fully
+     * responsive (True) or not (False) based on the values in `config.responsive`,
+     * `figure.layout.autosize`, `figure.layout.height`, `figure.layout.width`.
+     */
+    responsive: PropTypes.oneOf([true, false, 'auto']),
```
Marc-Andre-Rivet: Allow users to define the behavior they want for their graph, allowing them to override the normal behavior for their `config`/`figure` values.

src/components/Graph.react.js
```diff
@@ -491,6 +505,7 @@ PlotlyGraph.defaultProps = {
         layout: {},
         frames: [],
     },
+    responsive: false,
```
Marc-Andre-Rivet: Defaults to responsive `False` which is the current behavior.

src/fragments/Graph.react.js
```diff
@@ -1,9 +1,42 @@
...
+ * `responsive: true` causes Plotly.js to resize the graph on `window.resize`.
+ * This is necessary for `dcc.Graph` call to `Plotly.Plots.resize(target)` to do something.
+ *
+ * Users can override this value for specific use-cases by explicitly passing `responsive: false`
+ * if `responsive` is not set to True.
+ */
+const DEFAULT_CONFIG = {
+    responsive: true,
+};
+
```
Marc-Andre-Rivet: When responsive is true, whether through true or auto, will apply these values to config/figure to get the desired behavior.

src/fragments/Graph.react.js
```diff
@@ -154,7 +206,28 @@ class PlotlyGraph extends Component {
         clearExtendData();
     }
 
+    isResponsive(props) {
+        const {config, figure, responsive} = props;
+
+        if (type(responsive) === 'Boolean') {
+            return responsive;
+        }
```
Marc-Andre-Rivet: If responsive is a boolean, use as-is. Otherwise, determined auto responsiveness by looking at config/figure for the graph.

src/fragments/Graph.react.js
```diff
@@ -277,15 +346,27 @@ class PlotlyGraph extends Component {
...
-            />
+            >
+                <ResizeDetector
+                    handleHeight={true}
+                    handleWidth={true}
+                    refreshMode="debounce"
+                    refreshOptions={{trailing: true}}
+                    refreshRate={50}
+                    onResize={this.graphResize}
+                />
```
Marc-Andre-Rivet: element resize observer - either uses ResizeObserver in newer browsers, or polyfills with a DOM/css method if it doesn't. Debounce / trail / wait to prevent making useless calls.

src/fragments/Graph.react.js
```diff
@@ -277,15 +346,27 @@ class PlotlyGraph extends Component {
...
                 id={id}
-                ref={this.gd}
+                key={id}
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
                 style={style}
-                className={className}
-            />
+            >
```
Marc-Andre-Rivet: The graph is now wrapped in a `div` that applies the styling -- the graph div itself has consistent styling. This helps responsiveness to work correctly in flex flows.

nicolaskruchten:
I would recommend adding `px.defaults.height = None` to the testing sandbox to see if it makes certain things unnecessary. I'm open to the idea that PX is too opinionated about height if it's causing issues :) It certainly seems to fix the flash-of-too-tall-chart and the charts-are-too-tall-on-callback things.

tests/integration/graph/test_graph_responsive.py
```diff
@@ -0,0 +1,125 @@
...
+from dash.exceptions import PreventUpdate
+from utils import wait_for
+
+
+@pytest.mark.parametrize("responsive", [True, False, None])
+@pytest.mark.parametrize("autoresize", [True, False, None])
+@pytest.mark.parametrize("height", [600, None])
+@pytest.mark.parametrize("width", [600, None])
+@pytest.mark.parametrize("is_responsive", [True, False, 'auto'])
+def test_grrs001_graph(
```
Marc-Andre-Rivet: This test makes sure that all combinations of responsiveness / dimensions behave as intended.
Not taking any screenshot, we're just interested in the size of the graph div on 1st render and on resize.

src/fragments/Graph.react.js
```diff
@@ -154,11 +207,62 @@ class PlotlyGraph extends Component {
...
+            return;
         }
+
+        ++this._resizes;
+        gd.classList.add('dash-graph-resizing');
+
+        Plotly.Plots.resize(gd)
+            .catch(() => {})
+            .finally(() => {
+                --this._resizes;
```
Marc-Andre-Rivet: We don't really care if it's successful or not, calls and responses just need to balance out to know we're "ready"

src/fragments/Graph.react.js
```diff
@@ -154,11 +215,72 @@ class PlotlyGraph extends Component {
...
-            Plotly.Plots.resize(gd);
+        if (!gd) {
+            return;
         }
+
+        gd.classList.add('dash-graph-resizing');
+
+        Plotly.Plots.resize(gd)
+            .catch(() => {})
+            .finally(() => gd.classList.remove('dash-graph-resizing'));
```
alexcjohnson: looks like IE11 doesn't support `finally` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally
anyway you've already caught, so can't you just `.then`?

src/fragments/Graph.react.js
alexcjohnson: I'm still concerned about adding `js-plotly-plot` to the outer div - seems like having two nested divs with this class could lead to strange behavior, if it's actually getting used for something in CSS. Can you explain the reasoning here?

src/fragments/Graph.react.js
Marc-Andre-Rivet: Promise.finally gets polyfilled by https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.promise.finally.js

src/fragments/Graph.react.js
Marc-Andre-Rivet: Understandable. My goal here is for old versions of DDK to continue having the same behavior. Given the way DDK styling works and the stated goal of minimizing custom CSS, this should be fine. To break this behavior would require users to assign `display: flex;` to `js-plotly-plot`. The next DDK release will be using `.dash-graph`.

For the DCC world, this is in my opinion a non-issue, the general idea is that we support classes & attributes with the `dash-` prefix, everything else is implementation details.

Marc-Andre-Rivet: If we don't add `js-plotly-plot`, `dash>=1.8.0` will be a breaking version.
If we do add it, we can always create an issue to remove it and tag that issue as `Dash v2.x`.

Marc-Andre-Rivet:
Apart from waiting for the release from Plotly.js, the last remaining contention point might be the use of `js-plotly-plot` on the wrapper class for legacy purposes. Created https://github.com/plotly/dash-docs/issues/728 to update the documentation with information on the basic assumptions used when coding against classes in Dash.

.gitignore
```diff
@@ -17,6 +17,7 @@ venv/
 
 /build
 /dash_core_components
+/dash_core_components_base/plotly.min.js
```
Marc-Andre-Rivet: Cleaned up generation a bit to copy file directly from node_modules instead of keeping a to-be-synced copy in the py generation base folder

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
...
+### Added
+- [#706](https://github.com/plotly/dash-core-components/pull/706)
+  - Added new `responsive` property that overrides the underlying Plotly.js graph responsiveness from Dash-land
+  - Added responsiveness on graph parent element resize (previously only worked on window.resize)
+  - Added new `dash-graph--pending` class to dcc.Graph, present while resizing, (re-)rendering, loading
+
+### Updated
+- [#706](https://github.com/plotly/dash-core-components/pull/706)
+  - Upgraded plotly.js to [1.51.3](https://github.com/plotly/plotly.js/releases/tag/v1.51.3)
+
```
Marc-Andre-Rivet: Plotly.js 1.51.3 with the promise fix 🎉

CONTRIBUTING.md
```diff
@@ -34,9 +34,7 @@ Locally](README.md#testing-locally) section of README.md.
 ## Updating Plotly.js
 
 1. Update the version of `plotly.js` in package.json. Always use an exact version without "^" or "~"
-2. Run `npm install` followed by `npm run build`, this will ensure the latest version of Plotly.js is in `node_modules` and copy
-that version over with the other build artifacts
-3. Update `dash_core_components_base/__init__.py` plotly.js `relative_package_path` and `external_url`
+2. Run `npm install` followed by `npm run build`, the Plotly.js artifact will be copied and bundled over as required
```
Marc-Andre-Rivet: Plotly.js artifact is now always `plotly.min.js`, no need for weird updates

src/components/Graph.react.js
```diff
@@ -90,24 +94,63 @@ const RealPlotlyGraph = asyncDecorator(PlotlyGraph, () =>
...
+                    key={id}
+                    className={`${extendedClassName} dash-graph--pending`}
+                />
+            }
+        >
+            <RealPlotlyGraph
+                {...{
+                    ...props,
+                    className: extendedClassName,
+                }}
```
alexcjohnson: the double `...` looks funny - is this different from
`<RealPlotlyGraph {...props} className={extendedClassName} />`?
Or more explicitly replacing:
`<RealPlotlyGraph {...R.assoc('className', extendedClassName, props)} />`

src/fragments/Graph.react.js
Marc-Andre-Rivet: As pointed out by @alexcjohnson, in the end `js-plotly-plot` is not necessary as old ddk uses the forked version of the graph anyway.

The only scenario left is a new version of DDK pointing to an old version of Dash, but that can be handled through CSS with `*:not(.dash-graph) > .js-plotly-js`

src/components/Graph.react.js
Marc-Andre-Rivet: It's the same, will update for cleaner usage

src/components/Graph.react.js
```diff
@@ -90,24 +92,58 @@ const RealPlotlyGraph = asyncDecorator(PlotlyGraph, () =>
...
+                    className={`${extendedClassName} dash-graph--pending`}
+                />
+            }
+        >
+            <RealPlotlyGraph {...props} className={extendedClassName} />
+        </Suspense>
+    );
+});
+
+ControlledPlotlyGraph.propTypes = PropTypes.any;
```
alexcjohnson: @Marc-Andre-Rivet I'm seeing a funny warning message in debug mode:
```
The prop `isRequired` is marked as required in `<<anonymous>>`,
but its value is `undefined`
```

That I think is coming from here... Should this just be a copy of `PlotlyGraph.propTypes` perhaps?

bcliang:
@Mark-Andre-Rivet -- Apologies for the thread dig.

Re: the side-effect of this PR... it also breaks the ability for end users to modify the contained plotly figure using clientside callbacks. Why? Client-side "UI" changes are much faster and provides a better user experience for things like changing axis on ranges, updating titles, etc.

Specifically, I've been using a this approach to trigger relayout via `Plotly.relayout('graph-id', {..'layout-property':'value'; ..}`. 

This problem is resolved by shifting the component's `id` and `key` assignment into the nested div (i.e. the direct parent of `<div class="plot-container plotly">`). I created a simple integration test to check clientside callbacks in my fork (https://github.com/bcliang/dash-core-components/tree/fix-plotly-clientside-callbacks) which seems to work as expected.

However, running the full test suite, I get some flaky results (only passes on first run) due to parameterized `eager_loading`. Not sure if you have any thoughts.

PR #711, opened 2019-12-03
Issue 250 - Support for rendering links inside `dcc.Markdown` as `dcc.Link`

### LINKED ISSUE #250: Support for rendering links inside `dcc.Markdown` as `dcc.Link` for single page dash apps
I use `dcc.Markdown` really extensively in `dash-docs`. It's great! However, a few things would make my life a lot easier:
1. (Done!) GitHub style language tags, that is:

```
``python
def dash():
     pass
```

**Edit - This has been done!**

2. Ability for the hyper links to use `dcc.Link` instead of the HTML link

Currently, I have to break out my `dcc.Link` from `dcc.Markdown`, which is pretty tedious:
https://github.com/plotly/dash-docs/blob/58b6f84f2d8012d1ae686f1379f326a292370ee3/tutorial/getting_started_part_2.py#L260-L269

3. (Done!) Automatic dedenting. Right now, I use `textwrap.dedent` everwhere in my text. If I don't use `dedent`, then the markdown is formatted as code (4 indents in markdown is code). It would be nice if I could just pass in `dedent=True` or something

**Edit - This has been done!**

Marc-Andre-Rivet:
Fixes #250 

Adds additional node type for `dcc.Link` (vs. `link`) and the general mechanism to extend markdown syntax with any React Component of our choosing. Leaves the existing link syntax untouched.

~~Works as-is, no need to set `dangerously_allow_html=True` -- which I'm not sure makes sense? Need to investigate / make sure we're not doing something wrong..~~

Requires `dangerously_allow_html=True` to work correctly.

Todo

- [x] changelog
- [x] tests (dccLink w/ & w/o dangerouslySetInnerHtml, link works, html child/text, img child, dcc.Link vs link styling, inline dccLink, etc.)
- [x] docs (https://github.com/plotly/dash-docs/issues/721) -- tagged for `v1.8` also

```
import dash

app = dash.Dash(__name__)
app.layout = dcc.Markdown("""
    # Link
    [Outside Link](https://www.google.com)
    # DCC Link!
    <dccLink href="potato" children="potato" />
    <dccLink href="with_gravy" children="with gravy" />
""")

app.run_server(debug=True)
```

image


webpack.config.js
```diff
@@ -58,7 +58,7 @@ module.exports = (env, argv) => {
             rules: [
                 {
                     test: /\.jsx?$/,
-                    exclude: /node_modules/,
+                    exclude: /node_modules\/(?!react-jsx-parser\/)/,
```
Marc-Andre-Rivet: Sadly not transpiled to ES5 so needs to be re-transpiled.

src/fragments/Markdown.react.js
```diff
@@ -85,6 +88,10 @@ export default class DashMarkdown extends Component {
         const displayText =
             dedent && textProp ? this.dedent(textProp) : textProp;
 
+        const componentTransforms = {
+            DccLink: props => <DccLink {...props} />,
+        };
```
Marc-Andre-Rivet: Here we define the list of components we want to support are their name in markdown.

chriddyp:
Very cool syntax. Should we do `<DccLink ` or `<dccLink`? The latter is similar to the Python and R (cc @sudburyrob ) syntax but maybe less semantic in terms of XML

src/fragments/Markdown.react.js
```diff
@@ -110,6 +117,14 @@ export default class DashMarkdown extends Component {
                 <Markdown
                     source={displayText}
                     escapeHtml={!dangerously_allow_html}
+                    renderers={{
+                        html: props => (
```
chriddyp: should we toss in a danderous HTML / XSS test for future sanity?

chriddyp:
I like it. Nice general solution. 💃

chriddyp:
Does this work?
```
<DccLink href="/lahlah">
Test
</DccLink>
```

Also, does this work?
```
<DccLink href="/lahlah">
![Alt text for an image](https://plot.ly~chris/1638.png)
</DccLink>
```

chriddyp:
Also could we do a image test that our CSS is OK? That is, that inline `dcc.Link` is rendered the same way as an inline markdown link?

```
This is a [regular html link](/test) within a line of text.

This is a <DccLink href="/test">special dcc link</DccLink> within a line of text.

You can't tell them apart!
```

chriddyp:
Sorry @sudburyrob I meant to tag @rpkyle

Marc-Andre-Rivet:
@chriddyp We're in html land now so this does not work:
```
<DccLink href="/lahlah">
![Alt text for an image](https://plot.ly~chris/1638.png)
</DccLink>
```

But this works:
```
<dccLink href="lots_of_cheese">
        <img src="https://plot.ly/~chris/1638.png"/>
    </dccLink>
```

Marc-Andre-Rivet:
> Should we do `DccLink` or `dccLink`? The latter is similar to the Python and R

`dccLink` does make more sense, and it works (does not suffer React's capitalization limitations). Updating.

Marc-Andre-Rivet:
One difference between using `dccLink` through the html renderer and `link` (e.g. `[Link](/link-path)`) gets wrapped in a `<p>...</p>` along with everything else in the line.

chriddyp:
Another nice enhancement (down the line) would be to enable end users to specify the properties of their `html.A` like `target="blank"`, so:
```
dcc.Markdown('''
Long string of text with a <a href="https://plot.ly" target="blank">absolute link</a> in it.
''')
```

chriddyp:
> One difference between using dccLink through the html renderer and link (e.g. [Link](/link-path)) gets wrapped in a <p>...</p> along with everything else in the line.

Does this mean that using `<dccLink/>` in markdown will bring the content into a new line rather than keeping it with the text? If so, then that would reduce the ease of use of this enhancement.

Marc-Andre-Rivet:
> Does this mean that using <dccLink/> in markdown will bring the content into a new line rather than keeping it with the text? 

Testing it...

```
dcc.Markdown(['This is an inlined <dccLink href="title_crumb" children="Title" /> with text on both sides'], dangerously_allow_html=True,),
```

Currently renders as:
image

Checking if something can be done about it.

PR #716, opened 2019-12-13
Issue 688 - Speed up build for DCC

Marc-Andre-Rivet:
On my local machine, build time for DCC goes from ~5 minutes down to ~45 seconds for a no-cache rebuild and ~18 seconds for a "with cache" rebuild.

From the perspective of a DCC user or developer, having access to the Plotly.js sourcemap is of little practical value in most scenarios.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,4 +1,6 @@
 const path = require('path');
+const TerserPlugin = require('terser-webpack-plugin');
+const webpack = require('webpack');
 const WebpackDashDynamicImport = require('@plotly/webpack-dash-dynamic-import');
 
 const packagejson = require('./package.json');
@@ -34,8 +36,6 @@ module.exports = (env, argv) => {
 
     const entry = overrides.entry || { main: './src/index.js' };
 
-    const devtool = overrides.devtool || 'source-map';
-
     const externals = ('externals' in overrides) ? overrides.externals : ({
         react: 'React',
         'react-dom': 'ReactDOM',
@@ -52,9 +52,9 @@ module.exports = (env, argv) => {
             library: dashLibraryName,
             libraryTarget: 'window',
         },
-        devtool,
         externals,
         module: {
+            noParse: /node_modules\/plotly.js/,
             rules: [
                 {
                     test: /\.jsx?$/,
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -52,9 +52,9 @@ module.exports = (env, argv) => {
             library: dashLibraryName,
             libraryTarget: 'window',
         },
-        devtool,
         externals,
         module: {
+            noParse: /node_modules\/plotly.js/,
```
Marc-Andre-Rivet: We know the output of Plotly.js is (1) valid, (2) self-contained (doesn't it to resolve outward facing define/import/require), so don't parse it, just take the file as-is.

webpack.config.js
```diff
@@ -85,6 +85,17 @@ module.exports = (env, argv) => {
             }
         },
         optimization: {
+            minimizer: [
+                new TerserPlugin({
+                    sourceMap: true,
+                    parallel: true,
+                    cache: './.build_cache/terser',
```
Marc-Andre-Rivet: Two things matter here: (1) the cache folder, (2) source mapping in the minimizer, this allows the devtool plugin to override which chunk will get mapped (vs. `devtools` prop)

webpack.config.js
```diff
@@ -52,9 +52,9 @@ module.exports = (env, argv) => {
             library: dashLibraryName,
             libraryTarget: 'window',
         },
-        devtool,
```
Marc-Andre-Rivet: Now handled by the minimizer. Makes it possible to exclude chunks from mapping

webpack.config.js
```diff
@@ -99,7 +110,11 @@ module.exports = (env, argv) => {
             }
         },
         plugins: [
-            new WebpackDashDynamicImport()
+            new WebpackDashDynamicImport(),
+            new webpack.SourceMapDevToolPlugin({
+                filename: '[name].js.map',
+                exclude: ['async~plotlyjs']
```
Marc-Andre-Rivet: source mapping plotlyjs accounted for the majority of the build time (15s to build + 30s to map non-plotlyjs chunks + 30s to parse plotlyjs + 3.5mn to map plotlyjs)

PR #720, opened 2019-12-17
Refactor markdown syntax highlighting.

### LINKED ISSUE #713: Bundle highlight.js into dcc
Highlight.js used by dcc is not bundled and sits in `window.hljs`.
Similarly to plotly.js usage, use the user provided version if available, otherwise load the default.

This is related to https://github.com/plotly/dash-table/pull/606

shammamah:
Closes #713 

Related to https://github.com/plotly/dash-table/pull/606


src/fragments/Markdown.react.js
```diff
@@ -2,12 +2,18 @@ import React, {Component} from 'react';
...
 import Markdown from 'react-markdown';
 
+import Highlight from '../utils/Highlight';
 import {propTypes, defaultProps} from '../components/Markdown.react';
 import '../components/css/highlight.css';
 
 export default class DashMarkdown extends Component {
     constructor(props) {
         super(props);
+        /* eslint-disable no-new */
```
shammamah: Disabling this error as we really only do want the side-effects of creating a new `Highlight` object. An alternative is wrapping all of the `Highlight` code with a function instead of a class.

src/utils/Highlight.js
```diff
@@ -0,0 +1,21 @@
+import LazyLoader from './LazyLoader';
+
+export default class Highlight {
+
+    static hljsResolve() {}
+
+    constructor() {
+        if(!Highlight.isReady) {
```
shammamah: Put this here to handle the case in which there are multiple `dcc.Markdown` elements that need syntax highlighting; we won't create new promises for each of them.

third-party/highlight.js
```diff
@@ -0,0 +1,32 @@
+import highlightjs from 'highlight.js/lib/highlight';
+import 'highlight.js/styles/github.css';
+
+import bash from 'highlight.js/lib/languages/bash';
```
shammamah: Using all of the languages that are defined in the old `highlight.pack.js` file. We might want to standardize this set of languages so that `dash-table` and `dash-core-components` support the same languages.

src/fragments/Markdown.react.js
Marc-Andre-Rivet: Can you explain what's happening here?

dash_core_components_base/__init__.py
```diff
@@ -70,10 +71,6 @@
     } for async_resource in async_resources])
 
 _js_dist.extend([
-    {
-        'relative_package_path': 'highlight.pack.js',
-        'namespace': 'dash_core_components'
-    },
```
Marc-Andre-Rivet: :tada

src/fragments/Markdown.react.js
```diff
@@ -21,15 +29,15 @@ export default class DashMarkdown extends Component {
...
             const nodes = this.mdContainer.querySelectorAll('pre code');
 
-            for (let i = 0; i < nodes.length; i++) {
-                window.hljs.highlightBlock(nodes[i]);
+            if (MarkdownHighlighter.hljs) {
+                for (let i = 0; i < nodes.length; i++) {
+                    MarkdownHighlighter.hljs.highlightBlock(nodes[i]);
+                }
+            } else {
+                MarkdownHighlighter.loadhljs();
```
Marc-Andre-Rivet: I think the existing implementation does not take full advantage of the `react-markdown` ability to get its renderer overriden -- similarly to https://github.com/plotly/dash-core-components/pull/711/files#diff-5b56ed82805404e28302aef5147d04a2R129 it should be possible to hook up the highlight directly onto the `code` renderer and simplify this component / usage.

Marc-Andre-Rivet: Issue created: https://github.com/plotly/dash-core-components/issues/724, will not be addressed as part of this PR.

Marc-Andre-Rivet:
Snapshots involving dcc.Markdown are not consistent with how they were before -- they lack styling. This may be a timing issue or a new bug: https://percy.io/plotly/dash-core-components/builds/3530615/view/215175776/1280?mode=diff&browser=chrome&snapshot=215175776

Would like to see a test similar to what was done with the table. Behavior with and without highlighting override.

shammamah:
@Marc-Andre-Rivet It looks like the Percy diffs are because I am importing a stylesheet from `highlight.js` that overrides the `highlight.css` stylesheet we have in this repo, which has styles taken from the `arduino-light` and `monokai` stylesheets from the library. (https://github.com/plotly/dash-core-components/blob/dev/src/components/css/highlight.css)

cc @wbrgss as this might be relevant to DDK. For now I've just removed the stylesheet import and it should look the same as before.

wbrgss:
@shammamah I'm actually theming the `highlight.css` based on the DDK theme. It's in 
[override-highlight-syntax.css.js](https://github.com/plotly/dash-design-kit/blob/dev/src/lib/css_js/override-highlight-syntax.css.js) (private repo). 

Does that make sense? I wonder if I have to increase the specificity of those files' rules given your change.

shammamah:
@wbrgss That makes sense! I have no problem leaving out the `highlight.js` stylesheet. It should work exactly as before.

src/fragments/Markdown.react.js
```diff
@@ -2,12 +2,19 @@ import React, {Component} from 'react';
 import {type} from 'ramda';
 import Markdown from 'react-markdown';
 
+import MarkdownHighlighter from '../utils/MarkdownHighlighter';
 import {propTypes, defaultProps} from '../components/Markdown.react';
 import '../components/css/highlight.css';
```
Marc-Andre-Rivet: If we're to keep this custom css, might as well import it when we request hljs instead of getting it now.

src/utils/MarkdownHighlighter.js
```diff
@@ -0,0 +1,18 @@
+import lazyhljs from './LazyLoader/hljs';
+import '../components/css/highlight.css';
```
Marc-Andre-Rivet: Better, than it the Markdown component, but I would push it even further down the stack, right down to `src/third-party/highlight.js`

tests/integration/misc/test_markdown_highlight.py
```diff
@@ -0,0 +1,28 @@
...
+    window_hljs = dash_dcc.driver.execute_script('return window.hljs')
+    assert window_hljs is None
+
+
+def test_msmh001_window_override(dash_dcc):
+    app = dash.Dash(__name__)
+    app.layout = html.Div(dcc.Markdown(md_text))
+    dash_dcc.start_server(app)
+
+    dash_dcc.driver.execute_script('window.hljs = {highlightBlock: (block) => {block.innerHTML="hljs override"}};')
```
Marc-Andre-Rivet: Here we assign `window.hljs` but nothing in this test demonstrate that it is being used.
Could we add a `dash_dcc.percy_snapshot('<name of snapshot>')` or similar to see that the result is what was forced.

tests/integration/misc/test_markdown_highlight.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+    def trigger_md_rerender(_):
+        return dcc.Markdown(md_text)
+
+    dash_dcc.start_server(app)
+
+    dash_dcc.driver.execute_script('window.hljs = {highlightBlock: (block) => {block.innerHTML="hljs override"}};')
+
+    assert dash_dcc.find_element('code').text == 'hljs override'
+    dash_dcc.percy_snapshot('md_code_highlight_override')
```
Marc-Andre-Rivet: 🎉

PR #722, opened 2019-12-18
Update R package metadata YAML file for 1.6.0

rpkyle:
The YAML data used for populating title/description/examples should be up to date in the `dev` branch. This PR will backcommit the updates.


rpkyle:
@Marc-Andre-Rivet Could we merge this if there are no objections?

PR #723, opened 2019-12-18
Allow projects to use dcc source directly

Marc-Andre-Rivet:
- add /lib with babel transpiled js code to published package
- fix `deps` vs `devDeps` for source usage
- Rework LazyLoader to allow tree shaking of async resources


package.json
```diff
@@ -12,6 +12,7 @@
   "homepage": "https://github.com/plotly/dash-core-components",
   "main": "dash_core_components/dash_core_components.min.js",
   "scripts": {
+    "prepublishOnly": "rm -rf lib && babel src --out-dir lib --copy-files",
```
Marc-Andre-Rivet: On prepublish, transpile all source to /lib

package.json
```diff
@@ -48,9 +49,11 @@
     "react-markdown": "^4.0.6",
     "react-select-fast-filter-options": "^0.2.3",
     "react-virtualized-select": "^3.1.3",
+    "react-resize-detector": "^4.2.1",
```
Marc-Andre-Rivet: Required for 3rd party build inclusion

package.json
```diff
@@ -48,9 +49,11 @@
...
     "uniqid": "^5.0.3"
   },
   "devDependencies": {
+    "@babel/cli": "^7.4.0",
```
Marc-Andre-Rivet: Allows using `babel` from the command line

package.json
```diff
@@ -88,8 +90,10 @@
     "webpack-serve": "^2.0.3"
   },
   "files": [
-    "/dash_core_components/*{.js,.map}"
+    "/dash_core_components/*{.js,.map}",
+    "/lib/**"
```
Marc-Andre-Rivet: Include in the published package

package.json
```diff
@@ -88,8 +90,10 @@
...
   ],
+  "sideEffects": false,
```
Marc-Andre-Rivet: Declare this package side-effect free. Let's tools know they can tree-shake without problems.

src/utils/LazyLoader.js
```diff
@@ -1,24 +1,28 @@
...
-                    ({default: Plotly}) => {
-                        window.Plotly = Plotly;
-                        return Plotly;
-                    }
-                )
-        ),
-    upload: () =>
-        import(/* webpackChunkName: "upload" */ '../fragments/Upload.react'),
-};
+export const datePickerRang = () =>
```
Marc-Andre-Rivet: rework the lazy loader so that it can be tree-shaken.. so.. not all on the same object 😄

webpack.config.js
```diff
@@ -112,7 +112,7 @@ module.exports = (env, argv) => {
         plugins: [
             new WebpackDashDynamicImport(),
             new webpack.SourceMapDevToolPlugin({
-                filename: '[name].js.map',
+                filename: '[file].map',
```
Marc-Andre-Rivet: caused the entrypoint sourcemap to be `main.js.map` instead of `dash_core_components.min.js.map`

package.json
Marc-Andre-Rivet: Include the css files too! :)

src/utils/LazyLoader/plotly.js
```diff
@@ -0,0 +1,6 @@
+export default () => Promise.resolve(window.Plotly ||
+    import(/* webpackChunkName: "plotlyjs" */ 'plotly.js').then(({ default: Plotly }) => {
+        window.Plotly = Plotly;
+        return Plotly;
+    }));
```
Marc-Andre-Rivet: Splitting it all up for tree shaking reasons..

src/utils/LazyLoader/plotly.js
```diff
@@ -0,0 +1,6 @@
+export default () => Promise.resolve(window.Plotly ||
```
Marc-Andre-Rivet: While we're here, might be interesting to check if `window.Plotly.version` matches our expected minimal version, and if not, force loading the one in our bundle.. this will also be useful to protect against old versions of dash-bio dumping an old version of Plotly.js for DCC to use.

PR #725, opened 2019-12-19
Performance improvement: Fetch graph and plotly in parallel instead of sequentially

Marc-Andre-Rivet:
Updates the graph's async resources fetching to be done in parallel instead of sequentially. Reducing start to finish fetching time slightly.

### DIFF
--- a/src/components/Graph.react.js
+++ b/src/components/Graph.react.js
@@ -89,7 +89,10 @@ class PlotlyGraph extends Component {
     }
 }
 
-const RealPlotlyGraph = asyncDecorator(PlotlyGraph, () => plotly().then(graph));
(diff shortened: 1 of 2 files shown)


src/components/Graph.react.js
```diff
@@ -89,7 +89,9 @@ class PlotlyGraph extends Component {
     }
 }
 
-const RealPlotlyGraph = asyncDecorator(PlotlyGraph, () => plotly().then(graph));
+const RealPlotlyGraph = asyncDecorator(PlotlyGraph, () =>
+    Promise.all([plotly(), graph()]).then(([, graph]) => graph)
+);
```
Marc-Andre-Rivet: parallelize requests

PR #727, opened 2019-12-30
Setting Locale in DatePicker Components

### LINKED ISSUE #579: Feature request: DateRangePicker locale
There is no way to change the language used for this component. For instance, I would like to be able to set the locale to 'fr_CA' to render the dates in french.

The dates can currently only be displayed in english: image

I found that post https://community.plot.ly/t/changing-calendars-langage-to-french/19983 on the forum, but the solution proposed is hacky and it did not even work in my case.

I feel like this is a reasonable feature request and I'm sure a lot of people would find that feature useful.

SimonPavlik:
Introducing a new language property used to set locale in **DatePickerSingle** and **DatePickerRange** components.
* Passing the default parameter `auto` will pick the browser's default language to set the locale. 
* *For example:* `language='en'` sets the default locale of these components globally to English. 
* *Caveat:* The same locale needs to be set for all DatePicker components located on the same page
* This limitation is caused by the fact that the locale is being set globally using [Moment.js](https://momentjs.com/docs/#/i18n/) - a peer dependency of [react-dates](https://github.com/airbnb/react-dates).

This PR resolves #579 and https://community.plot.ly/t/changing-calendars-langage-to-french/19983.

### DIFF
--- a/src/components/DatePickerRange.react.js
+++ b/src/components/DatePickerRange.react.js
@@ -261,6 +261,16 @@ DatePickerRange.propTypes = {
      * session: window.sessionStorage, data is cleared once the browser quit.
      */
     persistence_type: PropTypes.oneOf(['local', 'session', 'memory']),
+
+    /**
+     * The language used to set locale globally.
+     * Pass the default parameter `auto` to use the browser's default language.
+     * For example: `language='en'` sets the default locale of this component globally to English.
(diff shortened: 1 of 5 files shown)


roskoN:
Hey,

I do have a great interest in moving this forward. 

Is there any way I could help? Are the unit-tests the only thing that is blocking this PR?

Thanks!

PR #729, opened 2020-01-11
feat: disable cookie/localstorage access under restricted sandboxes

josegonzalez:
When dash is embedded into an iframe with a sandbox attribute that only has allow-scripts, cookie/localstorage access is disabled and dash-core-components fails to load. As such, we need to restrict our cookie/localstorage usage by disabling functionality.

This patch removes the disabled functionality in a graceful manner, allowing dash-core-components to load in very restricted iframes.

### DIFF
--- a/src/components/Store.react.js
+++ b/src/components/Store.react.js
@@ -112,9 +112,6 @@ class WebStore {
     }
 }
 
-const _localStore = new WebStore(window.localStorage);
-const _sessionStore = new WebStore(window.sessionStorage);
-
 /**
  * Easily keep data on the client side with this component.
  * The data is not inserted in the DOM.
@@ -126,9 +123,9 @@ export default class Store extends React.Component {
         super(props);
 
         if (props.storage_type === 'local') {
-            this._backstore = _localStore;
+            this._backstore = new WebStore(window.localStorage);
         } else if (props.storage_type === 'session') {
-            this._backstore = _sessionStore;
+            this._backstore = new WebStore(window.sessionStorage);
         } else if (props.storage_type === 'memory') {
             this._backstore = new MemStore();
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Changed
+- [#729](https://github.com/plotly/dash-core-components/pull/729) Handle case where dcc fails to load when used inside an iframe with a sandbox attribute that only has allow-scripts
+
 ## [1.7.1] - 2020-01-15 (JS-only)
 ### Fixed
 - [#734](https://github.com/plotly/dash-core-components/pull/734) Fix JS-facing release bug where `Plotly.js` was listed in `devDependencies` instead of `dependencies`


src/components/Store.react.js
```diff
@@ -125,11 +133,11 @@ export default class Store extends React.Component {
...
 
-        if (props.storage_type === 'local') {
-            this._backstore = _localStore;
+        if (props.storage_type === 'local' && localStorageTest()) {
+            this._backstore = new WebStore(window.localStorage);
         } else if (props.storage_type === 'session') {
-            this._backstore = _sessionStore;
-        } else if (props.storage_type === 'memory') {
+            this._backstore = new WebStore(window.sessionStorage);
+        } else {
```
Marc-Andre-Rivet: This change is unlike the similarly themed table change (https://github.com/plotly/dash-table/pull/677) that impact logging but no user-facing feature.

This component defaults to `memory` storage and `local` needs to be explicitly set to potentially run into issues with iframes. This will silently break expectations and will look like a bug to most users as the expectation will be for `local` to store information across sessions and `session` to store information across page reloads (e.g. docs example in https://dash.plot.ly/dash-core-components/store).

I think the correct approach here is to **not** try and be smart and **not** degrade functionality. Instead, I would suggest that we throw an error that will appear in the browser's devtools and that will be picked up by the Dash devtools with `debug=True`. The application developers will then be able to fix their code or its environment settings to either respect the sandbox or unrestrict localStorage as needed.

Tagging @alexcjohnson for a second opinion.

If we do decide to go the degradation way, the localStorage should at least degrade to sessionStorage instead of memory storage.

alexcjohnson: yeah that makes sense. As long as the error happens on trying to create the component, rather than on initial script execution as it does prior to this PR.

> localStorage should at least degrade to sessionStorage instead of memory

Does `sessionStorage` still work when `localStorage` is forbidden?

josegonzalez: Were we validating the `storage_type` value anywhere previously?

I guess my overall desire here is to ensure we're not always attempting to use window.localStorage/window.sessionStorage if those aren't available, which we were previously doing.

How do we throw the error for dash devtools?

Marc-Andre-Rivet: I think `sessionStorage` is allowed but might be sandboxed from the parent page's `sessionStorage` based on attributes. Checking.

josegonzalez: I've reverted and added a fallback to sessionstorage, but I can also remove that fallback if it's okay to just error directly.

Marc-Andre-Rivet: There seems to be no way to both successfully (a) create the Store component and (b) throw an Error that will be caught by the React Error Boundary. Given that limitation, the correct implementation for now IMO would be to throw the error / fail the Store component rendering.

I will create a followup issue to look into making it possible to both (a) degrade / render a component behavior, and (b) expose the error in Dash devtools. (e.g. have a communication mechanism that does not use the Error Boundary to expose the error)

alexcjohnson: yeah, it's OK to just throw an error during the component lifecycle - that'll make it to the devtools.

> I guess my overall desire here is to ensure we're not always attempting to use window.localStorage/window.sessionStorage if those aren't available, which we were previously doing.

That's what I was getting at with "As long as the error happens on trying to create the component, rather than on initial script execution as it does prior to this PR." Don't access any storage systems on initial execution; when you make a component, try to access the requested storage system and only that one, and allow it to throw an error if that fails.

josegonzalez: I removed the fallback code and localstorage test.

src/components/Store.react.js
```diff
@@ -126,9 +123,9 @@ export default class Store extends React.Component {
         super(props);
 
         if (props.storage_type === 'local') {
-            this._backstore = _localStore;
+            this._backstore = new WebStore(window.localStorage);
```
Marc-Andre-Rivet: The error message we get when trying to access `localStorage` is fairly informative / helpful:

```
Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': The document is sandboxed and lacks the 'allow-same-origin' flag.
```

I don't think it needs to be improved to make sense to an app developer and as such can be exposed as-is.

Marc-Andre-Rivet:
Created https://github.com/plotly/dash/issues/1088 for follow up.

PR #730, opened 2020-01-13
Fix number inputs for IE11

### LINKED ISSUE #627: [BUG] Internet Explorer dcc.Input with type='numeric'
Hello,

The following example works in Chrome, firefox, new Edge (not old Edge), and not at all in Internet Explorer.

Note - if the input type is changed to 'text' then the issue is not present.
Also, this is a fairly recent regression - less than 1 month old.

Nothing happens when numbers are changed. Also, if you query what is inside `'my-id'`, you will find `None`

```python
import dash

app.layout = html.Div([
    dcc.Input(id='my-id', value='', type='number'),
    html.Div(id='my-div')
])


@app.callback(
    Output(component_id='my-div', component_property='children'),
    [Input(component_id='my-id', component_property='value'), Input(component_id='my-id', component_property='n_submit')]
)
def update_output_div(input_value,submit):
    print(input_value)
    return 'You\'ve entered "{}"'.format(input_value)


if __name__ == '__main__':
    app.run_server(debug=True)
```

shammamah:
Closes #627.

## About 
The core of the issue was that `valueAsNumber` was always returning `NaN` even if the input value was a number, which is a known issue with IE: https://caniuse.com/#feat=input-number

Screen Shot 2020-01-13 at 1 47 37 PM


src/components/Input.react.js
```diff
@@ -89,6 +92,13 @@ export default class Input extends PureComponent {
         );
     }
 
+    getValueAsNumber(value) {
+        if (parseInt(value, 10) !== parseFloat(value)) {
+            return parseFloat(value);
+        }
+        return parseInt(value, 10);
```
Marc-Andre-Rivet: Why not save the result of `parseInt` and `parseFloat` to variables and test / return those instead of re-parsing?

Marc-Andre-Rivet: Also, I think this will behave differently from the current implementation for something like "1e4", with `current.valueAsNumber` returning `10000` and `parseInt` returning `1`.

When possible we've been using https://github.com/plotly/fast-isnumeric combined with `+myString` to validate and cast strings into numbers.

src/components/Input.react.js
```diff
@@ -89,6 +92,11 @@ export default class Input extends PureComponent {
         );
     }
 
+    getValueAsNumber(value) {
+        const numericValue = convert(value);
+        return numericValue % 1 === 0 ? Math.floor(numericValue) : numericValue;
```
Marc-Andre-Rivet: @shammamah Can you explain the reasoning for this line?

shammamah: I wasn't sure if it would automatically convert to int vs. remaining as a float. Fixed in 24b3b9d.

src/components/Input.react.js
```diff
@@ -93,8 +93,7 @@ export default class Input extends PureComponent {
     }
 
     getValueAsNumber(value) {
-        const numericValue = convert(value);
-        return numericValue % 1 === 0 ? Math.floor(numericValue) : numericValue;
+        return convert(value);
```
Marc-Andre-Rivet: At this point why not just call convert directly?

shammamah: Fixed in 21281de!

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   - Added responsiveness on graph parent element resize (previously only worked on window.resize)
   - Added new `dash-graph--pending` class to dcc.Graph, present while resizing, (re-)rendering, loading
 
+### Fixed
+- [#731](https://github.com/plotly/dash-core-components/pull/731) Fixed bug in which input components with type `number` did not correctly update their values.
```
Marc-Andre-Rivet: Copypasta?

```suggestion
- [#730](https://github.com/plotly/dash-core-components/pull/730) Fixed bug in which input components with type `number` did not correctly update their values.
```

shammamah: 😅

PR #731, opened 2020-01-13
Ensure non-clearable dropdowns are actually not clearable. 

### LINKED ISSUE #673: Bug: Dropdowns with clearable=False are still clearable
Dropdown with clearable=False are still clearable by hitting the backspace or del key on the keyboard.

You can replicate the bug here in the Dropdown Clear section:
https://dash.plot.ly/dash-core-components/dropdown

shammamah:
Closes #673.


Marc-Andre-Rivet:
@shammamah I think you can just add the following instead to make it work..

```
                <ReactDropdown
                    clearable={this.props.clearable}
```

shammamah:
@Marc-Andre-Rivet `clearable` is already passed to `ReactDropdown`: https://github.com/plotly/dash-core-components/blob/9bea97b0564a747ec69eabe52bb5aae95970ae44/src/fragments/Dropdown.react.js#L95 -- might be a bug on their end.

PR #732, opened 2020-01-13
Update Plotly.js to 1.52.1

Marc-Andre-Rivet:
From Dash, for cross-testing

https://percy.io/plotly/dash/builds/3749412
https://circleci.com/gh/plotly/dash/tree/plotlyjs-1.52.1


tests/integration/tab/test_tabs_with_graphs.py
```diff
@@ -230,6 +230,7 @@ def check_graph_config_shape(dash_dcc):
         'showSources',
         'logging',
         'globalTransforms',
+        'notifyOnLogging',
```
Marc-Andre-Rivet: @alexcjohnson We already ignore `logging`, seems like we probably want to ignore another `logging` themed prop, but need to be certain.

package.json
```diff
@@ -28,7 +27,7 @@
     "prebuild:js": "cp node_modules/plotly.js/dist/plotly.min.js dash_core_components_base/plotly.min.js",
     "build:js": "webpack --mode production",
     "build:py_and_r": "dash-generate-components ./src/components dash_core_components -p package-info.json && cp dash_core_components_base/** dash_core_components/ && dash-generate-components ./src/components dash_core_components -p package-info.json --r-prefix 'dcc'",
-    "build": "run-s build:js generator build:py_and_r",
```
Marc-Andre-Rivet: Some extra clean up / simplification, this is no longer necessary b/c of the `prebuild` command.

tests/integration/tab/test_tabs_with_graphs.py
alexcjohnson: Yes, for now we can ignore this one. In fact `notifyOnLogging` *would* be useful to dash - and alongside it `logging` becomes useful - because this repeats console log messages into the UI as notification boxes. Unfortunately right now this can only be used globally (`Plotly.setPlotConfig`), not per-plot in the `config` prop. In fact the better solution for dash will be when we have logging events https://github.com/plotly/plotly.js/issues/4462, and we can hopefully pass these along to the main dash devtools.

CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - [#725](https://github.com/plotly/dash-core-components/pull/725) Improve async graph performance by parallelizing resource fetching instead of fetching sequentially
 
 ### Updated
+- [#732](https://github.com/plotly/dash-core-components/pull/732)
+  - Upgraded plotly.js to [1.52.1](https://github.com/plotly/plotly.js/releases/tag/v1.52.1)
+  - [Feature release 1.52.0](https://github.com/plotly/plotly.js/releases/tag/v1.52.0) which contains:
+    - Enable loading locale bundles before plotly.js bundles [#4453](https://github.com/plotly/plotly.js/pull/4453)
+    - `ko` localization [#4315](https://github.com/plotly/plotly.js/pull/4315)
```
Marc-Andre-Rivet: @alexcjohnson Added these based on previous entries (ko) and because of its significance for async/Dash. Feel free to add other items that are relevant.

alexcjohnson: That's fine, good call pointing out the ordering fix. Aside from that, there are no new trace types in 1.52, that's the most important thing to include.

PR #734, opened 2020-01-15
Move Plotly.js from devDeps to deps

Marc-Andre-Rivet:
`Plotly.js` is a required `dependencies` to correctly bundle `dcc.Graph` into another library without making that library responsible for providing some version of `Plotly.js`.


rpkyle:
🚀 💃 🚀

PR #740, opened 2020-01-22
Ensure that components are still in the DOM when they are loading.

### LINKED ISSUE #674: uirevision for mapbox in dcc.Graph breaks when in dcc.Loading
Package version
dash==1.3.1

I’m trying to add {‘uirevision’: True} for a mapbox in dcc.Graph and found out that dcc.Loading() is breaking the behavior of preserving UI, same as mentioned in this community issue https://community.plot.ly/t/dcc-loading-component-breaks-uirevision-for-mapbox-among-others/29098

Below is a reproducible example

```python
from urllib.request import urlopen
import json
import plotly.graph_objects as go
import dash

with urlopen('https://raw.githubusercontent.com/plotly/datasets/master/geojson-counties-fips.json') as response:
    counties = json.load(response)

import pandas as pd

df = pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/fips-unemp-16.csv",
                 dtype={"fips": str})

app = dash.Dash(__name__)
server = app.server

mapbox_token = "pk.eyJ1IjoicGxvdGx5bWFwYm94IiwiYSI6ImNqdnBvNDMyaTAxYzkzeW5ubWdpZ2VjbmMifQ.TXcBE-xg9BFdV2ocecc_7g"

fig = go.Figure(go.Choroplethmapbox(geojson=counties, locations=df.fips, z=df.unemp,
                                    colorscale="Viridis", zmin=0, zmax=12,
                                    marker_opacity=0.5, marker_line_width=0))
fig.update_layout(mapbox_style="carto-positron",
                  mapbox_accesstoken=mapbox_token,
                  mapbox_zoom=3, mapbox_center={"lat": 37.0902, "lon": -95.7129})
fig.update_layout(margin={"r": 0, "t": 0, "l": 0, "b": 0}, uirevision=True)

app.layout = html.Div(
    children=[
        html.Button(id='btn', children='update', n_clicks=0),
        html.Div('Figure No Loading'),
        dcc.Graph(id='no-loading-fig', figure=fig),
        html.Div('Figure Loading'),
        dcc.Loading(dcc.Graph(id='mapbox-loading', figure=fig))]
)


@app.callback(
    [Output('mapbox-loading', 'figure'), Output('no-loading-fig', 'figure')],
    [Input('btn', 'n_clicks')]
)
def update(click):
    if click:
        fig.update_layout(mapbox_style="light")
        return fig, fig
    else:
        raise PreventUpdate


if __name__ == '__main__':
    app.run_server(debug=True, port=8053, dev_tools_hot_reload=False, use_reloader=False)
```

loading

### LINKED ISSUE #684: dcc.Graph in dcc.Loading component isn't working right
See https://github.com/plotly/dash-core-components/issues/672#issuecomment-542784762

> This might be indicative of a deeper rendering problem within the loading component - why are we creating the graph twice?

### LINKED ISSUE #756: Mapbox Graph wrapped in dcc.Loading lags one update behind
Hi there,
when I wrap a Mapbox scatter plot in a dcc.Loading component, updates to this figure seem to be delayed by one update. The same logic with normal scatter plots works fine.  I could reproduce this behavior with Dash Core Components 1.8.0 and previous versions. 

Here's a demo:
withoutLoading
withLoading

and here's the code:
```python
import dash
import plotly.graph_objects as go

app = dash.Dash(__name__)

lat = [10, 10, 10, 10]
lon = [10, 20, 30, 40]

app.layout = html.Div([
    html.Div('Select a point on the map:'),
    dcc.Slider(min=0, max=3, step=1, value=0, marks=[0, 1, 2, 3], id='slider'),
    dcc.Loading(dcc.Graph(id='graphContainer'))
])

@app.callback(Output('graphContainer', 'figure'),
              [Input('slider', 'value')])
def UpdateGraph(value):
    return {'data': [go.Scattermapbox(lat=lat, lon=lon, selectedpoints=[value])],
            'layout': go.Layout(mapbox_style='carto-positron')}


app.run_server()
```

### LINKED ISSUE #780: [BUG ?] dcc.Loading leads to remount after selection change
Hello plotly team,

I am about to add my custom component to dash-bio (https://github.com/plotly/dash-bio/pull/496)

Your colleagues asked me why I commented out dcc.Loading for my component. In fact, it shows an unwanted behaviour (Bug?) As soon as you change the selection in a drop-down the component gets remounted causing me too lose all the previous states...

These are important because the component is a new molecular structure viewer in which one of its most important features is that a structure which has been selected before does not need to be reloaded from the back-end but from the browser. Thus, all already loaded structures have to be saved in the state of the app.

If you want to see it on your own just clone my repo (https://github.com/IvoLeist/dash_ngl) and change in usage.py viewer to dcc.Loading (viewer). In the console you will see after each drop-down change the message "component did mount" indicating that the old one got deleted and recreated (see screenshot below).This is not happening if I do not use dcc.Loading.

If you need a more stripped down version of my component 
which is able to reproduce that issue just let me know.

Best regards,
Ivo

screenshot_dash_ngl

shammamah:
Closes #674.

Also these are all the same thing:
Closes #780 
Closes #756 
Closes #684

## About 
Since the `children` property of the `dcc.Loading` component is not rendered when `loading_state.is_loading` is true, this leads to some information about user interactions being lost when a graph (specifically, a map) is wrapped in the loading component. In this PR, we still render the `children` but with `visibility: hidden`, and display the spinner over top of it. The component is not visible, but it still takes up the same amount of space when it is loading, which avoids issues that have to do with divs shifting up or down during loading if the spinner is smaller or larger than the `children`. 

We might want to instead always return this "wrapper" div that contains both the spinner and the component, and only change the visibility when `loading_state.is_loading` changes. One issue that stems from this is that the component itself would be moved "down" a level in the DOM, which might affect the CSS/layout of different apps.

### Before
Jan-23-2020 15-51-41

### After
Jan-23-2020 15-56-54

### Sample app code 
```python
from urllib.request import urlopen
import json
import pandas as pd
import plotly.graph_objects as go
import dash
from time import sleep


with urlopen('https://raw.githubusercontent.com/plotly/datasets/master/geojson-counties-fips.json') as response:
    counties = json.load(response)


df = pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/fips-unemp-16.csv",
                 dtype={"fips": str})

app = dash.Dash(__name__)
server = app.server

mapbox_token = "pk.eyJ1IjoicGxvdGx5bWFwYm94IiwiYSI6ImNqdnBvNDMyaTAxYzkzeW5ubWdpZ2VjbmMifQ.TXcBE-xg9BFdV2ocecc_7g"

fig = go.Figure(go.Choroplethmapbox(geojson=counties, locations=df.fips, z=df.unemp,
                                    colorscale="Viridis", zmin=0, zmax=12,
                                    marker_opacity=0.5, marker_line_width=0))
fig.update_layout(mapbox_style="carto-positron",
                  mapbox_accesstoken=mapbox_token,
                  mapbox_zoom=3, mapbox_center={"lat": 37.0902, "lon": -95.7129})
fig.update_layout(margin={"r": 0, "t": 0, "l": 0, "b": 0}, uirevision=True)

app.layout = html.Div(
    children=[
        html.Button(id='btn', children='update', n_clicks=0),
        dcc.Loading(dcc.Graph(id='mapbox-loading', figure=fig), type='default'),
        html.Div(style={'backgroundColor': 'pink'}, children='hello')]
)


@app.callback(
    Output('mapbox-loading', 'figure'),
    [Input('btn', 'n_clicks')]
)
def update(click):
    if click is None or click == 0:
        raise PreventUpdate
    sleep(2)
    fig.update_layout(mapbox_style="light")
    return fig


if __name__ == '__main__':
    app.run_server(debug=True, port=8053)
```


tests/unit/Loading.test.js
```diff
@@ -14,7 +14,12 @@ test('Loading renders', () => {
         </Loading>
     );
 
-    expect(loading.html()).toMatchSnapshot('Loading with is_loading=true');
+    expect(
+        loading
+            .find('.dash-spinner')
+            .parent()
+            .html()
+    ).toMatchSnapshot('Loading with is_loading=true');
```
Marc-Andre-Rivet: I don't see the corresponding snapshot baseline in this PR

shammamah: This snapshot is already here: https://github.com/plotly/dash-core-components/blob/803b67b5f234aa70f8e536f7c93bd730f8680978/tests/unit/__snapshots__/Loading.test.js.snap#L224

The baseline is the same, but the test itself had to change because I moved things "down" in the DOM.

shammamah:
Running into some strange test failures. 

This test https://github.com/plotly/dash-core-components/blob/b35a560712b74a1bccdf4ca1627c6a3095584b60/tests/integration/graph/test_graph_basics.py#L36 fails for both values of `is_eager` on the CI machine. Upon running the tests locally, they both pass. 

The versions of Python, all of the Python packages, and `chromedriver` are the same on my computer and the CI machine.

Removing this line https://github.com/plotly/dash-core-components/blob/b35a560712b74a1bccdf4ca1627c6a3095584b60/tests/integration/graph/test_graph_basics.py#L88 allows for `is_eager=True` to pass on the CI machine.

Relevant CI run: https://circleci.com/gh/plotly/dash-core-components/17161#tests/containers/1

alexcjohnson:
Some tests are currently failing because I didn't put `className` onto the wrapper div when loading is done, I only put it on the spinner itself, when the spinner is present. Is it useful for this class to be applied to the wrapper? At first glance I'd think this is actually unwanted, because then you can't use this class to target the spinner.

pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+testpaths = tests/
```
alexcjohnson, line 2: having this line cuts test discovery time locally from ~10 seconds to <1sec - particularly nice when calling `pytest -k ...`

I didn't pay attention to the other lines, just copied from `dash`

tests/unit/Loading.test.js
```diff
@@ -23,7 +28,7 @@ test('Loading renders without loading_state', () => {
         </Loading>
     );
 
-    expect(loading.html()).toEqual('<div>Loading is done!</div>');
+    expect(loading.html()).toEqual('<div>Loading is done!</div><div></div>');
```
alexcjohnson, line 31: The spinner now leaves its wrapper div around even after it disappears. I did it this way to ensure React had no chance of being confused about the identity of the real children components - loading or not doesn't change the immediate children of the outer containing div at all. Maybe overly paranoid, but didn't seem like it would hurt, right?

tests/integration/store/test_component_props.py
```diff
@@ -4,17 +4,18 @@
...
 
-def test_stcp100_clear_data_on_all_types(store_app, dash_dcc):
+def test_stcp001_clear_data_on_all_types(store_app, dash_dcc):
     dash_dcc.start_server(store_app)
 
     assert dash_dcc.wait_for_contains_text("#output", store_app.uuid)
 
     dash_dcc.multiple_click("#btn", 3)
-    assert dash_dcc.get_local_storage() == {"n_clicks": 3}
+    wait.until(lambda: dash_dcc.get_local_storage() == {"n_clicks": 3}, timeout=1)
```
alexcjohnson, line 18: As a bare `assert` this gave spurious errors.

tests/integration/loading/test_loading_component.py
```diff
@@ -0,0 +1,268 @@
...
+        dash_dcc.find_element(".loading .dash-spinner")
+        dash_dcc.find_element("#graph .js-plotly-plot")
+        dash_dcc.driver.execute_script(
+            "window.gd = document.querySelector('.js-plotly-plot');"
+            "window.gd.__test__ = 'boo';"
+        )
+
+    test_identity = (
+        "var gd_ = document.querySelector('.js-plotly-plot');"
+        "return gd_ === window.gd && gd_.__test__ === 'boo';"
```
alexcjohnson, line 255: `gd_.__test__ === 'boo'` should be redundant after we already have `gd_ === window.gd` to show the DOM element is unchanged, just being paranoid again...

pytest.ini
Marc-Andre-Rivet: 🎉

src/components/Loading.react.js
```diff
@@ -37,27 +49,26 @@ export default class Loading extends Component {
...
-                    fullscreen={fullscreen}
-                />
-            );
-        }
-
-        if (
-            type(this.props.children) !== 'Object' ||
-            type(this.props.children) !== 'Function'
-        ) {
-            return <div className={className}>{this.props.children}</div>;
```
chriddyp, line 58: Yeah, this part, right? Agreed, it isn't clear to me why this was there.

tests/integration/loading/test_loading_component.py
```diff
@@ -0,0 +1,268 @@
...
+
+    app = dash.Dash(__name__)
+
+    app.layout = html.Div(
+        [dcc.Loading([html.Div(id="div-1")], className="loading")], id="root"
+    )
+
+    @app.callback(Output("div-1", "children"), [Input("root", "n_clicks")])
+    def updateDiv(children):
+        with lock:
```
chriddyp, line 20: wow very nice 💯

tests/integration/loading/test_loading_component.py
```diff
@@ -0,0 +1,268 @@
...
+            bars = list(range(2, (n or 0) + 5))
+            return {
+                "data": [{"type": "bar", "x": bars, "y": bars}],
+                "layout": {"width": 400, "height": 400}
+            }
+
+    with lock:
+        dash_dcc.start_server(app)
+        dash_dcc.find_element(".loading .dash-spinner")
+        dash_dcc.find_element("#graph .js-plotly-plot")
```
Marc-Andre-Rivet, line 247: Could we also validate that the graph isn't visible?

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Changed
+- [#740](https://github.com/plotly/dash-core-components/pull/740) Keep components that are loading in the DOM, but not visible, as opposed to removing them entirely
```
chriddyp, line 7:
```suggestion
- [#740](https://github.com/plotly/dash-core-components/pull/740) Keep components that are loading in the DOM, but not visible, as opposed to removing them entirely. This will ensure that the size of the component's container does not shrink or expand when the component goes into the loading state.
```

tests/integration/loading/test_loading_component.py
alexcjohnson: Only the last test here (006) is new, the others I just pulled out of [test_integration_1.py](https://github.com/plotly/dash-core-components/pull/740/files#diff-47574444ebc5cb78be0c75fa212cc77a) into the `dash.testing` format before fixing them for the `className` update

tests/integration/loading/test_loading_component.py
```diff
@@ -0,0 +1,268 @@
...
+    assert len(dash_dcc.find_elements('.js-plotly-plot .bars path')) == 3
+    assert dash_dcc.driver.execute_script(test_identity)
+
+    with lock:
+        dash_dcc.find_element("#btn").click()
+        dash_dcc.find_element(".loading .dash-spinner")
+        assert len(dash_dcc.find_elements('.js-plotly-plot .bars path')) == 3
+        assert dash_dcc.driver.execute_script(test_identity)
+
+    assert len(dash_dcc.find_elements('.js-plotly-plot .bars path')) == 4
```
Marc-Andre-Rivet, line 267: To go along with the graph not being visible while loading, can we make sure it is visible now.

tests/integration/loading/test_loading_component.py
alexcjohnson: I did that with some of the [simpler items eg](https://github.com/plotly/dash-core-components/pull/740/files#diff-823b747b0b592ad60f40a9c017818677R219):
```
dash_dcc.wait_for_text_to_equal("#btn-3", "")
```
it looks like it's empty even though it actually still has text in it. Not sure exactly how to do that for a graph though - I suppose I could look directly at the `visibility` style on the element that sets it to `hidden`.

Marc-Andre-Rivet: If equivalent, all good :D

tests/integration/loading/test_loading_component.py
Marc-Andre-Rivet: @HammadTheOne, @harryturr When writing tests involving callbacks, using a `Lock` like ☝️ is the best practice. Gives you full control of when the callback and your test conditions will be run in respect to one another.

tests/integration/loading/test_loading_component.py
alexcjohnson: added `getComputedStyle` visibility tests in https://github.com/plotly/dash-core-components/pull/740/commits/4fc809a6c1328f922c0eddff315adc1bf79bd680

Marc-Andre-Rivet:
Seems fine in IE11, we seem to be using the part of flex that actually works
image

tests/integration/loading/test_loading_component.py
```diff
@@ -249,20 +257,24 @@ def update_graph(n):
...
     assert dash_dcc.driver.execute_script(test_identity)
+    assert get_graph_visibility() == "visible"
 
     with lock:
         dash_dcc.find_element("#btn").click()
         dash_dcc.find_element(".loading .dash-spinner")
-        assert len(dash_dcc.find_elements('.js-plotly-plot .bars path')) == 3
+        assert len(dash_dcc.find_elements(".js-plotly-plot .bars path")) == 3
         assert dash_dcc.driver.execute_script(test_identity)
+        assert get_graph_visibility() == "hidden"
```
Marc-Andre-Rivet, line 276: 🙇

PR #753, opened 2020-02-07
Separate CSS file on build

### LINKED ISSUE #752: Inline CSS vs. build to separate CSS file
CSP ([Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy)) is a good tool for enhancing security of web applications. [Mozilla Observatory](https://observatory.mozilla.org/analyze/observatory.mozilla.org) is a good place to check for CSP implementation across web applications. How easy it is to enforce strict CSP settings depends, to a large degree, on the frameworks used in the stack.

Dash out of the box is quite CSP settings friendly, e.g. you can do `pip install dash flask-talisman` (alternatively set the CSP headers directly instead of using [flask-talisman](https://github.com/GoogleCloudPlatform/flask-talisman)) and then run e.g. 
```
import dash

from flask_talisman import Talisman

app = dash.Dash(__name__)

CSP = {
    "default-src": "'self'",
    "script-src": [
        "'self'",
        # Due to https://github.com/plotly/dash/issues/630:
        "'sha256-jZlsGVOhUAIcH+4PVs7QuGZkthRMgvT2n0ilH6/zTM0='",
    ]
}

Talisman(app.server, content_security_policy=CSP, force_https=False)

app.layout = html.Div(children=["Hello Dash!"])

if __name__ == "__main__":
    app.run_server()
```
This will work with no CSP errors on localhost in the browser console - despite quite strict CSP settings.

If however `import dash_core_components as dcc` is added, you will need to either add
```python
    "style-src": ["'self'", "'unsafe-inline'"],
```
or
```python
    "style-src": [
        "'sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='",
        "'sha256-cJ5ObsneXZnEJ3xXl8hpmaM9RbOwaU6UIPslFf0/OOE='",
        "'sha256-joa3+JBk4cwm5iLfizxl9fClObT5tKuZMFQ9B2FZlMg='",
        "'sha256-Jtp4i67c7nADbaBCKBYgqI86LQOVcoZ/khTKsuEkugc='",
        "'sha256-MoQFjm6Ko0VvKSJqNGTl4e5H3guejX+CG/LxytSdnYg='",
        "'sha256-kkTbqhXgCW2olAc5oz4XYZtzEwh5ojnaxt9iOTAZcZo='",
    ],
```
to the `CSP` dictionary in order to allow the CSS that `dcc` is adding inline to the app.

The first is not optimal as you then again open for [inline CSS "XSS"](https://stackoverflow.com/questions/30653698/csp-style-src-unsafe-inline-is-it-worth-it) (as when not using CSP at all). The second is not optimal either as the hashes will need to update each time a new version of `dcc` changes its inline style content.

If `dcc` could output CSS as a separate file during build, instead of inline style, we could with Dash also enforce strict CSS CSP. I.e. use [`mini-css-extract-plugin`](https://webpack.js.org/plugins/mini-css-extract-plugin/) instead of `style-load` in `webpack.config.js` during (production?) build of `dcc`.

anders-kiaer:
This is a draft/sketch PR of how `dash-core-components` perhaps could build CSS into a separate `.css` file. 

Resolves #752, and enables e.g. running
```python
import dash

from flask_talisman import Talisman

app = dash.Dash(__name__)

csp = {
    "default-src": "'self'",
    "script-src": [
        "'self'",
        # Due to https://github.com/plotly/dash/issues/630:
        "'sha256-jZlsGVOhUAIcH+4PVs7QuGZkthRMgvT2n0ilH6/zTM0='",
    ]
}

Talisman(app.server, content_security_policy=csp, force_https=False)

app.layout = html.Div(children=["Hello Dash!"])

if __name__ == "__main__":
    app.run_server()
```


webpack.config.js
```diff
@@ -131,11 +134,14 @@ module.exports = (env, argv) => {
             }
         },
         plugins: [
-            new WebpackDashDynamicImport(),
+            new MiniCssExtractPlugin({
+                filename: `${dashLibraryName}.css`,
+            }),
+            //new WebpackDashDynamicImport(),
```
anders-kiaer: Had to comment this one out in order to be able to run `npm run build`, but I guess someone knowing the details of `WebpackDashDynamicImport` would be able to figure out how to include it again.

anders-kiaer:
Did another quick test, with this PR, to check if "inline style logic from Python" works with strict CSS CSP. 

The result was that it works, as might be expected since Dash probably pass the `style` dictionaries from Python to React (which again sets the style attributes in a CSP compatible way with `react >= 15`).

```python
import dash

from flask_talisman import Talisman

app = dash.Dash(__name__)

CSP = {
    "default-src": "'self'",
    "script-src": ["'self'", "'sha256-jZlsGVOhUAIcH+4PVs7QuGZkthRMgvT2n0ilH6/zTM0='"],
    "style-src": "'self'",
}

Talisman(app.server, content_security_policy=CSP, force_https=False)

app.layout = dcc.RadioItems(
    id="radioitems",
    options=[{"label": color, "value": color} for color in ["red", "blue", "green"]],
)

@app.callback(
    Output("radioitems", "style"), [Input("radioitems", "value")],
)
def _update_style(value):
    return {"backgroundColor": value}

if __name__ == "__main__":
    app.run_server()
```
**Result:**
image
image
image
image

Issue #742, opened 2020-01-24
Console warnings when using `dcc.Graph` on `dash==1.8` with `debug=True`


anders-kiaer:
Running some small example like
```python
# -*- coding: utf-8 -*-
import dash
import dash_core_components as dcc
import dash_html_components as html

app = dash.Dash(__name__)

app.layout = html.Div(children=[
    dcc.Graph(
        id='example-graph',
        figure={
            'data': [
                {'x': [1, 2, 3], 'y': [4, 1, 2], 'type': 'bar', 'name': 'SF'},
                {'x': [1, 2, 3], 'y': [2, 4, 5], 'type': 'bar', 'name': u'Montréal'},
            ],
            'layout': {
                'title': 'Dash Data Visualization'
            }
        }
    )
])

if __name__ == '__main__':
    app.run_server(debug=True)
```
returns console warnings in the browser (Chromium Version 79.0.3945.88):
```
Warning: Failed prop type: The prop `isRequired` is marked as required in `<<anonymous>>`, but its value is `undefined`.
    in Unknown (created by t)
    in t (created by CheckedComponent)
    in CheckedComponent (created by TreeContainer)
    in UnconnectedComponentErrorBoundary (created by Connect(UnconnectedComponentErrorBoundary))
    in Connect(UnconnectedComponentErrorBoundary) (created by TreeContainer)
    in TreeContainer (created by Connect(TreeContainer))
    in Connect(TreeContainer) (created by TreeContainer)
    in div (created by u)
    in u (created by CheckedComponent)
    in CheckedComponent (created by TreeContainer)
    in UnconnectedComponentErrorBoundary (created by Connect(UnconnectedComponentErrorBoundary))
    in Connect(UnconnectedComponentErrorBoundary) (created by TreeContainer)
    in TreeContainer (created by Connect(TreeContainer))
    in Connect(TreeContainer) (created by UnconnectedContainer)
    in div (created by UnconnectedGlobalErrorContainer)
    in div (created by GlobalErrorOverlay)
    in div (created by GlobalErrorOverlay)
    in GlobalErrorOverlay (created by DebugMenu)
    in div (created by DebugMenu)
    in DebugMenu (created by UnconnectedGlobalErrorContainer)
    in div (created by UnconnectedGlobalErrorContainer)
    in UnconnectedGlobalErrorContainer (created by Connect(UnconnectedGlobalErrorContainer))
[...]

PR #754, opened 2020-02-11
full propTypes for ControlledPlotlyGraph to avoid weird React warning

alexcjohnson:
@Marc-Andre-Rivet this appears to remove the warning I describe in https://github.com/plotly/dash-core-components/pull/706/files#r377687277 when showing graphs with debug mode

```
The prop `isRequired` is marked as required in `<<anonymous>>`,
but its value is `undefined`
```

### DIFF
--- a/src/components/Graph.react.js
+++ b/src/components/Graph.react.js
@@ -115,8 +115,6 @@ const ControlledPlotlyGraph = memo(props => {
     );
 });
 
-ControlledPlotlyGraph.propTypes = PropTypes.any;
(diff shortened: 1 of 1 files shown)


anders-kiaer:
I guess this will close #742 and https://github.com/plotly/dash/issues/1101 👍

alexcjohnson:
Ah indeed, thanks @anders-kiaer - I had been so focused on my own project, hadn't even noticed those two issues!

PR #760, opened 2020-02-25
Add R examples to dashCoreComponents

rpkyle:
This PR proposes to contribute a complete set of R examples for the entire suite of `dashCoreComponents`. 

The majority of the examples included here mirror those found in the online documentation available at https://dashr.plot.ly, and are intended as as complement to those already in active use.

Since accent marks may cause headaches depending on the encoding of the system on which R is used, I have attempted to remove them from the corresponding `dash-info.yaml`.

### DIFF
--- a/dash-info.yaml
+++ b/dash-info.yaml
@@ -35,6 +35,7 @@ r_examples:
             library(dash)
             library(dashHtmlComponents)
             library(dashCoreComponents)
+            library(plotly)
 
             app <- Dash$new()
 
--- a/dash-info.yaml
+++ b/dash-info.yaml
@@ -651,6 +651,7 @@ r_examples:
             library(dash)
             library(dashCoreComponents)
             library(plotly)
+            app <- Dash$new()
             
             year <- c(1995, 1996, 1997, 1998, 1999, 2000, 2001, 2002, 2003,
               2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012)
--- a/dash-info.yaml
+++ b/dash-info.yaml
@@ -200,9 +200,7 @@ r_examples:
             library(dashCoreComponents)
             library(dashHtmlComponents)
             library(dash)
-            
             app <- Dash$new()
-            
             app$layout(htmlDiv(list(
               # The memory store reverts to the default on every page refresh
               dccStore(id='memory'),
@@ -215,7 +213,7 @@ r_examples:
               dccStore(id='session', storage_type='session'),
               htmlTable(list(
                 htmlThead(list(
-                  htmlTr(htmlTh('Click to store in:', colSpan="3")),
+                  htmlTr(htmlTh('Click to store in:', colSpan='3')),
                   htmlTr(list(
                     htmlTh(htmlButton('memory', id='memory-button')),
                     htmlTh(htmlButton('localStorage', id='local-button')),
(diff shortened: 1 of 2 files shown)


dash-info.yaml
```diff
@@ -6,3 +6,782 @@ pkg_help_description: >-
...
+            )
+
+            app$run_server()
+    - name: dccInterval
+      dontrun: TRUE 
+      code: |
+            library(dash)
+            library(dashHtmlComponents)
+            library(dashCoreComponents)
+
```
HammadTheOne:
```suggestion
library(plotly)

```

dash-info.yaml
```diff
@@ -6,3 +6,782 @@ pkg_help_description: >-
...
+            
+            
+            app$callback(
+              output = list(id = 'output-confirm1', property = 'children'),
+              params=list(input(id = 'confirm', property = 'submit_n_clicks')),
+              function(n_clicks, value) {
+                if(length(n_clicks) == FALSE){
+                  sprintf('It wasnt easy but we did it %s', str(n_clicks))
+                }
+              })
```
HammadTheOne: It seems like this callback isn't behaving the way it should, I wasn't able to populate the children component.

dash-info.yaml
```diff
@@ -6,3 +6,782 @@ pkg_help_description: >-
...
+            )
+            
+            app$callback(
+              output = list(id = 'session-clicks', property = 'children'),
+              params = list(input(id = 'session', property = 'modified_timestamp'),
+                            state(id = 'session', property = 'data')),
+              returnStoreData
+            )
+            
+            app$run_server()
```
HammadTheOne:
```suggestion
library(dashCoreComponents)
library(dashHtmlComponents)
library(dash)

app <- Dash$new()

app$layout(htmlDiv(list(
  # The memory store reverts to the default on every page refresh
  dccStore(id='memory'),
  # The local store will take the initial data
  # only the first time the page is loaded
  # and keep it until it is cleared.
  dccStore(id='local', storage_type='local'),
  # Same as the local store but will lose the data
  # when the browser/tab closes.
  dccStore(id='session', storage_type='session'),
  htmlTable(list(
    htmlThead(list(
      htmlTr(htmlTh('Click to store in:', colSpan='3')),
      htmlTr(list(
        htmlTh(htmlButton('memory', id='memory-button')),
        htmlTh(htmlButton('localStorage', id='local-button')),
        htmlTh(htmlButton('sessionStorage', id='session-button'))
      )),
      htmlTr(list(
        htmlTh('Memory clicks'),
        htmlTh('Local clicks'),
        htmlTh('Session clicks')
      ))
    )),
    htmlTbody(list(
      htmlTr(list(
        htmlTd(0, id='memory-clicks'),
        htmlTd(0, id='local-clicks'),
        htmlTd(0, id='session-clicks')
      ))
    ))
  ))
)))

for (i in c('memory', 'local', 'session')) {
  app$callback(
    output(id = i, property = 'data'),
    params = list(
      input(id = sprintf('%s-button', i), property = 'n_clicks'),
      state(id = i, property = 'data')
    ),
    function(n_clicks, data){
      if(is.null(n_clicks)){
        return()
      }
      
      if(is.null(data[[1]])){
        data = list('clicks' = 0)
      } else{
        data = data
      }  
      data['clicks'] = data$clicks + 1
      return(data)
    }
  )
}

for (i in c('memory', 'local', 'session')) {
  app$callback(
    output(id = sprintf('%s-clicks', i), property = 'children'),
    params = list(
      input(id = i, property = 'modified_timestamp'),
      state(id = i, property = 'data')
    ),
    function(ts, data){
      if(is.null(ts)){
        return()
      }
      if(is.null(data[[1]])){
        data = list()
      } else{
        data = data
      }  
      return(data$clicks[[1]])
    }
  )
}

app$run_server()
```
Changed this to a more concise example based on plotly/dash-docs#793

dash-info.yaml
```diff
@@ -6,3 +6,782 @@ pkg_help_description: >-
...
+                            state(id = 'upload-image', property = 'last_modified')),
+              function(list_of_contents, list_of_names, list_of_dates){
+                if (!is.null(list_of_contents) && !is.null(list_of_names) && !is.null(list_of_dates)) {
+                  children = lapply(1:length(list_of_contents), function(x){
+                    parse_content(list_of_contents[[x]], list_of_names[[x]], list_of_dates[[x]])
+                  })
+                }
+                return(children)
+              }
+            )
```
HammadTheOne:
```suggestion
            app$callback(
              output = list(id='output-image-upload', property = 'children'),
              params = list(input(id = 'upload-image', property = 'contents'),
                            state(id = 'upload-image', property = 'filename'),
                            state(id = 'upload-image', property = 'last_modified')),
              function(list_of_contents, list_of_names, list_of_dates){
                if (!is.null(list_of_contents) && !is.null(list_of_names) && !is.null(list_of_dates[[1)) {
                  children = lapply(1:length(list_of_contents), function(x){
                    parse_content(list_of_contents[[x]], list_of_names[[x]], list_of_dates[[x]])
                  })
                }
                else {
                  children = "Upload a file to see the raw data."
                }
                return(children)
              }
            )
```

dash-info.yaml
```diff
@@ -6,3 +6,782 @@ pkg_help_description: >-
...
+            
+            app$run_server()
+    - name: dccGraph
+      dontrun: TRUE 
+      code: |
+            library(dash)
+            library(dashCoreComponents)
+            library(plotly)
+
```
HammadTheOne:
```suggestion
            app <- Dash$new()
            
```

dash-info.yaml
rpkyle: Yeah, good observation -- I think we'll chop this callback out, it's not really necessary here.

rpkyle:
```suggestion
```

PR #762, opened 2020-02-27
Issue 745 - Rename async modules

### LINKED ISSUE #745: Rename async modules
My company uses a centralized access management system to validate a user is authorized to view websites. The system rejects any requests going to URL's containing the tilde character. This prevents all dependencies with names like async~xxx from being downloaded when the browser requests them.

Is it possible to replace the tilde with a different character such as underscore ( _ )?

Marc-Andre-Rivet:
Closes #745

### DIFF
--- a/dash_core_components_base/__init__.py
+++ b/dash_core_components_base/__init__.py
@@ -52,20 +52,20 @@ async_resources = [
 _js_dist = []
 
 _js_dist.extend([{
-        'relative_package_path': 'async~{}.js'.format(async_resource),
+        'relative_package_path': 'async-{}.js'.format(async_resource),
         'external_url': (
             'https://unpkg.com/dash-core-components@{}'
-            '/dash_core_components/async~{}.js'
+            '/dash_core_components/async-{}.js'
         ).format(__version__, async_resource),
         'namespace': 'dash_core_components',
         'async': True
     } for async_resource in async_resources])
 
 _js_dist.extend([{
-        'relative_package_path': 'async~{}.js.map'.format(async_resource),
+        'relative_package_path': 'async-{}.js.map'.format(async_resource),
         'external_url': (
             'https://unpkg.com/dash-core-components@{}'
-            '/dash_core_components/async~{}.js.map'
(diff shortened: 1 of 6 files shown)


dash_core_components_base/__init__.py
```diff
@@ -116,19 +116,19 @@
         'async': 'eager'
     },
     {
-        'relative_package_path': 'async~plotlyjs.js',
+        'relative_package_path': 'async-plotlyjs.js',
         'external_url': (
             'https://unpkg.com/dash-core-components@{}'
-            '/dash_core_components/async~graph~plotlyjs.js'
+            '/dash_core_components/async-plotlyjs.js'
```
Marc-Andre-Rivet: Pretty bad error here and below with `serve_locally=False`

alexcjohnson: Yikes, nice find! Is there anything we can do to test this? I'm thinking like a standard test we can put in `dash.testing` and then run in all our repos with appropriate parameters, looping over `_js_dist` and ensuring `external_url` matches a certain pattern vs `relative_package_path`?

alexcjohnson: a followup issue would be fine for now.

Marc-Andre-Rivet: Follow up in https://github.com/plotly/dash-core/issues/171 - providing at least two approaches: pattern matching as suggested and something more robust / matching e2e usage but requiring more effort.

ManojSN:
Team, Thank you for changing the ~ to a -. The use @ sign in the URLS is may also not be considered safe and be blocked. 

Would it be possible to alter that too?

 'external_url': (
            'https://unpkg.com/dash-core-components@{}'
            '/dash_core_components/dash_core_components.min.js'
        ).format(__version__)

alexcjohnson:
> The use @ sign in the URLS is may also not be considered safe and be blocked.

That's just the way https://unpkg.com/ works, we have no control over that. But your app will only use access those URLs if you explicitly set `serve_locally=False`

PR #765, opened 2020-03-02
Fix issues with multiple listeners on URL change

AHMatsWillemsen:
If two callbacks listen to two different parts of the URL, currently this means that both pages are going to be refreshed. This change fixes that by only setting those properties that have actually changed.

Following script is an example of the issue:

```python
import dash

app = dash.Dash(__name__)

app.layout = html.Div([
    dcc.Location(
        id='location',
        refresh=False
    ),
    html.Div(id='output'),
    html.Div(id='output2')
])


@app.callback(Output('output', 'children'), [Input('location', 'pathname')])
def display_output(pathname):
    print(f"PATHNAME CALLBACK LOGGED: {pathname}")
    return dcc.Link(href='?test=test', children=[html.H2("Hey")], refresh=False)

@app.callback(Output('output2', 'children'), [Input('location', 'search')])
def display_output_search(search):
    print(f"SEARCH CALLBACK LOGGED: {search}")
    return "Hey"

if __name__ == '__main__':
    app.run_server(debug=True)
```

### DIFF
--- a/src/components/Location.react.js
+++ b/src/components/Location.react.js
@@ -84,15 +84,20 @@ export default class Location extends Component {
         }
     }
 
-    onLocationChange() {
-        const {setProps} = this.props;
-        setProps({
-            pathname: window.location.pathname,
-            href: window.location.href,
-            hash: window.location.hash,
-            search: window.location.search,
-        });
-
+    onLocationChange() {     
+        const {hash, href, pathname, refresh, search, setProps} = this.props;
+        if(search != window.location.search) {
+            setProps({search: window.location.search})
+        }
+        if(pathname != window.location.pathname) {
+            setProps({pathname: window.location.pathname})
+        }
+        if(hash != window.location.hash) {
+            setProps({hash: window.location.hash})
+        }
+        if(href != window.location.href) {
+            setProps({href: window.location.href})
+        } 
         History.dispatchChangeEvent();
     }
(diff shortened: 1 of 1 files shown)


alexcjohnson:
Thanks @AHMatsWillemsen! Very good point about only setting the props that changed. We should still combine them into one `setProps` call though. That may be what's responsible for the `test-legacy` failures. The `lint-unit` failures you can fix with `npm run format` (after the functionality is adapted).

Then we'll need a test - you can use the app you show above, with a couple of modifications:
- no f-strings unfortunately, we still need the tests to run in both py2 and py3
- and it's easiest if you can get the modified behavior onto the page rather than in the server logs, so you can use eg `dash_duo.wait_for_text_to_equal` - looks like we don't have integration tests for `Location` yet but the other components give good examples - see eg [tests/integration/dropdown/test_clearable_false.py](https://github.com/plotly/dash-core-components/blob/dev/tests/integration/dropdown/test_clearable_false.py)

Lastly we'll need a CHANGELOG entry.

alexcjohnson:
@AHMatsWillemsen this issue was noticed independently by our team, and a new PR #774 opened to resolve it. I've asked @HammadTheOne to use your test app above to inform the test case he's working on, but I'm going to close this PR. Thanks for your contribution!

PR #767, opened 2020-03-06
Added click modifiers, target prop

### LINKED ISSUE #748: Control-click is mishandled
On the [user guide](https://dash.plot.ly/), middle-clicking on any link will open it in a new tab. However, Control-clicking will just open the link in the current tab, _if it's internal_. If it's an _external_ link, then it will behave correctly and open in a new tab.

For example, on the [installation page](https://dash.plot.ly/installation), control-clicking on the "make your first Dash app" will open in the current tab, but control-clicking any of the "changelog" links will open in a background tab.

Using Firefox 72.0.1 on Linux.

### LINKED ISSUE #770: Add `target='_blank'` support to dcc.Link
In HTML, an anchor can be made to open a new tab on click like so:
```
<a href='my-url' target='_blank'>My URL</a>
```

Valid values are `_blank`, `_self` (default), `_parent`, `_top` and `<frame name>`.

If defined and not `_self`, the click event handler will need to be removed for the behavior to be as expected (https://github.com/plotly/dash-core-components/blob/dev/src/components/Link.react.js#L72)

HammadTheOne:
This PR addresses #748 and adds the ability to control+click a `Link` to open it in a new tab. Although `html.A` is a better component to use for external links, this adds that functionality for people who might use it within a multi-page Dash app.

This PR also adds in the `target` prop to add logic that one would expect from a traditional link. 

Closes #748 .
Closes #770


chriddyp:
Awesome! Could we do a little research on other single page app link components on GitHub and see if we are missing any other logic? We really want this link to look and feel like a link. 

This article describes this a bit more: http://jgaskins.org/blog/2016/07/30/don-t-change-perceived-browser-functionality. It would be great to check out the source code of a popular component on GitHub that has gone through all of these different edge cases: 😸 

From http://jgaskins.org/blog/2016/07/30/don-t-change-perceived-browser-functionality, they recommend this:

```
function handleClick(event) {
  var hasModifiers = (
    event.metaKey ||
    event.shiftKey ||
    event.altKey ||
    event.ctrlKey
  );

  // Only handle unmodified left click. Leave everything else alone.
  if(hasModifiers || event.button !== 1) return;

  event.preventDefault(); // Only prevent AFTER confirming you should handle this.
};
```

HammadTheOne:
That's a good point Chris. Added the other modifiers for now, I'll look for another similar component and see if we're missing anything else that might be nice to have.

chriddyp:
Handling "enter" on focussed elements (you can "select" links by "tabbing" through the page) might be another one to consider. We might already handle this one effectively though.

HammadTheOne:
Added `dcc.Link` Title prop in #768 .

HammadTheOne:
Added in the `external` prop, of type `bool`. When set to `True`, the link will open in a new tab. While most Dash apps run in single-page, this combined with the other fixes to `dcc.Location` means a person can open up a `dcc.Link` in a multi-page app to a specific section of a chapter or page in a new tab. By default it is false, so it isn't a breaking change.

src/components/Link.react.js
```diff
@@ -81,6 +87,7 @@ export default class Link extends Component {
                 href={href}
                 onClick={e => this.updateLocation(e)}
                 title={title}
+                target={external ? '_blank' : '_self'}
```
chriddyp: I think that I'd prefer that we just stick to the HTML convention here instead of defining a new API. That way, this component is more or less 1-1 with `html.A`. So, just let users pass through `target` directly rather than defining a new `external` property.

HammadTheOne: I've seen a couple of other link components define external as a separate property, but I think you raise a good point, for both maintaining parity with `html.A` and offering a bit more freedom for the user to decide what to target. 

Changed in 6514c83.

alexcjohnson: > I think that I'd prefer that we just stick to the HTML convention here

I agree @chriddyp - cc @Marc-Andre-Rivet - we had a similar discussion a couple of days ago, also because there *are* other values of `target` that still matter.

Issue #769, opened 2020-03-10
dcc.Location should not fire callback for pathname if the href changed


chriddyp:
Reproducible eaxmple:

```python
import dash
from dash.dependencies import Input, Output
import dash_core_components as dcc
import dash_html_components as html
import datetime
import numpy as np
print(np.__version__)

app = dash.Dash(__name__)

app.layout = html.Div([
    dcc.Location(id='location'),

    html.A('Anchor Link 1', href='#div'),
    html.Span(' - '),
    html.A('Anchor Link 2', href='#div-2'),

    html.Div(id='div'),
    html.Div(id='div-2')
])


@app.callback(Output('div', 'children'), [Input('location', 'pathname')])
def update_path(path):
    print(path)
    return str(datetime.datetime.now())


app.run_server(debug=True)
```

Clicking between the links fires the callback. The callback should not be fired because the `pathname` did not change, only the `href` changed.

It looks like we already have logic that detects _which_ property changed: https://github.com/plotly/dash-core-components/blob/b35a560712b74a1bccdf4ca1627c6a3095584b60/src/components/Location.react.js#L60-L70

I suspect that we just need to apply that logic to the `setProps` call below: https://github.com/plotly/dash-core-components/blob/b35a560712b74a1bccdf4ca1627c6a3095584b60/src/components/Location.react.js#L87-L97

A fix for this is needed in order to provide anchor links in a sidebar in the dash docs. I'm working on this in https://github.com/plotly/dash-docs/pull/802.

cc @Marc-Andre-Rivet  - This might be another nice one for @HammadTheOne ⚾️

chriddyp:
(Related https://github.com/plotly/dash-core/issues/89 & https://github.com/plotly/dash-docs/pull/735, but https://github.com/plotly/dash-docs/pull/802 is the main PR that has its own self-contained sidebar with anchor links)

AHMatsWillemsen:
Hi, thanks for this! We found this bug while developing a multi-page app with one part listening for the pathname changes, and one part for the search-changes, resulting in double callbacks for the search callback. In what version do you expect this fix to be included? We can then remove a custom component.

chriddyp:
@HammadTheOne  - this should be closed, right?

HammadTheOne:
Yup, forgot to put the `closes` tag in the PR, but this is merged in.

PR #774, opened 2020-03-11
dcc.Location props fix for callbacks

HammadTheOne:
As pointed out in #769, `dcc.Location` was firing callbacks for the `pathname` when the `href` was changed. This behaviour carried over to all props, including `hash` and `search`. This PR fixes that behaviour by adding in conditionals to check whether the new value is different from the previous one. 
 
To Do:

- [x] Add a basic integration test
- [x] Update changelog with PR tag and link.


alexcjohnson:
Nice! supersedes community PR #765 - take a look at the app there as a test case.

src/components/Location.react.js
```diff
@@ -106,7 +116,7 @@ export default class Location extends Component {
         this.updateLocation(this.props);
     }
 
-    UNSAFE_componentWillReceiveProps(nextProps) {
+    componentWillReceiveProps(nextProps) {
```
alexcjohnson: We want to keep the `UNSAFE_` prefix - unless you can figure out how to use a different (safe) method.

HammadTheOne: Hmm, that's strange, I don't even remember removing that. Might be a side-effect of formatting? 

In any case, re-added in c0e72b7. Thanks for pointing it out!

alexcjohnson: I'd guess a mis-merge... or perhaps you had the file open already when you pulled the latest `dev`. Anyway it's fixed now, thanks!

AHMatsWillemsen:
Hi, thanks for this! We found this bug while developing a multi-page app with one part listening for the pathname changes, and one part for the search-changes, resulting in double callbacks for the search callback. In what version do you expect this fix to be included? We can then remove a custom component.

alexcjohnson:
@AHMatsWillemsen this fix is expected to be included in the next Dash version (1.10)

PR #776, opened 2020-03-13
772 set href as a required prop and as the default value for children

### LINKED ISSUE #772: dcc.Link: Use `href` as `children` if no children are defined
With HTML anchors, users are forced to provide a displayed content for the anchor, which can be as simple as the href itself..

```
<a href='myurl'>myurl</a>
```

1:1 correspondance in dcc.Link forces users to do the same thing:
```
dcc.Link(
  href='myurl',
  children=['myurl']
)
```

Defaulting to href for inner content if no children is defined provides a default behavior that will be acceptable in certain situations and will simplify usage.

### LINKED ISSUE #773: Mark dcc.Link `href` prop as required
`href` is not currently required.

A `dcc.Link` without `href` makes no sense, I propose to mark it as `isRequired` / the old behavior being considered a bug, this is not a breaking change.

HammadTheOne:
This PR addresses #772 and #773. The `href` prop is marked as required, and if no `children` prop is defined, the `href` is set as the `children` by default. As @Marc-Andre-Rivet  pointed out, this will simplify usage, and makes more sense for the `dcc.Link` component as a whole. 

- [x] Updated changelog
- [x] Added a basic integration test

Any suggestions on how to better structure the conditional are welcome, would it perhaps be better to set the children directly in the return function (maybe as a ternary operator) instead of as a separate named function?

Closes #772 
Closes #773


src/components/Link.react.js
```diff
@@ -55,6 +55,16 @@ export default class Link extends Component {
 
     render() {
         const {className, style, id, href, loading_state} = this.props;
+
+        const setChildren = () => {
+            if (
+                this.props.children === null ||
+                typeof this.props.children === 'undefined'
+            ) {
```
alexcjohnson: There's a Ramda helper [`isNil`](https://ramdajs.com/docs/#isNil) that does exactly that - you can probably inline this whole thing below using it:
`{isNil(children) ? href : children}`

tests/integration/link/test_link_children.py
```diff
@@ -0,0 +1,30 @@
...
+    def display_children(children):
+        return children
+
+    dash_dcc.start_server(app)
+
+    href_as_children = dash_dcc.driver.execute_script(
+        '''
+        return document.getElementById("link1").text;
+        '''
+    )
```
alexcjohnson: `dash_dcc.wait_for_text_to_equal("#link1", "/page-1")`

tests/integration/link/test_link_children.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+import dash
+from dash.dependencies import Input, Output
+import dash_core_components as dcc
+import dash_html_components as html
+
+
+@pytest.mark.DCC776
+def test_children001_default(dash_dcc):
```
alexcjohnson: the convention we've been using elsewhere (though I see it's only used sporadically in this repo, but let's try to move in this direction) for the test id is `ccffnnn`
- `cc` is 2 chars for the component (`li`?)
- `ff` is 2 chars for the file (`ch`?)
- `nnn` is a serial number within the file

```suggestion
def test_lich001_default(dash_dcc):
```
That way you can run all tests in this file with `pytest -k lich` or just this single test `pytest -k lich001` etc and you can be pretty sure these target just the tests you want.

src/components/Link.react.js
HammadTheOne: Perfect, exactly what I was looking for, I think I'm going to look at using some more Ramda functions for more efficient code. Added in 263984a.

tests/integration/link/test_link_children.py
HammadTheOne: Added in d25a0e4.

alexcjohnson: That statement replaces the whole `href_as_children = ...` and `assert href_as_children == ...`

src/components/Link.react.js
```diff
@@ -71,7 +73,9 @@ export default class Link extends Component {
                 href={href}
                 onClick={e => this.updateLocation(e)}
             >
-                {this.props.children}
+                {isNil(this.props.children)
+                    ? this.props.href
+                    : this.props.children}
```
alexcjohnson: While we're golfing - add `children` to the destructuring of `this.props` above (`href` is already there) and this collapses to just `{isNil(children) ? href : children}`

That's a nicer pattern anyway, as it shows up front what props get used during the render.

HammadTheOne: Makes sense, destructuring does look cleaner. Updated in 065c462.

tests/integration/link/test_link_children.py
HammadTheOne: I'm guessing that method acts as a wrapper of sorts to check the text? Interesting. Updated in da999d5.

tests/integration/link/test_link_children.py
```diff
@@ -0,0 +1,24 @@
...
+def test_lich001_default(dash_dcc):
+    app = dash.Dash(__name__)
+    app.layout = html.Div(
+        [
+            dcc.Link(id="link1", href="/page-1"),
+            dcc.Location(id="url", refresh=False),
+            html.Div(id="content")
+        ]
+    )
+    @app.callback(Output("content", "children"), [Input("link1", "children")])
```
Marc-Andre-Rivet: Does this callback and `content` Div do anything in this test?

Marc-Andre-Rivet: Can you add another simple test checking that children are indeed children when defined

HammadTheOne: Added in 5eac66e.

PR #777, opened 2020-03-13
Git hooks husky

Marc-Andre-Rivet:
Companion to https://github.com/plotly/dash/pull/1151

Husky
`Husky` is a JS package that automagically runs commands on git hooks. For this repo it will run the `lint` script and expect no errors before allowing a commit to go through.

For the hooks to be present, the developer will need to run `npm ci` or `npm install` locally when cloning the repo.

Also adds `black`.


tests/test_integration_2.py
```diff
@@ -158,7 +152,7 @@ def display_data(n_intervals):
 
         self.startServer(app=app)
 
-        start_button = self.wait_for_element_by_css_selector("#start")
+        self.wait_for_element_by_css_selector("#start")
```
alexcjohnson: non-blocking, but we can just remove this line.

Marc-Andre-Rivet: https://github.com/plotly/dash-core-components/pull/777/commits/2ce39da054dbfd68bc5b24f3501a51f2f51d2d9e

Issue #771, opened 2020-03-11
dcc.Link should work with absolute paths


Marc-Andre-Rivet:
In order for dcc.Link click events to behave like HTML anchors,

https://github.com/plotly/dash-core-components/blob/dev/src/components/Link.react.js#L47

```
window.location.pathname = href;
```

needs to be modifed to 
```
window.location= href;
```

For example

```
dcc.Link(
  href='https://www.google.com',
  refresh=True,
  children=['www.google.com']
)
```
will change the path to `http://localhost/https://www.google.com` instead fo opening `https://www.google.com`

wbrgss:
Related to #750 — I wonder if this proposed change would allow for "hotlinking" to anchor links with a URL, via an `id`, instead of just clicking e.g. 

https://en.wikipedia.org/wiki/Python_(programming_language)#Expressions

PR #782, opened 2020-03-20
Change dcc.Link for absolute paths

HammadTheOne:
This PR addresses #771 to open absolute paths directly with `dcc.Link` in the browser page with a refresh rather than appending it to the end of the pathname. 

Rather than `http://localhost/https://www.google.com`, it will now open `https://www.google.com` in the page.


Marc-Andre-Rivet:
@HammadTheOne Can you add a simple test that locks-in the behavior. eg. relative path and absolute path links update the current page as expected.

Issue #685, opened 2019-10-17
"next" button on tabs?


chriddyp:
this would enable tabs to be used as a pagination-like component for story telling.

![next-button-tabs](https://user-images.githubusercontent.com/1280389/67055458-3575d500-f116-11e9-8b74-6c1ce91e50e5.gif)

chriddyp:
![next-button-tabs-2](https://user-images.githubusercontent.com/1280389/67055612-d2387280-f116-11e9-8e75-88a445596311.gif)
https://archive.nytimes.com/www.nytimes.com/interactive/2013/02/04/science/girls-lead-in-science-exam-but-not-in-the-united-states.html

PR #786, opened 2020-04-06
Next Tab Button for dcc.Tabs

HammadTheOne:
This PR addresses #685 and adds a next tab button to sequentially scroll through tabs within `dcc.Tabs`. In it's current implementation, the `next_tab` prop of `dcc.Tabs` is a bool which determines whether or not it is added to the component.

To Do: 

- [ ] Basic Integration Test
- [ ] Support for next tab specific styling.

### DIFF
--- a/src/components/Tabs.react.js
+++ b/src/components/Tabs.react.js
@@ -121,6 +121,95 @@ EnhancedTab.defaultProps = {
     },
 };
 
+const NextTab = ({
+    className,
+    style,
+    selectHandler,
+    value,
+    disabled,
+    disabled_style,
+    disabled_className,
+    mobile_breakpoint,
+    amountOfTabs,
+    colors,
+    vertical,
+    loading_state,
+}) => {
+    let tabStyle = style;
+    if (disabled) {
+        tabStyle = {tabStyle, ...disabled_style};
+    }
+    let tabClassName = `tab ${className || ''}`;
+    if (disabled) {
+        tabClassName += ` tab--disabled ${disabled_className || ''}`;
+    }
+    return (
+        <div
+            data-dash-is-loading={
+                (loading_state && loading_state.is_loading) || undefined
+            }
+            className={tabClassName}
+            id="next-tab"
+            style={tabStyle}
+            onClick={() => {
+                if (!disabled) {
+                    selectHandler(value);
+                }
+            }}
+        >
+            <span>Next Tab</span>
+            <style jsx>{`
+                .tab {
+                    display: inline-block;
+                    background-color: ${colors.background};
+                    border: 1px solid ${colors.border};
+                    border-bottom: none;
+                    padding: 20px 25px;
+                    transition: background-color, color 200ms;
+                    width: 100%;
+                    text-align: center;
+                    box-sizing: border-box;
+                }
+                .tab:last-of-type {
+                    border-right: 1px solid ${colors.border};
+                    border-bottom: 1px solid ${colors.border};
+                }
+                .tab:hover {
+                    cursor: pointer;
+                    background-color: #f0f0f0;
+                }
+                .tab--selected {
+                    border-top: 2px solid ${colors.primary};
+                    color: black;
+                    background-color: white;
+                }
+                .tab--selected:hover {
+                    background-color: white;
+                }
+                .tab--disabled {
+                    color: #d6d6d6;
+                }
+
+                @media screen and (min-width: ${mobile_breakpoint}px) {
+                    .tab {
+                        border: 1px solid ${colors.border};
+                        border-right: none;
+                        ${vertical
+                            ? ''
+                            : `width: calc(100% / ${amountOfTabs});`};
+                    }
+                    .tab--selected,
+                    .tab:last-of-type.tab--selected {
+                        border-bottom: none;
+                        ${vertical
+                            ? `border-left: 2px solid ${colors.primary};`
+                            : `border-top: 2px solid ${colors.primary};`};
+                    }
+                }
+            `}</style>
+        </div>
+    );
+};
 /**
  * A Dash component that lets you render pages with tabs - the Tabs component's children
  * can be dcc.Tab components, which can hold a label that will be displayed as a tab, and can in turn hold
@@ -168,6 +257,7 @@ export default class Tabs extends Component {
     render() {
         let EnhancedTabs;
         let selectedTab;
+        let nextTab;
 
         if (this.props.children) {
             const children = this.parseChildrenToArray();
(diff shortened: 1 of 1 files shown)


src/components/Tabs.react.js
```diff
@@ -121,6 +121,96 @@ EnhancedTab.defaultProps = {
...
+    }
+    return (
+        <div
+            data-dash-is-loading={
+                (loading_state && loading_state.is_loading) || undefined
+            }
+            className={tabClassName}
+            id="next-tab"
+            style={tabStyle}
+            onClick={() => {
```
Marc-Andre-Rivet, line 155: Use the `useCallback` hook instead so as to not create a new function for this callback each time the component is rendered.

src/components/Tabs.react.js
```diff
@@ -121,6 +121,96 @@ EnhancedTab.defaultProps = {
     },
 };
 
+const NextTab = ({
```
Marc-Andre-Rivet, line 124: Consider wrapping in `React.memo` to make this component Pure.
https://reactjs.org/docs/react-api.html#reactmemo

Marc-Andre-Rivet, line 124: This file is becoming very big. Consider breaking up the `Tabs` component file into multiple files. One way to do this would be to rename `src/components/Tabs.react.js` to `src/components/Tabs.react/index.js` (existing imports will recognize `<folder>/index.js` as the same as `<folder>.js` and won't require changes) and moving `EnhancedTabs` and `NextTab` into files of their own, referenced here.

src/components/Tabs.react.js
```diff
@@ -229,6 +320,69 @@ export default class Tabs extends Component {
             });
         }
 
+        if (this.props.children) {
```
Marc-Andre-Rivet, line 323: Could combine both `if` https://github.com/plotly/dash-core-components/pull/786/files#diff-87a94af645034b5ac219cbbfb96a6295R263

src/components/Tabs.react.js
```diff
@@ -168,6 +258,7 @@ export default class Tabs extends Component {
     render() {
         let EnhancedTabs;
```
Marc-Andre-Rivet, line 259: Not a change from this PR, but consider renaming to `enhancedTabs` - Pascal-Casing for variables deviates from the norm.

src/components/Tabs.react.js
```diff
@@ -229,6 +320,69 @@ export default class Tabs extends Component {
...
+                const length = this.props.children.length;
+                let currentIndex = 0;
+
+                const nextTabHandler = () => {
+                    currentIndex = this.props.value.slice(-1) - 1;
+                    currentIndex = ((currentIndex + 1) % length) + 1;
+                    this.props.setProps({value: `tab-${currentIndex}`});
+                };
+
+                if (
```
Marc-Andre-Rivet, line 338: 🌴🐫Same as https://github.com/plotly/dash-core-components/pull/786/files#diff-87a94af645034b5ac219cbbfb96a6295R273 - share common logic.

src/components/Tabs.react.js
```diff
@@ -229,6 +320,69 @@ export default class Tabs extends Component {
...
+                            colors={this.props.colors}
+                            loading_state={childProps.loading_state}
+                            title={this.props.next_tab_layout.title}
+                        />
+                    </div>
+                );
+            };
+        }
+
+        if (this.props.next_tab) {
```
Marc-Andre-Rivet, line 382: Could also be combined in https://github.com/plotly/dash-core-components/pull/786/files#diff-87a94af645034b5ac219cbbfb96a6295R263

src/components/Tabs.react.js
```diff
@@ -326,8 +480,10 @@ Tabs.defaultProps = {
         background: '#f9f9f9',
     },
     vertical: false,
+    next_tab: false,
     persisted_props: ['value'],
     persistence_type: 'local',
+    next_tab_layout: {title: 'Next', style: null},
```
Marc-Andre-Rivet, line 486: `next_tab` and `next_tab_layout` could be combined into a single `navigation` prop that would be extensible in the future. For example:

```
navigation: PropTypes.object({
  next: PropTypes.oneOf([
    true,
    PropTypes.object({
      title: PropTypes.string,
      style: PropTypes.object
    })
  ])
})
```

Using `true` would default to
```
{ title: 'Next', style: undefined }
```

Defining `title`, `style` would combine/override the defaults above.

Marc-Andre-Rivet: Later we can add `Previous`, `Last`, `First`, `Hide Tabs`, etc. to suit our needs.

src/components/Tabs.react.js
```diff
@@ -229,6 +320,69 @@ export default class Tabs extends Component {
             });
         }
 
+        if (this.props.children) {
+            nextTab = () => {
```
Marc-Andre-Rivet, line 324: This doesn't need to be a function, just create the instance of `NextTab` here directly

tests/integration/tab/test_tabs_next_button.py
```diff
@@ -0,0 +1,40 @@
...
+    def display_value(value):
+        return value
+
+    dash_dcc.start_server(app)
+
+    dash_dcc.wait_for_element('#tab-1').click()
+
+    dash_dcc.wait_for_element('#next-tab').click()
+
+    dash_dcc.wait_for_text_to_equal("#tabs-output", "tab-2")
```
Marc-Andre-Rivet, line 40: Would like to see another test case without child tabs that makes sure `Next` is not present.

src/components/Tabs.react.js
```diff
@@ -121,6 +121,96 @@ EnhancedTab.defaultProps = {
...
+    if (disabled) {
+        tabClassName += ` tab--disabled ${disabled_className || ''}`;
+    }
+    return (
+        <div
+            data-dash-is-loading={
+                (loading_state && loading_state.is_loading) || undefined
+            }
+            className={tabClassName}
+            id="next-tab"
```
Marc-Andre-Rivet, line 153: This component fragment can't have a constant `id`. If there are multiple `dcc.Tabs` in the page, they will clash. You could use the component's `id` and append something to it like `${id}-next-tab`.

Marc-Andre-Rivet: For example, `EnhancedTab` uses the child component's `id`, which should not be duplicated.
https://github.com/plotly/dash-core-components/pull/786/files#diff-87a94af645034b5ac219cbbfb96a6295R301

src/components/Tabs.react.js
```diff
@@ -121,6 +121,96 @@ EnhancedTab.defaultProps = {
...
+            style={tabStyle}
+            onClick={() => {
+                if (!disabled) {
+                    selectHandler(value);
+                }
+            }}
+        >
+            <span>{title}</span>
+            <style jsx>{`
```
Marc-Andre-Rivet, line 162: 🌴🐫 Share common styling for `EnhancedTab` / `NextTab`
https://github.com/plotly/dash-core-components/pull/786/files#diff-87a94af645034b5ac219cbbfb96a6295R64

src/components/Tabs.react.js
```diff
@@ -229,6 +320,69 @@ export default class Tabs extends Component {
...
+                            loading_state={childProps.loading_state}
+                            title={this.props.next_tab_layout.title}
+                        />
+                    </div>
+                );
+            };
+        }
+
+        if (this.props.next_tab) {
+            EnhancedTabs = EnhancedTabs.concat(nextTab());
```
Marc-Andre-Rivet, line 383: Since it's the exact opposite, I think we should consider adding `Previous` as part of this PR.

src/components/Tabs.react.js
chriddyp: Instead of `title`, let's do `label` so that it matches `dcc.Tab` (https://github.com/plotly/dash-core-components/blob/3bcc2384576bfc7e9ae12a166f1e38615d9a2c8d/src/components/Tab.react.js#L18-L21)

PR #790, opened 2020-04-09
Issue 788 - dcc.Dropdown dropdown should have priority over table fixed rows/columns

### LINKED ISSUE #788: Header z-index is larger than dcc.Dropdown z-index
From https://community.plot.ly/t/datatable-header-overlaps-on-dropdown-list/16542/6

image

Marc-Andre-Rivet:
Closes #788 

Default z-index for `.Select-menu-outer` is `1`. Better to increase it than to use `0` and negative z-index throughout the table.


tests/integration/dropdown/test_visibility.py
```diff
@@ -0,0 +1,46 @@
+from dash import Dash
+from dash_core_components import Dropdown
+from dash_html_components import Div
+from dash_table import DataTable
+
+import pytest
+
+
+@pytest.mark.DCC788
+def test_visi001_fixed_table(dash_duo):
```
alexcjohnson, line 10:
```suggestion
def test_ddvi001_fixed_table(dash_duo):
```
so this test is still scoped under dropdowns.

alexcjohnson:
Agreed that this is the best approach. Do we have an upper bound for z-index based on modals perhaps, either here or ddk? Is 1000 OK for these?

Marc-Andre-Rivet:
@alexcjohnson ddk modals have `z-index:1011` and the backdrop has `z-index: 1010`

Apart from modals very few things should take priority over an open menu (eg. maybe a tooltip on said menu?), 1000 sounds appropriate in this case.

alexcjohnson:
OK good. `dbc.Modal` uses [1050 by default](https://github.com/facultyai/dash-bootstrap-components/blob/master/src/components/modal/Modal.js#L144) which is also compatible

PR #792, opened 2020-04-13
Make Store component more robust

### LINKED ISSUE #456: issue with `dcc.Store('local')` ?
I haven't investigated this, but the community report looks valid: https://community.plot.ly/t/dcc-store-local-cause-error-loading-dependencies-but-not-dcc-store-memory/19056

### LINKED ISSUE #778: Undesired behaviour (interaction?) with two `dcc.Store`
The issue originates from https://community.plot.ly/t/components-triggered-by-table-not-updating/36288/4
With 
```
import dash
import dash_table
import pandas as pd

import plotly.graph_objs as go

app = dash.Dash(__name__)
server = app.server

	dash_table.DataTable(
	id='table-data',
	data=[{'x':'test', 'graph': 0}],
	columns=[
		{'id': 'x', 'name': 'x', 'editable':False},
		{'id': 'graph', 'name': 'graph', 'presentation': 'dropdown', 'editable':True}],
	dropdown={
		'graph': {
			'options': [
				{'label': 'Do not show', 'value': 0x0},
				{'label': 'Plot 1', 'value': 1},
				{'label': 'Plot 2', 'value': 2}],
			},
		},
	row_deletable=False,
	editable=True,
	),
	dcc.Store(id='g1buffer', storage_type='memory'),
	dcc.Store(id='g2buffer', storage_type='memory'),
	dcc.Graph(id='plot-graph1'),
	dcc.Graph(id='plot-graph2'),
])


@app.callback(
	Output('plot-graph1', 'figure'),
	[Input('g1buffer', 'data')],
)
def update_graph1(data):
	if data is None:
		raise PreventUpdate
	return data

@app.callback(
	Output('plot-graph2', 'figure'),
	[Input('g2buffer', 'data')],
)
def update_graph2(data):
	if data is None:
		raise PreventUpdate
	return data

@app.callback(
	[
		Output('g1buffer', 'data'),
		Output('g2buffer', 'data'),
	],
	[Input('table-data', 'data')],
)
def update_on_table(table_data):
	data = go.Scatter(
				x=[1,2,3,4],
				y=[2,5,1,3],
				)
	g1 = {}
	g2 = {}
	if table_data[0]['graph'] == 1:
		g1 = {'data': [data]}
	if table_data[0]['graph'] == 2:
		g2 = {'data': [data]}
	return g1, g2

if __name__ == '__main__':
    app.run_server(debug=True)
```
the dropdown does not update the figures as expected (sometimes plot 1 does not show up as it should). Modifying the layout to have a single `dcc.Store` seems to solve the problem (see below). Could there be undesired interactions between the two `dcc.Store`?
```
import dash
import dash_table
import pandas as pd

import plotly.graph_objs as go

app = dash.Dash(__name__)
server = app.server



	dash_table.DataTable(
	id='table-data',
	data=[{'x':'test', 'graph': 0}],
	columns=[
		{'id': 'x', 'name': 'x', 'editable':False},
		{'id': 'graph', 'name': 'graph', 'presentation': 'dropdown', 'editable':True}],
	dropdown={
		'graph': {
			'options': [
				{'label': 'Do not show', 'value': 0x0},
				{'label': 'Plot 1', 'value': 1},
				{'label': 'Plot 2', 'value': 2}],
			},
		},
	editable=True,
	),
	dcc.Store(id='gbuffer'),
	dcc.Graph(id='plot-graph1'),
	dcc.Graph(id='plot-graph2'),
])


@app.callback(
	Output('plot-graph1', 'figure'),
	[Input('gbuffer', 'data')],
)
def update_graph1(data):
    print('update_graph1', data)
    if data is None:
        raise PreventUpdate
    return data[0]

@app.callback(
	Output('plot-graph2', 'figure'),
	[Input('gbuffer', 'data')],
)
def update_graph2(data):
    print('update_graph2', data)
    if data is None:
        raise PreventUpdate
    return data[1]

@app.callback(
	 Output('gbuffer', 'data'),
	[Input('table-data', 'data')],
)
def update_on_table(table_data):
    data = go.Scatter(
                            x=[1,2,3,4],
                            y=[2,5,1,3],
                            )
    g1 = {}
    g2 = {}
    if table_data[0]['graph'] == 1:
        g1 = {'data': [go.Scatter(x=[1, 2], y=[1, 2])]}
    if table_data[0]['graph'] == 2:
        g2 = {'data': [go.Scatter(x=[1, 3], y=[2, 3])]}
    return [g1, g2]

if __name__ == '__main__':
    app.run_server(debug=True)
```

alexcjohnson:
Fixes #456
Fixes #778 (unless something else already fixed that)

In particular these fixes support unmounting & remounting Store components, and various data edge cases

This should allow us to remove the confusing special handling of Store in the docs.

### DIFF
--- a/src/components/Store.react.js
+++ b/src/components/Store.react.js
@@ -159,7 +159,7 @@ export default class Store extends React.Component {
         // Note: this still allows you to set data to null
         if (!equals(data, old)) {
             if (data === undefined) {
-                setProps({data: old})
+                setProps({data: old});
             } else {
                 this._backstore.setItem(id, data);
                 setProps({
--- a/src/components/Store.react.js
+++ b/src/components/Store.react.js
@@ -1,45 +1,6 @@
-import {any, includes, isNil, type} from 'ramda';
+import {equals, isNil} from 'ramda';
 import React from 'react';
 import PropTypes from 'prop-types';
-
-/**
- * Deep equality check to know if the data has changed.
- *
- * @param {*} newData - New data to compare
- * @param {*} oldData - The old data to compare
- * @returns {boolean} The data has changed.
- */
-function dataChanged(newData, oldData) {
-    const oldNull = isNil(oldData);
-    const newNull = isNil(newData);
-    if (oldNull || newNull) {
-        return newData !== oldData;
-    }
-    const newType = type(newData);
-    if (newType !== type(oldData)) {
-        return true;
-    }
-    if (newType === 'Array') {
-        if (newData.length !== oldData.length) {
-            return true;
-        }
-        for (let i = 0; i < newData.length; i++) {
-            if (dataChanged(newData[i], oldData[i])) {
-                return true;
-            }
-        }
-    } else if (includes(newType, ['String', 'Number', 'Boolean'])) {
-        return oldData !== newData;
-    } else if (newType === 'Object') {
-        const oldEntries = Object.entries(oldData);
-        const newEntries = Object.entries(newData);
-        if (oldEntries.length !== newEntries.length) {
-            return true;
-        }
-        return any(([k, v]) => dataChanged(v, oldData[k]))(newEntries);
-    }
-    return false;
-}
 
 /**
  * Abstraction for the memory storage_type to work the same way as local/session
@@ -88,7 +49,13 @@ class WebStore {
     }
 
     getItem(key) {
-        return JSON.parse(this._storage.getItem(key));
+        try {
+            return JSON.parse(this._storage.getItem(key));
+        } catch (e) {
+            // in case we somehow got a non-JSON value in storage,
+            // just ignore it.
+            return null;
+        }
     }
 
     setItem(key, value) {
@@ -150,7 +117,7 @@ export default class Store extends React.Component {
         }
 
         const old = this._backstore.getItem(id);
-        if (isNil(old) && data) {
+        if (isNil(old) && !isNil(data)) {
             // Initial data mount
             this._backstore.setItem(id, data);
             setProps({
@@ -159,7 +126,7 @@ export default class Store extends React.Component {
             return;
         }
 
-        if (dataChanged(old, data)) {
+        if (!equals(old, data)) {
             setProps({
                 data: old,
                 modified_timestamp: this._backstore.getModified(id),
@@ -186,11 +153,19 @@ export default class Store extends React.Component {
         }
         const old = this._backstore.getItem(id);
         // Only set the data if it's not the same data.
-        if (dataChanged(data, old)) {
-            this._backstore.setItem(id, data);
-            setProps({
-                modified_timestamp: this._backstore.getModified(id),
-            });
+        // If the new data is undefined, we got here by overwriting the entire
+        // component with a new copy that has no `data` specified - so pull back
+        // out the old value.
+        // Note: this still allows you to set data to null
+        if (!equals(data, old)) {
+            if (data === undefined) {
+                setProps({data: old})
+            } else {
+                this._backstore.setItem(id, data);
+                setProps({
+                    modified_timestamp: this._backstore.getModified(id),
+                });
+            }
         }
     }
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,7 @@ jobs:
             # build main dash
             git clone --depth 1 https://github.com/plotly/dash.git dash-main
             cd dash-main && pip install -e .[dev] --progress-bar off && python setup.py sdist && mv dist/* ../packages/
-            cd dash-renderer && npm run build
+            cd dash-renderer && npm ci && npm run build
             python setup.py sdist && mv dist/* ../../packages/ && cd ../..
             # build dcc
             npm ci && npm run build && python setup.py sdist && mv dist/* ./packages
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - [#793](https://github.com/plotly/dash-core-components/pull/793) Added title key (i.e. HTML `title` attribute) to option dicts in `dcc.Dropdown` `options[]` list property.
 ### Fixed
+- [#792](https://github.com/plotly/dash-core-components/pull/792) Improved the robustness of `dcc.Store` components, fixing [#456](https://github.com/plotly/dash-core-components/issues/456) whereby persistent storage could become corrupted, and fixing lifecycle issues that prevented adding `Store` components to the page after initial loading.
 - [#790](https://github.com/plotly/dash-core-components/pull/790) Fixed bug where the dcc.Dropdown dropdown was hidden by the dash_table.DataTable fixed rows and columns.
 
 ## [1.9.1] - 2020-04-10
--- a/tests/integration/store/test_component_props.py
+++ b/tests/integration/store/test_component_props.py
@@ -70,3 +70,123 @@ def test_stcp002_modified_ts(store_app, dash_dcc):
     assert ts == approx(
         output_data.get("ts"), abs=40
     ), "the modified_timestamp should be updated right after the click action"
+
+
+def test_stcp003_initial_falsy(dash_dcc):
+    app = dash.Dash(__name__)
+    app.layout = html.Div([html.Div([
+        storage_type,
+        dcc.Store(storage_type=storage_type, id="zero-" + storage_type, data=0),
+        dcc.Store(storage_type=storage_type, id="false-" + storage_type, data=False),
+        dcc.Store(storage_type=storage_type, id="null-" + storage_type, data=None),
+        dcc.Store(storage_type=storage_type, id="empty-" + storage_type, data=""),
+    ]) for storage_type in ("memory", "local", "session")], id="content")
+
+    dash_dcc.start_server(app)
+    dash_dcc.wait_for_text_to_equal("#content", "memory\nlocal\nsession")
+
+    for storage_type in ("local", "session"):
+        getter = getattr(dash_dcc, "get_{}_storage".format(storage_type))
+        assert getter("zero-" + storage_type) == 0, storage_type
+        assert getter("false-" + storage_type) is False, storage_type
+        assert getter("null-" + storage_type) is None, storage_type
+        assert getter("empty-" + storage_type) == "", storage_type
+
+    assert not dash_dcc.get_logs()
+
+
+def test_stcp004_remount_store_component(dash_dcc):
+    app = dash.Dash(__name__, suppress_callback_exceptions=True)
+
+    content = html.Div(
+        [
+            dcc.Store(id="memory", storage_type="memory"),
+            dcc.Store(id="local", storage_type="local"),
+            dcc.Store(id="session", storage_type="session"),
+            html.Button("click me", id="btn"),
+            html.Button("clear data", id="clear-btn"),
+            html.Div(id="output"),
+        ]
+    )
+
+    app.layout = html.Div([html.Button("start", id="start"), html.Div(id="content")])
+
+    @app.callback(Output("content", "children"), [Input("start", "n_clicks")])
+    def start(n):
+        return content if n else "init"
+
+    @app.callback(
+        Output("output", "children"),
+        [
+            Input("memory", "modified_timestamp"),
+            Input("local", "modified_timestamp"),
+            Input("session", "modified_timestamp")
+        ],
+        [State("memory", "data"), State("local", "data"), State("session", "data")],
+    )
+    def write_memory(tsm, tsl, tss, datam, datal, datas):
+        return json.dumps([datam, datal, datas])
+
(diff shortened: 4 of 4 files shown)


src/components/Store.react.js
```diff
@@ -1,46 +1,7 @@
...
 import PropTypes from 'prop-types';
 
-/**
- * Deep equality check to know if the data has changed.
- *
- * @param {*} newData - New data to compare
- * @param {*} oldData - The old data to compare
- * @returns {boolean} The data has changed.
- */
-function dataChanged(newData, oldData) {
```
alexcjohnson, line 12: Any reason to maintain a custom deep equality test here, instead of just `Ramda.equals`? The tests (particularly [`stda001`](https://github.com/plotly/dash-core-components/blob/3bcc2384576bfc7e9ae12a166f1e38615d9a2c8d/tests/integration/store/test_store_data.py#L13) say they're equivalent.

src/components/Store.react.js
```diff
@@ -150,7 +117,7 @@ export default class Store extends React.Component {
         }
 
         const old = this._backstore.getItem(id);
-        if (isNil(old) && data) {
+        if (isNil(old) && !isNil(data)) {
```
alexcjohnson, line 120: [`test_stcp003_initial_falsy`](https://github.com/plotly/dash-core-components/pull/792/files#diff-0736b190bed15e75e2dd5b4f66d5cca9R75)

src/components/Store.react.js
```diff
@@ -186,11 +153,19 @@ export default class Store extends React.Component {
...
-            this._backstore.setItem(id, data);
-            setProps({
-                modified_timestamp: this._backstore.getModified(id),
-            });
+        // If the new data is undefined, we got here by overwriting the entire
+        // component with a new copy that has no `data` specified - so pull back
+        // out the old value.
+        // Note: this still allows you to set data to null
+        if (!equals(data, old)) {
+            if (data === undefined) {
```
alexcjohnson, line 161: I bet this would all be even simpler if we moved to a functional component with `useEffect`, but anyway this is the key point allowing you to re-supply the same component with no `data` and have the old value persist.

If you actually remove it completely and bring it back later, it already worked, via the `UNSAFE_componentWillMount` pathway. This is only if the component is already on the page and you provide it again - ie as part of a larger `children` prop in which some other things may have changed.

src/components/Store.react.js
```diff
@@ -88,7 +49,13 @@ class WebStore {
     }
 
     getItem(key) {
-        return JSON.parse(this._storage.getItem(key));
+        try {
+            return JSON.parse(this._storage.getItem(key));
+        } catch (e) {
+            // in case we somehow got a non-JSON value in storage,
+            // just ignore it.
+            return null;
```
alexcjohnson, line 57: closes #456

src/components/Store.react.js
Marc-Andre-Rivet: 🔪Looks legit.

src/components/Store.react.js
Marc-Andre-Rivet: We could also try and resolve the `undefined` case gracefully and only resort to the `catch` if it's a truly unexpected value.

```
const value = this._storage.getItem(key);
return isNil(value) ? null : JSON.parse(value);
```

alexcjohnson: The problem I was hitting is sometimes the string `’undefined’` was making its way into storage. That’s neither nil nor JSON parsable. Shouldn’t happen anymore, but if an app running an old version of this component corrupts the storage, we have no way to recover. I suppose we could make a narrow exception for just the string I’ve seen but I didn’t see the harm in catching more general corruption.

src/components/Store.react.js
```diff
@@ -186,11 +153,19 @@ export default class Store extends React.Component {
...
-            setProps({
-                modified_timestamp: this._backstore.getModified(id),
-            });
+        // If the new data is undefined, we got here by overwriting the entire
+        // component with a new copy that has no `data` specified - so pull back
+        // out the old value.
+        // Note: this still allows you to set data to null
+        if (!equals(data, old)) {
+            if (data === undefined) {
+                setProps({data: old});
```
Marc-Andre-Rivet, line 162: Only setting `data` here prevents scenarios using `Input=modified_timestamp, State=data --> Output=data` from being triggered. Could we add `modified_timestamp: this._backstore.getModified(id),` here too?

Marc-Andre-Rivet, line 162: Running this against a modified version of https://dash.plotly.com/dash-core-components/store, I expected to be able to rename the `dcc.Store` component in there for ones that were unique to this layout and everything work 👌 

Turns out the callback triggered by https://github.com/plotly/dash-docs/blob/master/dash_docs/chapters/dash_core_components/Store/examples/store_clicks.py#L74 is not triggered on load, resulting in all field having value `0`. On click, they are updated to `previous+1` as expected.

src/components/Store.react.js
Marc-Andre-Rivet: No worries. Just nitpicking. I agree there's no scenario where we dont `try..catch`, might as well keep it as-is.

src/components/Store.react.js
alexcjohnson: > is not triggered on load, resulting in all field having value `0`.

Can you post the exact app you're using? The app seems to behave correctly when I try it.

Marc-Andre-Rivet: Prior to changes,

1. Opening http://127.0.0.1:8060/dash-core-components/store, all values for the stores will be `0`. Click the buttons to update and see that the callbacks work.
2. Refresh the page, local and session storage counters are restore to their value, memory is `0`

Here's the modified code for `store_clicks.py`
- remove `# no-exec` comments
- rename memory/local/session to `new-[memory|session|local]`
- update callbacks accordingly
 
```
import dash

from dash.dependencies import Output, Input, State

# This stylesheet makes the buttons and table pretty.

app.layout = html.Div([
    # The memory store reverts to the default on every page refresh
    dcc.Store(id='new-memory'), # stores are loaded from the beginning
    # The local store will take the initial data
    # only the first time the page is loaded
    # and keep it until it is cleared.
    dcc.Store(id='new-local', storage_type='local'),
    # Same as the local store but will lose the data
    # when the browser/tab closes.
    dcc.Store(id='new-session', storage_type='session'),
    html.Table([
        html.Thead([
            html.Tr(html.Th('Click to store in:', colSpan="3")),
            html.Tr([
                html.Th(html.Button('memory', id='memory-button')),
                html.Th(html.Button('localStorage', id='local-button')),
                html.Th(html.Button('sessionStorage', id='session-button'))
            ]),
            html.Tr([
                html.Th('Memory clicks'),
                html.Th('Local clicks'),
                html.Th('Session clicks')
            ])
        ]),
        html.Tbody([
            html.Tr([
                html.Td(0, id='memory-clicks'),
                html.Td(0, id='local-clicks'),
                html.Td(0, id='session-clicks')
            ])
        ])
    ])
])

# Create two callback for every store.
for store in ('memory', 'local', 'session'):

    # add a click to the appropriate store.
    @app.callback(Output('new-{}'.format(store), 'data'),
                  [Input('{}-button'.format(store), 'n_clicks')],
                  [State('new-{}'.format(store), 'data')])
    def on_click(n_clicks, data):
        if n_clicks is None:
            # prevent the None callbacks is important with the store component.
            # you don't want to update the store for nothing.
            raise PreventUpdate

        # Give a default data dict with 0 clicks if there's no data.
        data = data or {'clicks': 0}

        data['clicks'] = data['clicks'] + 1
        return data

    # output the stored clicks in the table cell.
    @app.callback(Output('{}-clicks'.format(store), 'children'),
                  # Since we use the data prop in an output,
                  # we cannot get the initial data on load with the data prop.
                  # To counter this, you can use the modified_timestamp
                  # as Input and the data as State.
                  # This limitation is due to the initial None callbacks
                  # https://github.com/plotly/dash-renderer/pull/81
                  [Input('new-{}'.format(store), 'modified_timestamp')],
                  [State('new-{}'.format(store), 'data')])
    def on_data(ts, data):
        if ts is None:
            raise PreventUpdate

        data = data or {}

        return data.get('clicks', 0)


if __name__ == '__main__':
    app.run_server(debug=True, port=8077, threaded=True)
```

1. Opening http://127.0.0.1:8060/dash-core-components/store, all values for the stores will be `0`. Click the buttons to update and see that the callbacks work.
2. Refresh the page, all counters are at `0` again
3. Click on the local or session counter button, notice that it will jump from `0` to `previous+1`

Running this against latest of `dash@dev`.

alexcjohnson: got it, thanks! Investigating... Here's a standalone version that shows the bug, just wrapping the layout in a content callback:
```py
import dash

import dash_core_components as dcc  # noxx-exec
from dash.dependencies import Output, Input, State

# This stylesheet makes the buttons and table pretty.

app = dash.Dash(
    __name__,
    external_stylesheets=external_stylesheets,
    suppress_callback_exceptions=True
)

layout = html.Div([
    # The memory store reverts to the default on every page refresh
    dcc.Store(id='memory'),  # noxx-exec - stores are loaded from the beginning
    # The local store will take the initial data
    # only the first time the page is loaded
    # and keep it until it is cleared.
    dcc.Store(id='local', storage_type='local'),  # noxx-exec
    # Same as the local store but will lose the data
    # when the browser/tab closes.
    dcc.Store(id='session', storage_type='session'),  # noxx-exec
    html.Table([
        html.Thead([
            html.Tr(html.Th('Click to store in:', colSpan="3")),
            html.Tr([
                html.Th(html.Button('memory', id='memory-button')),
                html.Th(html.Button('localStorage', id='local-button')),
                html.Th(html.Button('sessionStorage', id='session-button'))
            ]),
            html.Tr([
                html.Th('Memory clicks'),
                html.Th('Local clicks'),
                html.Th('Session clicks')
            ])
        ]),
        html.Tbody([
            html.Tr([
                html.Td(0, id='memory-clicks'),
                html.Td(0, id='local-clicks'),
                html.Td(0, id='session-clicks')
            ])
        ])
    ])
])

app.layout = html.Div(id="content")


@app.callback(Output("content", "children"), [Input("content", "style")])
def content(_):
    return layout


# Create two callback for every store.
for store in ('memory', 'local', 'session'):
    # add a click to the appropriate store.
    @app.callback(Output(store, 'data'),
                  [Input('{}-button'.format(store), 'n_clicks')],
                  [State(store, 'data')])
    def on_click(n_clicks, data):
        if n_clicks is None:
            # prevent the None callbacks is important with the store component.
            # you don't want to update the store for nothing.
            raise PreventUpdate

        # Give a default data dict with 0 clicks if there's no data.
        data = data or {'clicks': 0}

        data['clicks'] = data['clicks'] + 1
        return data

    # output the stored clicks in the table cell.
    @app.callback(Output('{}-clicks'.format(store), 'children'),
                  # Since we use the data prop in an output,
                  # we cannot get the initial data on load with the data prop.
                  # To counter this, you can use the modified_timestamp
                  # as Input and the data as State.
                  # This limitation is due to the initial None callbacks
                  # https://github.com/plotly/dash-renderer/pull/81
                  [Input(store, 'modified_timestamp')],
                  [State(store, 'data')])
    def on_data(ts, data):
        if ts is None:
            raise PreventUpdate

        data = data or {}

        return data.get('clicks', 0)


if __name__ == '__main__':
    app.run_server(debug=True)
```

alexcjohnson: Ha, this is actually yet another renderer bug... PR coming in Dash as soon as I figure out how to write a test for it!

alexcjohnson: Fixed in https://github.com/plotly/dash/pull/1224

alexcjohnson: https://github.com/plotly/dash-docs/pull/871 shows that between this PR and https://github.com/plotly/dash/pull/1224 there are no issues anymore running the store examples unmodified within the docs.

Marc-Andre-Rivet:
@alexcjohnson Took the liberty of updating the PR https://github.com/plotly/dash-core-components/pull/792/commits/d771fc7eb11e5f7a9b214b60f4f260f95de2d659 for side-effect of the dash-renderer build that uncovered the missing `npm ci` when building the renderer from dev in CI.

wbrgss:
> Took the liberty of updating the PR d771fc7 for side-effect of the dash-renderer build that uncovered the missing npm ci when building the renderer from dev in CI.

Nice, thanks @Marc-Andre-Rivet! Depending on how close this is, it might warrant its own PR. DCC (among other libraries) build `dash-renderer` in their CI runs, so `sh: 1: babel: not found` is blocking these runs at the moment. 

E.g. https://circleci.com/gh/plotly/dash-core-components/20126

Marc-Andre-Rivet:
@wbrgss Afaik, this is ready to go. https://github.com/plotly/dash-table/pull/766 fixes it for the table, the html component fix is already in. Since these should all the last items to make it into `Dash v1.12` prior to the fix, it's probably OK to just wait it out. Let me know if you have some additional timing concerns and I'll make a side PR for table/dcc.

PR #793, opened 2020-04-15
Expose Dropdown option 'title' property

wbrgss:
The HTML property `title` is already available upstream in `react-virtualized-select`:

https://github.com/bvaughn/react-virtualized-select/blob/master/source/VirtualizedSelect/VirtualizedSelect.js#L186

So I exposed it in the Dropdown component as a solution was requested by @KPhans:

>  is there a way to have hover descriptions of dropwdown selections?

If the core team is on board with this I'll add an integration test and a docs example.

Demo:

title

### DIFF
--- a/tests/integration/dropdown/test_option_title_prop.py
+++ b/tests/integration/dropdown/test_option_title_prop.py
@@ -6,6 +6,7 @@ import dash_html_components as html
 
 from utils import wait_for
 
+
 @pytest.mark.DCC793
 @pytest.mark.parametrize("multi", [True, False])
 def test_ddot001_option_title(dash_dcc, multi):
--- a/tests/integration/dropdown/test_option_title_prop.py
+++ b/tests/integration/dropdown/test_option_title_prop.py
@@ -1,3 +1,5 @@
+# -*- coding: UTF-8 -*-
+
 import pytest
 import dash
 from dash.dependencies import Input, Output
--- a/tests/integration/dropdown/test_option_title_prop.py
+++ b/tests/integration/dropdown/test_option_title_prop.py
@@ -1,0 +1,54 @@
+import pytest
+import dash
+from dash.dependencies import Input, Output
+import dash_core_components as dcc
+import dash_html_components as html
+
+from utils import wait_for
+
+@pytest.mark.DCC793
+@pytest.mark.parametrize("multi", [True, False])
+def test_ddot001_option_title(dash_dcc, multi):
+    app = dash.Dash(__name__)
+    app.layout = html.Div([
+        dcc.Input(
+            id="dropdown_title_input",
+            type="text",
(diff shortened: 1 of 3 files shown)


chriddyp:
I like it!

wbrgss:
@Marc-Andre-Rivet I would appreciate a review on c38eb00; I'm writing integration tests in another project and a review might help educate me on any bad practices I should correct

I'll leave docs to another issue https://github.com/plotly/dash-docs/issues/855

KPhans:
If this is done by tmrw we can implement this for Loblaws 😊

tests/integration/dropdown/test_option_title_prop.py
```diff
@@ -0,0 +1,57 @@
...
+    dropdown_title_input.send_keys("The Big Apple")
+
+    wait_for(
+        lambda: dropdown_option_element.get_attribute("title")
+        == "The Big Apple",
+        lambda: '`title` is {}, expected {}'.format(
+            dropdown_option_element.get_attribute("title"),
+            "The Big Apple",
+        ),
+    )
```
Marc-Andre-Rivet, line 57: @wbrgss I'm not sure why there's a callback to update the `title` on the 1st option: If all we want to do is make sure it's present, we don't need to. If we want to make sure it's present after an update, we should also test that the `title` attribute is missing or empty prior to the change, to be consistent.

I'm ok with us not testing further (eg. `title` present in the dropdown, not just the selected value) as that would be the same as testing the 3rd party component that we're wrapping.

---

Using `wait_for_element` instead of `find_element_by_css_selector` makes for a slightly more verbose test but also for a slightly more resilient test (eg. timing issues in CI). That said the test app is pretty small and probably will rarely be an issue. Same goes for `wait_for` vs. simply getting the attribute. This is 👌

wbrgss: > If we want to make sure it's present after an update, we should change it more than once and make sure it changes None --> "Big Apple" --> "Knicks Game" --> None for example.

Agreed, that makes sense. I'll update it.

> makes for a slightly more verbose test but also for a slightly more resilient test

Yeah, I've had timing issues here and there with more complicated test apps so I've been using `wait_for` for overkill to be safe.

alexcjohnson:
@wbrgss looks like this one should be easy to wrap up for inclusion in the upcoming release?

tests/integration/dropdown/test_option_title_prop.py
wbrgss: >  If we want to make sure it's present after an update, we should change it more than once and make sure it changes None --> "Big Apple" --> "Knicks Game" --> None for example.

@Marc-Andre-Rivet I've done this in d5bee5d; if that looks right to you I'm going to merge.

Issue #632, opened 2019-05-30
[BUG] data-dash-is-loading attribute not updated on Graph


fl4p:
In `dash-0.41.0` the `data-dash-is-loading` attribute always stays at `true` value on Graph elements.

ethanopp:
Has there been any update on this? I am experiencing the same issue on 1.9.1

ohaibbq:
@fl4p @ethanopp I have this patch if you want to try it out-
https://github.com/plotly/dash-core-components/pull/837

ned2:
It's working for me!

HammadTheOne:
Closed via #837

PR #837, opened 2020-07-24
Update Graph component as `loading_state` changes

ohaibbq:
This fixes an issue where the `dcc.Graph` element would not appropriately update `[data-dash-is-loading]`. 
This was experienced in our production `dash` deployment, and has been reported by other folks in #632 

The attached test case fails consistently in `1.10.1` & passes consistently in this feature branch.
The behavior works as expected using this patch in production application.

Let me know if you have a better approach to the integration test assertions for toggling between loading/not loading. 

Cheers~

### DIFF
--- a/tests/integration/graph/test_graph_basics.py
+++ b/tests/integration/graph/test_graph_basics.py
@@ -1,6 +1,6 @@
 import pytest
 import pandas as pd
-from multiprocessing import Value
+from multiprocessing import Value, Lock
 import numpy as np
 from time import sleep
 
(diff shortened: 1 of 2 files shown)


tests/integration/graph/test_graph_basics.py
```diff
@@ -135,3 +136,34 @@ def update_graph(value):
...
+        html.H1('loading state updates'),
+        dcc.Interval(id='interval'),
+        dcc.Graph(id='my-graph'),
+    ])
+
+    @app.callback(Output('my-graph', 'figure'), [Input('interval', 'n_intervals')])
+    def update_graph(value):
+        values = [random.randint(0, 20), random.randint(0, 20)]
+        ranges = [random.randint(0, 20), random.randint(0, 20)]
+        sleep(0.25)
```
alexcjohnson, line 154: @ohaibbq Thanks for the test! Two comments about it though:
1. I know you aren't depending on the random numbers at all, but it still makes me uneasy to put random numbers in a test. Just use the callback input to make unique values if needed.
2. We should check both the loading state and the finished state, and do so without an explicit `sleep` time - this is both slow and potentially flaky, as timing can be erratic particularly on CI. Instead, use a `multiprocessing.Lock` pattern like we do in test `ldcp001`:

https://github.com/plotly/dash-core-components/blob/b7d1ebee327cf1b1938569a07cb5bf0dae4ecc54/tests/integration/loading/test_loading_component.py#L18-L29

If it's easier, you can test the finished state by using `dash_dcc.wait_for_no_elements(selector)` rather than finding a selector that matches only with `data-dash-is-loading` removed.

tests/integration/graph/test_graph_basics.py
```diff
@@ -135,3 +136,34 @@ def update_graph(value):
...
+                'y': values,
+                'line': {
+                    'shape': 'spline'
+                }
+            }],
+        }
+
+    dash_dcc.start_server(app)
+
+    my_graph = dash_dcc.wait_for_element('#my-graph[data-dash-is-loading="true"]')
```
alexcjohnson, line 167: The one test failure you have is from here:

```
./tests/integration/graph/test_graph_basics.py:167:5: F841 local variable 'my_graph' is assigned to but never used
```

alexcjohnson:
Thanks for the PR @ohaibbq - nice find re: `shouldComponentUpdate` 🏆

Just a couple of comments on the new test and we'll be ready to go 🎉

tests/integration/graph/test_graph_basics.py
ohaibbq: Thanks! Its been a few years since doing selenium testing, I'll update to the Lock pattern and assert that it updates appropriately.

ohaibbq:
🎺 💀  updated to use the `Lock` pattern.

What's the release timing looking like. Any chance we could get a pre-release in the next week or two?

ohaibbq:
🔥

Issue #831, opened 2020-07-08
Graph dimension problem with dcc.Loading — regression (?) in DCC 1.9.1


wbrgss:
Given a certain layout (I've tried to trim down to a minimal example as best I can), `dcc.Loading` causes its `dcc.Graph` children to not obey their set heights in CSS. 

In the below examples, I expect the Graph to be constrained within the purple boundary.

With the code below, with Dash 1.11.0 and Dash Core Components 1.9.1:

![image](https://user-images.githubusercontent.com/9068290/86860128-4692b780-c092-11ea-9fa7-685d7f871a48.png)

With Dash 1.11.0 and Dash Core Components 1.9.0 (or simply Dash <= 1.10.0):

![image](https://user-images.githubusercontent.com/9068290/86860208-6b872a80-c092-11ea-976a-8b844d3ca576.png)

Unfortunately, there was one rule I had to put in an `/assets/style.css`, I left a comment in the app.

**app.py**

```py
import dash
import dash_html_components as html
import dash_core_components as dcc

app = dash.Dash(__name__)

app.layout = html.Div(
    children=[
        # this outer div is necessary to reproduce
        html.Div(
            dcc.Loading(
                # This must load CSS from an external file
                # in /assets via the className
                # style={"height": "100%"} won't have the same effect
                className="loading-container",
                children=dcc.Graph(
                    style={"height": "100%"},
                    figure={
                        'data': [{
                            'x': [1, 2, 3, 4],
                            'y': [4, 1, 6, 9],
                            'line': {'shape': 'spline'}
                        }]
                    }
                ),
            ),
        )
    ],
    style={
        # these rules are necessary to reproduce
        "display": "flex",
        "maxHeight": "30vh",
        # the below two rules are for visual demo purposes
        "padding": "15px",
        "background": "purple"
    },
)

if __name__ == "__main__":
    app.run_server(debug=True)
```

**style.css**
```css
.loading-container {
    height: 100%;
}
```

wbrgss:
My guess is that this is coming from https://github.com/plotly/dash-core-components/pull/740. If you replace the `dcc.Loading` with an `html.Div` in the example above, the problem goes away. I welcome @alexcjohnson to take a look at this please, as you recently got your hands dirty with graph dimensions, and it looks like you were the main reviewer on/a contributor to #740 :slightly_smiling_face:

wbrgss:
Some more details after some investigation:

>  ```py
>    # This must load CSS from an external file
[...]

PR #840, opened 2020-08-06
Allow parent_className and parent_style attributes on dcc.Loading components

wbrgss:
Fixes https://github.com/plotly/dash-core-components/issues/831

I plan to make a test for this in the vein of the test app in https://github.com/plotly/dash-core-components/issues/831#issue-652786009

### DIFF
--- a/src/components/Loading.react.js
+++ b/src/components/Loading.react.js
@@ -123,7 +123,7 @@ Loading.propTypes = {
     debug: PropTypes.bool,
 
     /**
-     * CSS class for the spinner root DOM node
+     * Additional CSS class for the spinner root DOM node
      */
     className: PropTypes.string,
 
--- a/src/components/Loading.react.js
+++ b/src/components/Loading.react.js
@@ -45,6 +45,7 @@ export default class Loading extends Component {
             color,
             className,
             style,
+            parent_className,
             parent_style,
             fullscreen,
             debug,
@@ -56,10 +57,11 @@ export default class Loading extends Component {
 
         return (
             <div
-                className={isLoading ? null : className}
-                style={isLoading
-                    ? mergeRight(hiddenContainer, parent_style)
-                    : parent_style
+                className={parent_className}
+                style={
+                    isLoading
+                        ? mergeRight(hiddenContainer, parent_style)
+                        : parent_style
                 }
             >
                 {this.props.children}
@@ -121,11 +123,14 @@ Loading.propTypes = {
     debug: PropTypes.bool,
 
     /**
-     * Additional CSS styling, applied to one of:
-     *   - the spinner root DOM node, if the component's children are currently loading
-     *   - the parent div of the children, once they are loaded
+     * CSS class for the spinner root DOM node
      */
     className: PropTypes.string,
+
+    /**
+     *  Additional CSS class for the outermost dcc.Loading parent div DOM node
+     */
+    parent_className: PropTypes.string,
 
     /**
      * Additional CSS styling for the spinner root DOM node
--- a/src/components/Loading.react.js
+++ b/src/components/Loading.react.js
@@ -5,6 +5,7 @@ import DefaultSpinner from '../fragments/Loading/spinners/DefaultSpinner.jsx';
 import CubeSpinner from '../fragments/Loading/spinners/CubeSpinner.jsx';
 import CircleSpinner from '../fragments/Loading/spinners/CircleSpinner.jsx';
 import DotSpinner from '../fragments/Loading/spinners/DotSpinner.jsx';
+import {mergeRight} from 'ramda';
 
 function getSpinner(spinnerType) {
     switch (spinnerType) {
@@ -44,6 +45,7 @@ export default class Loading extends Component {
             color,
             className,
             style,
+            parent_style,
             fullscreen,
             debug,
             type: spinnerType,
@@ -53,7 +55,13 @@ export default class Loading extends Component {
         const Spinner = isLoading && getSpinner(spinnerType);
 
         return (
-            <div style={isLoading ? hiddenContainer : {}}>
+            <div
+                className={isLoading ? null : className}
+                style={mergeRight(
+                    isLoading ? hiddenContainer : {},
+                    parent_style
+                )}
+            >
                 {this.props.children}
                 <div style={isLoading ? coveringSpinner : {}}>
                     {isLoading && (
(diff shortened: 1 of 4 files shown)


src/components/Loading.react.js
```diff
@@ -53,7 +55,13 @@ export default class Loading extends Component {
         const Spinner = isLoading && getSpinner(spinnerType);
 
         return (
-            <div style={isLoading ? hiddenContainer : {}}>
+            <div
+                className={isLoading ? null : className}
```
wbrgss, line 59: In the original behavior, modified in #740 which introduced the regression, the `className` was either applied to 
- the Spinner, if in a loading state
- the `this.props.children` div parent, otherwise (at least, if `this.props.children` wasn't an `Object` or `Function`, for some reason. If it was, no `className` would be applied. See below for these three conditions.)

I'm attempting to mimic this behaviour here. Ideally this would be two different props, in the pattern of `style` and the new (in this PR) `parent_style`, but in order to fix the regression and have current apps working without breaking changes I think we need to replicate the old behaviour here.

This was the previous render return value:

https://github.com/plotly/dash-core-components/blob/aa4fce7b739de4c91fbae9e746c0b06883dd8569/src/components/Loading.react.js#L40-L73

wbrgss:
Some relevant comments from #740 that I wish to address in this PR:

> One issue that stems from this is that the component itself would be moved "down" a level in the DOM, **which might affect the CSS/layout of different apps**.

(https://github.com/plotly/dash-core-components/pull/740#issue-366021779)

> Some tests are currently failing because I didn't put className onto the wrapper div when loading is done, I only put it on the spinner itself, when the spinner is present. Is it useful for this class to be applied to the wrapper? At first glance I'd think this is actually unwanted, because then you can't use this class to target the spinner. 

https://github.com/plotly/dash-core-components/pull/740#issuecomment-611638124

The problem here is that you do need to be able to apply a class to the wrapper in order to avoid the situation described in #831, and e.g. [this community issue](https://community.plotly.com/t/percentage-height-of-dcc-loading-children/42222/2) (and previously, this was the `className` prop of the Loading component). It's maybe not ideal to have this class possibly applied to two different elements, but this was the behavior before, and it might be OK considering it only applies to one element at a time depending on the loading state. cc @alexcjohnson

wbrgss:
> It's maybe not ideal to have this class possibly applied to two different elements, but this was the behavior before, and it might be OK considering it only applies to one element at a time depending on the loading state

An alternative to this would be to keep `className` for the `Spinner`, and add `parent_className` to accompany the `parent_style` in this PR. This wouldn't restore the previous behaviour before the regression, but rather provide a method for users relying on previous behaviour to fix their apps — namely by changing their `dcc.Loading` `className` prop to `parent_className`. The advantage is this would prevent the consistency problem with `className` conditionally assigned to different elements in the hierarchy depending on loading state.

What do you think @alexcjohnson? It's a breaking change, but arguably that ship has sailed already with DCC 1.9.1. If you think `parent_className` is preferable, I'll document `parent_className` in this PR and make sure to notify the users who have reported this issue so they can fix their apps.

alexcjohnson:
Talking with @wbrgss just now, it seems like there was never really a use case for having the same className on both the spinner and the parent. There was a use case for the spinner className (to style the spinner itself) and a separate use case to style the parent (primarily to inherit sizing from its own parent). But if you happened to require both of these use cases at the same time, you'd be out of luck - one of them would break the other behavior.

Based on that, I think we can consider the previous behavior (prior to #740) to be a bug, and we can use @wbrgss's [suggestion](https://github.com/plotly/dash-core-components/pull/840#issuecomment-670598263) of adding `parent_className` as a separate prop as a clean way to support both use cases without them colliding.

src/components/Loading.react.js
```diff
@@ -53,7 +55,13 @@ export default class Loading extends Component {
...
+                style={mergeRight(
+                    isLoading ? hiddenContainer : {},
+                    parent_style
+                )}
```
alexcjohnson, line 63:
```suggestion
                style={isLoading
                    ? mergeRight(hiddenContainer, parent_style)
                    : parent_style
                }
```
Minor, just to avoid having to do a noop merge.

wbrgss:
Another question is where to put `id` — which is in the `PropTypes`, but doesn't (and hasn't ever, AFAIK) appear anywhere in the render function. I think it should be added.

I guess it would be consistent to add it as a prop of the `Spinner` but it still doesn't quite seem right to me, as when the app is first initialized, one could have a callback that uses an `id`, but the Spinner won't necessarily exist in the DOM at that time. But maybe that's OK? We could have `parent_id` — but I don't think that would fit with the Dash callback architecture, the stringification of `id` objects in pattern-matching callbacks, etc.

alexcjohnson:
The only reason to want `id` in the DOM, AFAIK, is for CSS. For that purpose you could make a `parent_id` if you wanted to, but I'd argue `className` is better for CSS targeting anyway. The pattern-matching callbacks object-ID concept is useless, or at least extremely awkward, for CSS, so this would be string-only.

So sure, you can put `id` in the DOM, and for consistecy of naming I'd put it on the spinner next to `style` and `className`. But there's also no harm leaving it out of the DOM and having its sole purpose be to wire up callbacks.

wbrgss:
> But there's also no harm leaving it out of the DOM and having its sole purpose be to wire up callbacks.

OK, sounds good to me. Let's keep the scope of this to just address the original issue then, and leave out the `id` in the DOM.

alexcjohnson:
@wbrgss this looks good to me. Just needs a changelog entry, and either update the failing test not to use `output=` or wait for me to address https://github.com/plotly/dash/issues/1366

tests/test_integration_1.py
```diff
@@ -43,22 +43,25 @@ def test_loading_slider(self):
...
-            dcc.Slider(
-                id='horizontal-slider',
-                min=0,
-                max=9,
-                marks={i: 'Label {}'.format(i) if i == 1 else str(i)
-                       for i in range(1, 6)},
-                value=5,
-            ),
-        ])
+        app.layout = html.Div(
```
wbrgss, line 46: The only manual changes here were removing the `output=`, `input=`, `state=` keyword arguments in the `test_location_link` test in order to work around plotly/dash#1366; the other changes (double `"` quotes, indentation) are formatting changes from `black`ening this file.

wbrgss:
@alexcjohnson thanks for info re plotly/dash#1366; didn't realize that was the cause of the failing legacy tests. Tests are passing and CHANGELOG has been updated :slightly_smiling_face:

PR #844, opened 2020-08-15
Fix string comparison for lint:black in package.json

noisycomputation:
Fixed non-functional string comparison in `private::lint:black`
by adding `$` prefix to `PYTHON_VERSION`.

Made the conditional POSIX compliant by removing the double
bracket bashism and using escaped double quotes. This is
**not** academic -- the CircleCI image used to build this
package uses `dash` as its default shell, and `[[` doesn't work
in `dash`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "private::format:black": "black --exclude dash_core_components .",
     "private::format.eslint": "eslint src tests --fix",
     "private::format.prettier": "prettier --config .prettierrc --write src/**/*.js tests/unit/*.js",
-    "private::lint:black": "if [[ PYTHON_VERSION != 'py27' ]]; then black --check --exclude dash_core_components .; fi",
+    "private::lint:black": "if [ \"$PYTHON_VERSION\" != \"py27\" ]; then black --check --exclude dash_core_components .; fi",
     "private::lint.eslint": "eslint src tests",
     "private::lint.flake8": "flake8 --exclude=dash_core_components,node_modules,venv",
     "private::lint.prettier": "prettier --config .prettierrc src/**/*.js tests/unit/*.js --list-different",
--- a/package.json
+++ b/package.json
@@ -12,10 +12,10 @@
   "homepage": "https://github.com/plotly/dash-core-components",
   "main": "dash_core_components/dash_core_components.min.js",
   "scripts": {
-    "private::format:black": "black --exclude dash_core_components .",
+    "private::format.black": "black --exclude dash_core_components .",
     "private::format.eslint": "eslint src tests --fix",
     "private::format.prettier": "prettier --config .prettierrc --write src/**/*.js tests/unit/*.js",
-    "private::lint:black": "if [ \"$PYTHON_VERSION\" != \"py27\" ]; then black --check --exclude dash_core_components .; fi",
+    "private::lint.black": "if [ \"$PYTHON_VERSION\" != \"py27\" ]; then black --check --exclude dash_core_components .; fi",
     "private::lint.eslint": "eslint src tests",
     "private::lint.flake8": "flake8 --exclude=dash_core_components,node_modules,venv",
     "private::lint.prettier": "prettier --config .prettierrc src/**/*.js tests/unit/*.js --list-different",
--- a/setup.py
+++ b/setup.py
@@ -1,19 +1,19 @@
 from setuptools import setup
 import json
 
-with open('package.json') as f:
+with open("package.json") as f:
     package = json.load(f)
 
 package_name = str(package["name"].replace(" ", "_").replace("-", "_"))
 
 setup(
-    name='dash_core_components',
+    name="dash_core_components",
     version=package["version"],
-    author=package['author'],
-    author_email='chris@plotly.com',
+    author=package["author"],
+    author_email="chris@plotly.com",
     packages=[package_name],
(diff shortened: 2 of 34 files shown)


alexcjohnson:
Thanks @noisycomputation - I noticed the same thing recently in the Dash repo and adapted in https://github.com/plotly/dash/commit/6edd176ca4e6f0d2ead6b22bbf27a567c69e2ce4 - it didn't look to me as though single quotes are a problem over there but I confess to not knowing a whole lot about shell variants. Anyway this is fine, I'm just curious in case we're doing it wrong elsewhere.

More importantly though, seems `lint:black` also needs to be changed to `lint.black` (and same for `format:black`) so that `"lint": "run-s private::lint.*"` will pick it up!

noisycomputation:
> Thanks @noisycomputation - I noticed the same thing recently in the Dash repo and adapted in [plotly/dash@6edd176](https://github.com/plotly/dash/commit/6edd176ca4e6f0d2ead6b22bbf27a567c69e2ce4) - it didn't look to me as though single quotes are a problem over there but I confess to not knowing a whole lot about shell variants. Anyway this is fine, I'm just curious in case we're doing it wrong elsewhere.

(This is long and is not meant as a lecture, or at least not the patronizing or arrogant kind. The issue comes up enough in various contexts that I wanted to have something to which I can point in the future.)

The escaping mechanisms of early shells (C and Bourne) were clunky, this is probably why there is so much divergence in this area between modern shells. Most people are used to `bash`, which is just about POSIX-y enough to get by in most cases. One exception is the construct `[[ ... ]]`, the contents of which are parsed as a single conditional expression and therefore shell escaping rules are loosened. This is convenient, but the construct is an optional POSIX extension and is only available in a few shells, including `bash`, `zsh`, and `ksh`.

These rules are safe for all POSIX-compliant shells and should be used when one doesn't have control over which shells are available or which shell is being used to execute a script:

1. Use single brackets `[ ... ]` for conditional expressions (**not** double brackets `[[ ... ]]`).
  `if [ "$number" -gt 5 ]; then ...; fi`

2. Quote everything where you do not want the shell to perform [field splitting](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#:~:text=2.6.5%20Field%20Splitting) and [pathname expansion](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#:~:text=2.6.6%20Pathname%20Expansion).

3. Use single quotes if [you don't want the shell to touch the string at all](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#:~:text=2.2.2%20Single%2DQuotes,A%20single%2Dquote%20cannot%20occur%20within%20single%2Dquotes.).
    `printf '%s\n' 'This $will appear exac\tly like this'` yields `This $will appear exac\tly like this`

4. Use double quotes around variables and parameter expansions ([explanation](https://unix.stackexchange.com/questions/68694/when-is-double-quoting-necessary#answer-68748:~:text=What%20happens%20without%20quotes,a%20command%20substitution%20followed%20by%20split%2Bglob.) and [POSIX reference](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#:~:text=2.6%20Word%20Expansions,shall%20always%20be%20performed%20last.)).
  `"$MYVARIABLE"`, `"${MYVARIABLE:-default_value}"`
5. Use single *or* double quotes around literal strings (either use one consistently, or switch to avoid escaping, see next bullet).
    `[ "$VAR" = 'foo' ]` or `[ "$VAR" = "foo" ]`

6. Using single quotes within double quotes and vice versa is permitted.
    `printf '%s\n' "Shell doesn't mind this"`, `printf '%s\n' 'The word "foo"'`

7. To use a literal single quote within single quotes, first close the quotation, then include an escaped single quote, then open another quotation. This is cumbersome but [necessary](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#:~:text=A%20single%2Dquote%20cannot%20occur%20within%20single%2Dquotes.).
    `printf '%s\n' 'Let'\''s do this right'` (breakdown: `'Let'`  `\'`  `'s do this right'`)

8. To use a literal double quote within double quotes, just escape the double quote.
    `printf "if [ \"$PYTHON_VERSION\" != 'py27' ]; then ...; fi"`
    >Some people prefer to follow the previous rule for double quotes. POSIX does not require this, but the C shell and its derivatives do.

You'll note in the last example that I put the literal string in single quotes, whereas I placed it in double quotes in my pull request. In my personal usage, I tend to use double quotes unless I affirmatively want to prevent the shell from doing any expansion. In this case, single quotes are probably neater because they do not require escaping.

As for the conditional statement in plotly/dash@6edd176](https://github.com/plotly/dash/commit/6edd176ca4e6f0d2ead6b22bbf27a567c69e2ce4), the lack of double quotes around the parameter expansion `${PYLINTRC:-x}` will cause the shell to throw an error if that variable contains any of the IFS characters (by default, a space, tab, or newline). If the variable is enclosed in double quotes, the default value substitution can be omitted: `if [ "$PYLINTRC" != '.pylintrc' ]` will evaluate correctly even if PYLINTRC is unset or null. Leaving the line as is will probably work without any problems for a long time, but if the PYLINTRC variable becomes mangled at some point, it will prevent CI builds from succeeding.

Issue #700, opened 2019-11-14
dcc.DatePickerSingle() persistence not working


jezlax:
I'm on dcc 1.5.1 ... I have a date picker single which populates a date using the following... 

`dcc.DatePickerSingle(className = 'no-print',
                                         id='choose-date',
                                         date = datetime.today() - timedelta(hours=7),
                                         persistence=True,
                                         persistence_type='session')`

however the persistence does not take. In this example, the date picker populates a set of links (html.A()) elements for the given date. The user then selects the link of choice to populate the remainder of the charts on the app. 

While the date picker works fine, when the user selects the link, it triggers an HREF callback and the date picker resets to the default date. My goal is to have the date persist until the user exits the page.

alexcjohnson:
Thanks for the report, investigating... that said, since the link is to a different view within the same dash app you may be able to do this more easily using `dcc.Link` and `dcc.Location` https://dash.plot.ly/urls and leaving the date picker outside the part of the page that's changed.

alexcjohnson:
@jezlax I'm guessing this `DatePickerSingle` component is generated either in a layout *function* or in a callback - ie `datetime.today() - timedelta(hours=7)` is changing every time it's called because it goes all the way to microseconds - it gets serialized something like `date: "2019-11-14T14:45:12.741865"`. This is a problem for persistence because we'll only reapply the user's choice if the initial value provided with the component is the same as before the user made that choice.

If you strip out the (unused) time portion - something like `(datetime.today() - timedelta(hours=7)).date()` - then it will give a consistent value for 24 hours - so if the user keeps the app open for multiple days it'll reset (on reload) once a day, but the persistence should be maintained until that point.

jezlax:
Thanks @alexcjohnson - I will dev this rework today and let you know.

Marc-Andre-Rivet:
@harryturr As discussed IRL, this is actionable in at least 3 ways

* We can modify the persistence logic so that we only extract/apply the `date` portion of the datetime - this should provide 24 hours stability without any additional code from the users
* We should plan on adding some best practice info to the docs, specifically about explicit timezones information on the datetime passed to Dash. This is useful if the user/server are not located in the same timezone
* Might be interesting to consider some alternate values to represent today / tomorrow. It might be that for many use cases `today` or `today+/-N` might be an interesting constant value to pass to the component - to be processed, and would provide a constant baseline for persistence

Marc-Andre-Rivet:
[...]

PR #848, opened 2020-08-21
PersistenceTransforms for date in datePickerRange and datePickerSingle

harryturr:
In conjunction with https://github.com/plotly/dash/pull/1376, fixes https://github.com/plotly/dash-core-components/issues/700. 

The persistence logic in `DatePickerRange` and `DatePickerSingle` have been modified to extract/apply just the `date` portion of the `date-time`. This will provide stability to date pickers (eg defined in callbacks) for the entire day they are defined.


.circleci/config.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           command: |
             . venv/bin/activate && mkdir packages
             # build main dash
-            git clone --depth 1 https://github.com/plotly/dash.git dash-main
+            git clone -b persistence-hg --depth 1 https://github.com/plotly/dash.git dash-main
```
harryturr, line 83: Testing with Dash branch https://github.com/plotly/dash/pull/1376.

PR #850, opened 2020-08-22
Enhancement: add property Graph.prependData to support Plotly.prependTraces

sleighsoft:
Very similar to https://github.com/plotly/dash-core-components/pull/461 but adds `prependData` capabilities which are already supported  by `Plotly.prependTraces`.


sleighsoft:
Linting fails with the same message as in circleci on my local machine

sleighsoft:
Can someone rerun this, looks to me like a fluke

sleighsoft:
Would love to see this feature added in the near future!

alexcjohnson:
Also we'll need a changelog entry

sleighsoft:
Reworked it based on your suggestions. Let me know what you think :)

src/components/Graph.react.js
```diff
@@ -55,22 +83,23 @@ class PlotlyGraph extends Component {
...
-        this.setState(({extendData}) => {
+    clearState(dataKey) {
+        this.setState(props => {
+            var data = props[dataKey];
             const res =
-                extendData && extendData.length
+                data && data.length
                     ? {
-                          extendData: EMPTY_EXTEND_DATA,
+                        data: EMPTY_DATA,
```
alexcjohnson, line 102:
```suggestion
                        [dataKey]: EMPTY_DATA,
```

Not sure how that would cause the prepend tests to fail while the extend tests still succeed... there may be something else funny going on. Aside from the fact that it broke the tests, yes this refactoring is exactly what I had in mind 😅 

Might be easier to debug those locally with `pytest -k test_graph_prepend_trace` or, if you want to play with the test app before the automation commands run, `pytest -k test_graph_prepend_trace --pause`

(also the lint test failed, but that can be fixed with with `npm run format`)

sleighsoft: Not super familiar with JavaScript yet. What is that `[..]` syntax you are proposing?

sleighsoft: Is there a guide on how to get the whole Chrome thing working for tests? I am on Windows with WSL and have not got it working yet.

sleighsoft: When running the prepend and extend examples isolated on my machine (I bascially extracted the test code into a separate dash app .py file) then it works for both prepend and extend. Weird.

sleighsoft: Finally got selenium working after upgrading to WSL 2. Will look into this in the coming days.

sleighsoft:
So I ran the extend test at commit: 0ceb2deae5e855847ab12a5f8ea6277b67afffa9
And it failed there already.

sleighsoft:
For whatever reason the following will actually prepend/extend 3 times:

https://github.com/plotly/dash-core-components/pull/850/files#diff-5d342d454f72aac2300a2066db0defb3R251
```python
    layout.append(
        dcc.Interval(
            id="interval_prependablegraph_prependtwice",
            interval=500,
            n_intervals=0,
            max_intervals=2,
        )
    )
```

Though, this is not something that I added.

sleighsoft:
Yeah, looks like for whatever reason data is no prepended/extended an additional time ...

src/components/Graph.react.js
sleighsoft: Oh it took me so long to understand your suggestion... Total Javascript/React rooky here

sleighsoft:
Failing tests do not seem related to my changes?

sleighsoft:
Added Changelog entry.

alexcjohnson:
@sleighsoft the failing legacy tests have to do with a behavior change in the latest dash version, and they've been fixed in the dev branch. Can you merge dev back into this branch (and fix the changelog conflict)? That should fix these, then we can figure out what's going on with the new tests.

alexcjohnson:
(The changelog conflict is normal, we just released a new dash & dcc version today)

sleighsoft:
I hope this is it now :)

sleighsoft:
Legacy still fails

Marc-Andre-Rivet:
> @Marc-Andre-Rivet I don't imagine you'll be too excited about extendData getting a friend, given the headaches it creates, but do you have any other thoughts about this PR?

@alexcjohnson Haha. Actually I like the unified solution + the extra test. Supporting both vs, one shouldn't make much of a difference 😄  My only concern atm is the `https://` --> `http://` change in the package-lock.json file. Seems that could just be 🔪  out.

alexcjohnson (approved):
💃 After a little test tweaking (unrelated to your code, these were just old tests that needed some modernizing to be more robust) and undoing the `package-lock.json` changes you had (which also happen for me locally, still don't get what's going on there) I think we're ready. Thanks for all your work @sleighsoft !

sleighsoft:
Thank you for your help @alexcjohnson !

PR #854, opened 2020-08-25
PersistenceTransforms for date in datePickerRange and datePickerSingle

harryturr:
In conjunction with https://github.com/plotly/dash/pull/1376, fixes https://github.com/plotly/dash-core-components/issues/700. 

The persistence logic in `DatePickerRange` and `DatePickerSingle` have been modified to extract/apply just the `date` portion of the `date-time`. This will provide stability to date pickers (eg defined in callbacks) for the entire day they are defined.


src/components/DatePickerRange.react.js
```diff
@@ -263,6 +264,31 @@ DatePickerRange.propTypes = {
     persistence_type: PropTypes.oneOf(['local', 'session', 'memory']),
 };
 
+DatePickerRange.persistenceTransforms = {
+    end_date: {
+        extract: propValue => {
+            if (!(propValue === null || propValue === undefined)) {
```
Marc-Andre-Rivet, line 270: This check is fine but can be simplified by using https://ramdajs.com/docs/#isNil instead

src/components/DatePickerRange.react.js
```diff
@@ -263,6 +264,31 @@ DatePickerRange.propTypes = {
...
+    start_date: {
+        extract: propValue => {
+            if (!(propValue === null || propValue === undefined)) {
+                return moment(propValue)
+                    .startOf('day')
+                    .format('YYYY-MM-DD');
+            }
+            return propValue;
+        },
+        apply: storedValue => storedValue,
```
Marc-Andre-Rivet, line 288: The implementation for `end_date` and `start_date` extract and apply are identical (same for DatePickerSingle `date`) -- DRY by defining a functions that you can share between the 3 cases

tests/integration/calendar/test_date_picker_persistence.py
```diff
@@ -0,0 +1,158 @@
...
+    def display_dps2(value):
+        return value
+
+    @app.callback(Output("dpr2-p-end", "children"), [Input("dpr2", "end_date")])
+    def display_dps2(value):
+        return value
+
+    dash_dcc.start_server(app)
+    dpr1 = dash_dcc.find_element("div#dpr1 div div div div .DateInput_input")
+    dpr2 = dash_dcc.find_element("div#dpr2 div div div div .DateInput_input")
```
Marc-Andre-Rivet, line 143: Does it work with only `div#dpr2 .DateInput_input` ? This `find_element` seems to be very dependent on the implementation details of the underlying third-party component

harryturr: See 605594c . Now we use only the `dash_dcc` functions for interacting with date pickers, and are testing the behaviour DatePickerSingle and Range defined in callbacks, with the `date`, `start/end_date` props changing value each time the callback is fired.

src/components/DatePickerRange.react.js
harryturr: 57e5019

src/components/DatePickerRange.react.js
harryturr: 32852c0

.circleci/config.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           command: |
             . venv/bin/activate && mkdir packages
             # build main dash
-            git clone --depth 1 https://github.com/plotly/dash.git dash-main
+            git clone -b persistence-hg --depth 1 https://github.com/plotly/dash.git dash-main
```
Marc-Andre-Rivet, line 83: With https://github.com/plotly/dash/pull/1376 now merged, this can be 🔪

src/components/DatePickerRange.react.js
```diff
@@ -263,6 +266,11 @@ DatePickerRange.propTypes = {
     persistence_type: PropTypes.oneOf(['local', 'session', 'memory']),
 };
 
+DatePickerRange.persistenceTransforms = {
+    end_date,
```
Marc-Andre-Rivet, line 270: Assign the value to the prop instead of assigning to a value that matches the prop for shortcut assignation
```suggestion
    end_date: transformDate,
```

Marc-Andre-Rivet (commented):
- now needs to be run against dash `dev`
- needs a changelog entry

src/components/DatePickerRange.react.js
harryturr: c0f98b4

.circleci/config.yml
harryturr: 6df0230

PR #858, opened 2020-09-03
Plotly.js 1.55.1

Marc-Andre-Rivet:


nicolaskruchten:
1.55.1 just came out

tests/test_integration_2.py
```diff
@@ -70,7 +70,7 @@ def display_page(pathname):
         time.sleep(2)
 
         # callback is called twice when defined
-        self.assertEqual(call_count.value, 2)
+        self.assertEqual(call_count.value, 1)
```
alexcjohnson, line 73: What happened here? Is this related to https://github.com/plotly/dash/pull/1385 ? We should fix the comments on these two changed lines once this is understood. Any chance this means https://github.com/plotly/dash/issues/883 or https://github.com/plotly/dash/issues/1049 has incidentally been resolved?

alexcjohnson (approved):
💃 once the test changes (unrelated to the plotly.js upgrade) are understood and the comments fixed.

tests/test_integration_2.py
Marc-Andre-Rivet: @alexcjohnson Yes. This is caused by plotly/dash#1385.

1. The initial callback gets added to the list of requested callbacks
2. Before `await wait(0);` resolves the `dcc.Location` component renders and triggers a `setProps` because its props are not populated / misaligned with the current url (before, the callback would have been executing before the component rendered)
3. An additional callback is added to the list of requested callbacks
4. `await wait(0);` resolves
5. There are now two callbacks for `test-url.pathname` and the oldest one gets pruned

astrowonk: I'm still seeing plotly/dash#1049 running what I think is the latest version of everything. Is it supposed to be resolved as of this merge?

alexcjohnson: The boundaries of the problem have changed a little, as evidenced by this test, but it’s definitely not fixed.

PR #865, opened 2020-09-15
Add `_dashprivate_onFigureModified`

Marc-Andre-Rivet:


src/fragments/Graph.react.js
```diff
@@ -352,6 +353,14 @@ class PlotlyGraph extends Component {
         if (this.props.extendData) {
             this.mergeTraces(this.props, 'extendData', 'extendTraces');
         }
+
+        if (
+            (this.props.prependData?.length ?? 0) +
+                (this.props.extendData?.length ?? 0) >
+            0
```
alexcjohnson, line 360: Just curious, is there a reason you prefer this to
`if (this.props.prependData?.length || this.props.extendData?.length)`
?

alexcjohnson (approved):
💃

src/fragments/Graph.react.js
Marc-Andre-Rivet: None whatsoever, will update.

Marc-Andre-Rivet: https://github.com/plotly/dash-core-components/pull/865/commits/a156dccb05f916968d0c2170d9a7a86a9b8acb8a

PR #866, opened 2020-09-17
Remove julia artifacts from package lib/

Marc-Andre-Rivet:
Ran into this issue during Dash 1.16.1 release

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
   },
   "files": [
     "/dash_core_components/*{.js,.map}",
-    "/lib/**"
+    "/lib/**/!(*.jl)"
   ],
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -102,7 +102,7 @@
   },
   "files": [
     "/dash_core_components/*{.js,.map}",
-    "/lib/**"
+    "/lib/**/!(*.jl)"
```
Marc-Andre-Rivet, line 105: Can't exclude the Julia artifacts with `babel` since we're using `--copy-files` for non-`/(j|t)s(x)?/ files, excluding from npm package instead

Issue #751, opened 2020-02-06
slider tooltip.always_visible doesn't position properly


michaelbabyn:
If I put a slider into a div that has overflow-y=scroll, then the tooltip doesn't scoll with the slider

![Screenshot from 2020-02-06 13-05-47](https://user-images.githubusercontent.com/41019918/73965394-10ce8480-48e2-11ea-8a0a-4f9eda92a652.png)

```python
import dash
import dash_core_components as dcc
import dash_html_components as html

app = dash.Dash(
    __name__,
)


app.layout = html.Div([
    html.Div([
        html.Div(dcc.Slider(
            min=0,
            max=100,
            value=65,
            tooltip={'always_visible':True, 'placement':'bottom'}
        ),
        style=dict(height=100))
    ]*5,
    style=dict(maxHeight=300, overflowX='scroll'))
])


if __name__ == '__main__':
    app.run_server(debug=True)
```

wbrgss:
possible related to react-component/slider/issues/374

alb-erandelcastillo:
Since dcc.Slider is calling the react tooltip API, what would be the correct container to set for the tooltip when using getTooltipContainer?

alb-erandelcastillo:
Since dcc.Slider is calling the react tooltip API, what would be the correct container to set for the tooltip when using getTooltipContainer?

wbrgss:
There is a fix here in the original/official version of this repo: https://github.com/ant-design/ant-design/pull/16717. We could make a PR onto react-component/slider upstream or try to migrate `dcc.Slider` to the original (https://github.com/ant-design/ant-design/tree/master/components/slider) (cc @chriddyp if you were planning to take a look at this)

wbrgss:
A short term fix would be to document around it with the suggestion here https://github.com/react-component/slider/issues/374#issuecomment-426086672

> getTooltipContainer works as long as the slider has a parent that has position relative

jdamiba:
I'm currently taking a look at this. This is a non-exhaustive list of some things I'll want to figure out:

- [ ] figure out what combination of css rules is leading to this
- [ ] identify a combination of css rules that lead to the desired behavior - presumably the tooltip should only be visible when parent slider is visible and the tooltip should scroll with the parent slider has the `overflow=y` style
- [ ] understand how this bug was fixed in the ant-design component
- [ ] figure out how to update `dcc_slider` to implement the desired behavior - patch dcc_slider or migrate it entirely to another implementation?

PR #878, opened 2020-10-27
make sure tooltips aren't visible when parent slider isn't

jdamiba:
closes https://github.com/plotly/dash-core-components/issues/751

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Updated
 - [#878](https://github.com/plotly/dash-core-components/pull/878)
-  - used the `getTooltipContainter` prop of dcc.Slider to make sure tooltips are positioned correctly when the parent slider is not visible due to it being in a HTML element with an `overflow-x` style, fixing [#751](https://github.com/plotly/dash-core-components/issues/751)
+  - Fixed [#751](https://github.com/plotly/dash-core-components/issues/751), a bug that causes `dcc.Slider` and `dcc.RangerSlider` tooltips to be visible even if the slider component isn't visible (e.g. overflow),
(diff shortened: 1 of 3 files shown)


Marc-Andre-Rivet (changes requested):
@jdamiba This PR will need:

- a changelog.md entry for the bug being fixed
- a test that shows this works -- I think in this case a simple visual test saved in Percy will do the trick. Can be based on the issue's example code. No need for interactivity or anything like that. 
- I suggest you add a comment in the code to the effect that if/when we'll update rc-slider / rc-tooltip we will need to revisit this code as the getTooltipContainer function will need to be a prop instead of a nested property

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - [#871](https://github.com/plotly/dash-core-components/pull/871) Add Julia syntax highlighting support for dcc.Markdown
 
 ## Updated
+- [#878](https://github.com/plotly/dash-core-components/pull/878)
+  - used the `getTooltipContainter` prop of dcc.Slider to make sure tooltips are positioned correctly when the parent slider is not visible due to it being in a HTML element with an `overflow-x` style, fixing [#751](https://github.com/plotly/dash-core-components/issues/751)
```
Marc-Andre-Rivet, line 11: I think I missed the wrong level of details of https://github.com/plotly/dash-core-components/blame/dev/CHANGELOG.md#L26 which you are using as your baseline.

Make the entry for focused on what this fixes for the user. Something along the lines of:

Fix a bug causing the dcc.Slider and dcc.RangerSlider tooltip to be visible even if the component isn't visible (e.g. overflow)

Marc-Andre-Rivet (approved):
🎚️ 🚢

PR #884, opened 2020-11-17
Adding drag_value to Slider

### LINKED ISSUE #107: Ability to have different Outputs react on either a 'drag' or 'mouseup' for a single Slider
Would it be feasible to have two different Inputs from a slider:
 'drag-value' and a 'mouseup-value'

This would then allow the user to, for example, update a label that reflects the value of the slider with a drag and then update a longer callback on mouseup.

almarklein:
Closes #107

This is an attempt to add a `drag_value` property to the Slider, such that different callbacks can be applied to the slider being dragged, and the slider being released, as proposed in #107.

Disclaimer: I am not yet familiar with the codebase of this repo, so I just tried to update the new property where the `value` property was updated. 

Todo:

* [x] The value is initially `undefined`, and I'm not sure how to initialize it to the same value as `value`.
* [ ] Also implement this for RangeSlider.
* [x] update the docstring for `updatemode` to incorporate the new prop.
* [ ] More docs.


almarklein:
cc @emmanuelle

emmanuelle:
@alexcjohnson @Marc-Andre-Rivet there are still some todo items in this PR but could you please comment whether the PR is going in the right direction?

alexcjohnson:
Yes, I think this is reasonable. One possible concern is performance - will there be a problem in very large apps with just having this prop exist even if it isn't used? But I think the `hoverData` prop in `dcc.Graph` shows that this is OK.

We should update the docstring for `updatemode` to incorporate the new prop.

> Current changes are only of the src, should this also include changes from after running the built?

No, just `src` should be changed in this PR. Build artifacts will be updated when we make the next release.

almarklein:
Oh oh, it looks like I did a git screw-up. I'm closing this in favor of #888. Sorry!

PR #887, opened 2020-12-01
Update contributing.md to remove reference to outdated components archetype

almarklein:


CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@ Refer to the [readme](README.md) for installation and development instructions.
 
 ## Contributions
 
-[Dash Core Components][] consist of pluggable components for creating interactive user interfaces. For generic HTML5 elements, see [Dash HTML Components][]. Contributions are welcome! This repository's open [issues][] are a good place to start. Another way to contribute is to [write your own components][] using the [Dash Components Archetype][].
+[Dash Core Components][] consist of pluggable components for creating interactive user interfaces. For generic HTML5 elements, see [Dash HTML Components][]. Contributions are welcome! This repository's open [issues][] are a good place to start. Another way to contribute is to [write your own components][].
```
rpkyle, line 9: Thanks for this edit @almarklein! I wonder if we might mention `dash-component-boilerplate` here instead with a link to [the repository](https://github.com/plotly/dash-component-boilerplate)?

rpkyle (changes requested):

CONTRIBUTING.md
almarklein: I'm quite new to dcc, but that looks like a good replacement!

rpkyle (approved):

almarklein:
> I think instead of "using for instance" here I might suggest something like

I just see that I overlooked your suggestion. Sorry!

PR #888, opened 2020-12-01
Adding drag_value to Slider

### LINKED ISSUE #107: Ability to have different Outputs react on either a 'drag' or 'mouseup' for a single Slider

almarklein:
closes #107, replaces #884

This is an attempt to add a `drag_value` property to the Slider, such that different callbacks can be applied to the slider being dragged, and the slider being released, as proposed in #107.

Disclaimer: I am not yet familiar with the codebase of this repo, so I just tried to update the new property where the `value` property was updated. 

Todo:

* [x] The value is initially `undefined`, and I'm not sure how to initialize it to the same value as `value`.
* [x] Also implement this for RangeSlider.
* [x] update the docstring for `updatemode` to incorporate the new prop.
* [x] `RangeSlicer` should be updated to use the same approach as `Slicer`.
* [x] Tests.
* [ ] More docs.

----

Script to test this:

```py
import dash


app = dash.Dash(__name__, update_title=None)

app.layout = html.Div(
    [
        #dcc.Slider(id='slider', min=100, max=200, value=150),
        dcc.RangeSlider(id='slider', min=100, max=200, value=(133, 166)),
        html.Br(),
        html.Div(id='drag_value', children="drag_value:"),
        html.Div(id='value', children="value:"),
    ]
)

counts = {"drag_value": 0, "value": 0}

@app.callback(
    Output("drag_value", "children"),
    [Input("slider", "drag_value")],
)
def show_slider_drag_value(v):
    counts["drag_value"] += 1
    return f"drag_value: {v}  ({counts['drag_value']})"


@app.callback(
    Output("value", "children"),
    [Input("slider", "value")],
)
def show_slider_value(v):
    counts["value"] += 1
    return f"value: {v}  ({counts['value']})"


if __name__ == "__main__":
    app.run_server(debug=True)
```

### DIFF
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -135,8 +135,9 @@ Slider.propTypes = {
      * will only trigger its value when the user has finished
      * dragging the slider. If `drag`, then the slider will
      * update its value continuously as it is being dragged.
-     * Note that for the latter case, the `drag_value`
-     * property could be used instead.
+     * If you want different actions during and after drag,
+     * leave `updatemode` as `mouseup` and use `drag_value`
+     * for the continuously updating value.
      */
     updatemode: PropTypes.oneOf(['mouseup', 'drag']),
 
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -141,7 +141,7 @@ Slider.propTypes = {
     updatemode: PropTypes.oneOf(['mouseup', 'drag']),
 
     /**
-     * Dash-assigned callback that gets fired when the value changes.
+     * Dash-assigned callback that gets fired when the value or drag_value changes.
      */
     setProps: PropTypes.func,
 
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -130,13 +130,13 @@ Slider.propTypes = {
     verticalHeight: PropTypes.number,
 
     /**
-     * Determines when the component should update
-     * its value. If `mouseup`, then the slider
-     * will only trigger its value when the user has
-     * finished dragging the slider. If `drag`, then
-     * the slider will update its value continuously
-     * as it is being dragged.
-     * Only use `drag` if your updates are fast.
+     * Determines when the component should update its `value`
+     * property. If `mouseup` (the default) then the slider
+     * will only trigger its value when the user has finished
+     * dragging the slider. If `drag`, then the slider will
+     * update its value continuously as it is being dragged.
+     * Note that for the latter case, the `drag_value`
+     * property could be used instead.
      */
     updatemode: PropTypes.oneOf(['mouseup', 'drag']),
 
--- a/src/components/Slider.react.js
+++ b/src/components/Slider.react.js
@@ -49,6 +49,11 @@ Slider.propTypes = {
     value: PropTypes.number,
 
     /**
+     * The value of the input during a drag
+     */
+    drag_value: PropTypes.number,
+
+    /**
      * Additional CSS class for the root DOM node
      */
     className: PropTypes.string,
--- a/src/fragments/Slider.react.js
+++ b/src/fragments/Slider.react.js
@@ -23,6 +23,7 @@ export default class Slider extends Component {
             value: props.value,
             drag_value: props.value,
         };
+        this.props.setProps({drag_value: props.value});
     }
 
     propsToState(newProps) {
@@ -117,6 +118,7 @@ export default class Slider extends Component {
                             'setProps',
                             'updatemode',
                             'value',
+                            'drag_value',
                             'marks',
                             'verticalHeight',
                         ],
--- a/src/fragments/Slider.react.js
+++ b/src/fragments/Slider.react.js
@@ -18,7 +18,6 @@ export default class Slider extends Component {
             ? createSliderWithTooltip(ReactSlider)
             : ReactSlider;
         this._computeStyle = computeSliderStyle();
-        this.drag_value = props.value;
         this.state = {
             value: props.value,
             drag_value: props.value,
--- a/src/fragments/Slider.react.js
+++ b/src/fragments/Slider.react.js
@@ -17,9 +17,7 @@ export default class Slider extends Component {
             ? createSliderWithTooltip(ReactSlider)
             : ReactSlider;
         this._computeStyle = computeSliderStyle();
-        this.state = {
-            value: props.value
-        };
+        this.state = {value: props.value};
     }
 
     UNSAFE_componentWillReceiveProps(newProps) {
@@ -35,8 +33,10 @@ export default class Slider extends Component {
     }
 
     UNSAFE_componentWillMount() {
-        this.props.setProps({drag_value: this.props.value});
-        this.setState({value: this.props.value});
+        if (this.props.value !== null) {
+            this.props.setProps({drag_value: this.props.value});
+            this.setState({value: this.props.value});
+        }
     }
 
     render() {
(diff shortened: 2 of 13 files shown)


src/components/Slider.react.js
```diff
@@ -125,18 +130,18 @@ Slider.propTypes = {
...
-     * the slider will update its value continuously
-     * as it is being dragged.
-     * Only use `drag` if your updates are fast.
+     * Determines when the component should update its `value`
+     * property. If `mouseup` (the default) then the slider
+     * will only trigger its value when the user has finished
+     * dragging the slider. If `drag`, then the slider will
+     * update its value continuously as it is being dragged.
+     * Note that for the latter case, the `drag_value`
+     * property could be used instead.
```
alexcjohnson, line 139:
```suggestion
     * If you want different actions during and after drag,
     * leave `updatemode` as `mouseup` and use `drag_value`
     * for the continuously updating value.
```
Just trying to be a little clearer on the purpose and intended usage.

alexcjohnson:
@almarklein this PR is looking good! Seems like mainly it just needs a test or two added to [test_sliders.py](https://github.com/plotly/dash-core-components/blob/dev/tests/integration/sliders/test_sliders.py). That may be a bit tricky to get right, but seems like the test app would have a callback on both `value` and `drag_value`, then the test would require `ActionChains` like what we have for [`click_at_coord_fractions`](https://github.com/plotly/dash/blob/ebe10a7772673d425c86b3ef3f42a5f69c7e5755/dash/testing/browser.py#L551) and presumably using `click_and_hold` then `move_by_offset`, then waiting for the right results to appear (and not appear) on the page, finally `release` and waiting for the other result to appear on the page.

src/fragments/Slider.react.js
```diff
@@ -20,12 +20,14 @@ export default class Slider extends Component {
         this._computeStyle = computeSliderStyle();
         this.state = {
             value: props.value,
+            drag_value: props.value,
```
alexcjohnson, line 23: Does `drag_value` actually need to be included in `state`? I think that's what `this.state.value` already is...

almarklein:
I refactored the code for the `Slider` a bit based on your suggestions. Using `slider.value` as an output now also updates `slider.drag_value`. This works fine locally (both in an example, and when running the test), but so far I have can't get CI working. Could you have a look at the current changes in `fragments/Slider.react.js`, and see if I am possibly missing something?

I have not updated the `RangeSlider` yet, not added a test for it, let's first get the slider working ...

src/fragments/Slider.react.js
almarklein: I removed it.

almarklein:
BTW, with `assert dash_dcc.find_element("#out-drag-value").text == "You have dragged 5"` I got to see the actual value: "You have dragged None".

Marc-Andre-Rivet:
@alexcjohnson @almarklein I think the problem here is that the branch is out-of-date `1.13.0` https://github.com/almarklein/dash-core-components/blob/drag_value2/package.json#L3 and the current version is `1.14.1`. Updating the branch will probably fix the tests.

You can see in the `Run Integration Tests` step of `test-37` that the installed version of DCC is `1.14.1`

```
Name: dash-core-components
Version: 1.14.1
Summary: Core component suite for Dash
Home-page: UNKNOWN
Author: Chris Parmer <chris@plotly.com>
Author-email: chris@plotly.com
License: MIT
Location: /home/circleci/project/venv/lib/python3.7/site-packages
Requires: 
Required-by: dash
```

alexcjohnson:
Ah, so perhaps installing dash caused it to go fetch the published dcc 1.14.1 rather than use the source in this branch, because it listed a lower version number? I'm still confused by this, because even in the [build-dash step on your merge commit](https://app.circleci.com/pipelines/github/plotly/dash-core-components/615/workflows/1e0720f3-f399-4811-8396-935560ff85bc/jobs/26387) I see it downloading a published version of dcc
```
Collecting dash-core-components==1.14.1 (from dash==1.18.1)
  Downloading https://files.pythonhosted.org/packages/0f/ab/5ffeeed41117383d02485f5b9204dcfaa074bfbb3ff2559afac7b904ad5c/dash_core_components-1.14.1.tar.gz (3.5MB)
```

So how do I know we're not using that version (other than that the test passes? Is there some way we could definitively check at the beginning of the test that the version of this repo that we're running is precisely the intended commit? Or set this up in a more robust way so this issue can't recur?

Marc-Andre-Rivet:
@alexcjohnson Not sure about that. It's that way because we use the tarball from the previous build step -- in other repos we simply do `pip install -e .` and so the error would be more obvious.

almarklein:
Wow, thanks @Marc-Andre-Rivet !

Marc-Andre-Rivet:
@alexcjohnson I think we could simply `pip uninstall dash-core-components -y` before installing the local build and see it 💥 if misaligned.

almarklein:
🎆

almarklein:
About documentation, I think we'd need to add an example to the docs at the dash-docs repo, right?

tests/integration/sliders/test_sliders.py
```diff
@@ -160,3 +160,82 @@ def test_slsl005_slider_tooltip(dash_dcc):
...
+    slider = dash_dcc.find_element("#slider")
+
+    dash_dcc.wait_for_text_to_equal("#out-value", "You have selected 5")
+    dash_dcc.wait_for_text_to_equal("#out-drag-value", "You have dragged 5")
+
+    dash_dcc.click_and_hold_at_coord_fractions(slider, 0.25, 0.25)
+    dash_dcc.move_to_coord_fractions(slider, 0.75, 0.25)
+    dash_dcc.wait_for_text_to_equal("#out-drag-value", "You have dragged 15")
+    dash_dcc.move_to_coord_fractions(slider, 0.5, 0.25)
+    dash_dcc.wait_for_text_to_equal("#out-drag-value", "You have dragged 10")
```
alexcjohnson, line 200:
```suggestion
    dash_dcc.wait_for_text_to_equal("#out-drag-value", "You have dragged 10")
    dash_dcc.wait_for_text_to_equal("#out-value", "You have selected 5")
```

Just to be explicit that `#out-value` doesn't change until the `release`.

tests/integration/sliders/test_sliders.py
```diff
@@ -160,3 +160,82 @@ def test_slsl005_slider_tooltip(dash_dcc):
...
+
+    dash_dcc.start_server(app)
+    slider = dash_dcc.find_element("#slider")
+
+    dash_dcc.wait_for_text_to_equal("#out-value", "You have selected 5-15")
+    dash_dcc.wait_for_text_to_equal("#out-drag-value", "You have dragged 5-15")
+
+    dash_dcc.click_and_hold_at_coord_fractions(slider, 0.25, 0.25)
+    dash_dcc.move_to_coord_fractions(slider, 0.5, 0.25)
+    dash_dcc.wait_for_text_to_equal("#out-drag-value", "You have dragged 10-15")
```
alexcjohnson, line 239:
```suggestion
    dash_dcc.wait_for_text_to_equal("#out-drag-value", "You have dragged 10-15")
    dash_dcc.wait_for_text_to_equal("#out-value", "You have selected 5-15")
```

alexcjohnson (approved):
Fantastic! 💃 

Yes, the final step will be to add an example in `dash-docs`.

PR #889, opened 2020-12-03
Bump Plotly.js to v1.58.0

archmoj:
Similar to #875.

@Marc-Andre-Rivet 
Here are some TODOs:
 - please update CHANGELOG in respect to https://github.com/plotly/plotly.js/releases/tag/v1.58.0 
 - please check if you get the same diff for `package-lock.json`?
 - watch out as it is my very first PR to this repo.
Thanks!

cc: @nicolaskruchten @alexcjohnson

PR #891, opened 2020-12-03
Robustify test `stdl001`

Marc-Andre-Rivet:
Issue encountered as part of reviewing #889

stdl001 fails from time to time because the field's text hasn't had time to update itself.

Update to use wait_for_text_to_equal instead of an assertion.


alexcjohnson (approved):
Nice find. 💃

harryturr (approved):

PR #902, opened 2020-12-09
version bump 1.14.1

harryturr:
Bump to v1.14.1. Remove `./lib/*jl` files and add to  `.gitignore` cc @Marc-Andre-Rivet


.gitignore
```diff
@@ -1,6 +1,7 @@
 .build_cache/
 build/
 dist/
+lib/*.jl
```
Marc-Andre-Rivet, line 4: Small note here. The Julia build artifacts we care about are in `src/` the ones in lib were copied over by Babel pre-1.13.0 in dev but the equivalent change in master's .gitignore file hadn't been added.

rpkyle, line 4: Interesting. Are there typically Julia source files in lib? I thought these usually live within src instead.

CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@ Refer to the [readme](README.md) for installation and development instructions.
 
 ## Contributions
 
-[Dash Core Components][] consist of pluggable components for creating interactive user interfaces. For generic HTML5 elements, see [Dash HTML Components][]. Contributions are welcome! This repository's open [issues][] are a good place to start. Another way to contribute is to [write your own components][] using the [Dash Components Archetype][].
+[Dash Core Components][] consist of pluggable components for creating interactive user interfaces. For generic HTML5 elements, see [Dash HTML Components][]. Contributions are welcome! This repository's open [issues][] are a good place to start. Another way to contribute is to [write your own components][] using for instance the [component boilerplate](https://github.com/plotly/dash-component-boilerplate).
```
rpkyle, line 9: I made a suggestion in the PR that merged this update -- could one of you take a look and consider implementing it? Minor textual edit.

rpkyle (approved):
One minor comment above related to text in CONTRIBUTING.md.

.gitignore
Marc-Andre-Rivet: Babel copied from src, these jl files should never have made it there :)

CONTRIBUTING.md
Marc-Andre-Rivet: Sure but not here. Should be done in Dev. Master is a one way street.

PR #905, opened 2020-12-14
Dont clone figure.layout

### LINKED ISSUE #879: figure as state out of sync
In this simple app from @emmanuelle, `figure.layout.shapes` doesn't seem to be updated when `figure` is captured as `State`, meaning that we have to do all sorts of gymnastics with `relayoutData` to work with shapes. 

@alexcjohnson can you help me figure out if this is something in `dcc` or if Plotly.js is doing something odd here, in which case help me translate this into a CodePen that @archmoj can use?

```python
import plotly.graph_objects as go
import dash
import json

app = dash.Dash(__name__)

fig = go.Figure().update_layout(dragmode='drawrect')

app.layout = html.Div([
    dcc.Graph(id='graph', figure=fig),
    html.Pre(id='annotations-data'),
])

@app.callback(
    Output('annotations-data', 'children'),
    [Input("graph", "relayoutData")],
    [State("graph", "figure")]
)
def on_new_annotation(relayout_data, fig):
    return json.dumps(go.Figure(fig).layout.shapes, indent=2)

app.run_server(debug=True)
```

almarklein:
Fixes #879 (hopefully)

This small changes fixes the issue (for e.g. the example below). That said, I cannot oversee whether this potentially breaks other code. In theory, the tests will tell ;)

Example to test this:
```py
import dash


app = dash.Dash(__name__, update_title=None)

fig = {"data": [], "layout": {"dragmode": "drawrect"}}
graph = dcc.Graph(id="graph", figure=fig)

app.layout = html.Div(
    [
        graph,
        html.Br(),
        html.Button(id='button', children="Clone figure"),
        html.Div(id='output', children=""),
    ]
)

app.clientside_callback(
    """function clone_figure(_, figure) {
        let new_figure = {...figure};
        let shapes = new_figure.layout.shapes || [];
        return [new_figure, shapes.length];
    }
    """,
    [Output("graph", "figure"), Output("output", "children")],
    [Input("button", "n_clicks")],
    [State("graph", "figure")],
)



if __name__ == "__main__":
    app.run_server(debug=True)
```

### DIFF
--- a/src/fragments/Graph.react.js
+++ b/src/fragments/Graph.react.js
@@ -131,6 +131,8 @@ class PlotlyGraph extends Component {
         this.getLayoutOverride = this.getLayoutOverride.bind(this);
         this.graphResize = this.graphResize.bind(this);
         this.isResponsive = this.isResponsive.bind(this);
+
+        this.state = {override: {}, originals: {}};
     }
 
     plot(props) {
@@ -227,8 +229,7 @@ class PlotlyGraph extends Component {
             return layout;
         }
         const override = this.getLayoutOverride(responsive);
-        const prev_override = (this.state && this.state.override) || {};
-        const prev_originals = (this.state && this.state.originals) || {};
+        const {override: prev_override, originals: prev_originals} = this.state;
         // Store the original data that we're about to override
         const originals = {};
         for (const key in override) {
@@ -238,7 +239,7 @@ class PlotlyGraph extends Component {
                 originals[key] = prev_originals[key];
             }
         }
-        this.setState({override: override, originals: originals});
+        this.setState({override, originals});
         // Undo the previous override, but only for keys that the user did not change
         for (const key in prev_originals) {
             if (layout[key] === prev_override[key]) {
--- a/src/fragments/Graph.react.js
+++ b/src/fragments/Graph.react.js
@@ -226,8 +226,23 @@ class PlotlyGraph extends Component {
         if (!layout) {
             return layout;
         }
-        let override = this.getLayoutOverride(responsive);
-        for (let key in override) {
+        const override = this.getLayoutOverride(responsive);
+        const prev_layout = this.gd.current.layout;  // === this.props.figure.layout
+        const prev_override_originals = (this.state && this.state.override_originals) || {};
+        // Store the original data that we're about to override
+        const override_originals = {};
+        for (const key in override) {
+            override_originals[key] = layout[key];
+        }
+        this.setState({override_originals: override_originals});
+        // Undo the previous override, but only for keys that the user did not change
+        for (const key in prev_override_originals) {
+            if (layout[key] === prev_layout[key]) {
+                layout[key] = prev_override_originals[key];
+            }
+        }
+        // Apply the current override
+        for (const key in override) {
             layout[key] = override[key];
         }
         return layout;  // not really a clone
--- a/src/fragments/Graph.react.js
+++ b/src/fragments/Graph.react.js
@@ -151,13 +151,12 @@ class PlotlyGraph extends Component {
         }
 
         const configClone = this.getConfig(config, responsive);
-        const layoutClone = this.getLayout(figure.layout, responsive);
 
         gd.classList.add('dash-graph--pending');
 
         return Plotly.react(gd, {
             data: figure.data,
-            layout: layoutClone,
+            layout: figure.layout,
             frames: figure.frames,
             config: configClone,
         }).then(() => {
--- a/src/fragments/Graph.react.js
+++ b/src/fragments/Graph.react.js
@@ -227,19 +227,22 @@ class PlotlyGraph extends Component {
             return layout;
         }
         const override = this.getLayoutOverride(responsive);
-        const prev_layout = this.gd.current.layout; // === this.props.figure.layout
-        const prev_override_originals =
-            (this.state && this.state.override_originals) || {};
+        const prev_override = (this.state && this.state.override) || {};
+        const prev_originals = (this.state && this.state.originals) || {};
         // Store the original data that we're about to override
-        const override_originals = {};
+        const originals = {};
         for (const key in override) {
-            override_originals[key] = layout[key];
+            if (layout[key] !== prev_override[key]) {
+                originals[key] = layout[key];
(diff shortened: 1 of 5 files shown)


almarklein:
Mmm, there seems to be a problem with CI in general.

alexcjohnson:
hmm, not sure what's going on with CI, I'll have to investigate.

We're going to need to do more than just dropping the `getLayout` step though - check out what happens in there with `responsive` and how it modifies `autosize`, `height`, and `width`.

In principle we may need to save the original values of these attributes in `state`, so cases where the user provides a figure and then later changes the `responsive` prop *without* changing the figure itself we can bring those original values back as needed. In practice though I doubt users change `responsive` much.

src/fragments/Graph.react.js
```diff
@@ -226,8 +226,11 @@ class PlotlyGraph extends Component {
         if (!layout) {
             return layout;
         }
-
-        return mergeDeepRight(layout, this.getLayoutOverride(responsive));
+        let override = this.getLayoutOverride(responsive);
+        for (let key in override) {
+            layout[key] = override[key];
+        }
+        return layout;  // not really a clone
```
almarklein, line 233: Something like this? I'm not sure how to keep original state though. If we e.g. store `autosize` here, we cannot simply restore it the next time, because we don't know whether the user changed `autoresize` on the figure. But I may be misunderstanding.

alexcjohnson:
> Mmm, there seems to be a problem with CI in general.

The `dash-html-components` build process broke. I'm investigating and will update you when it's resolved.

alexcjohnson:
Fixed the build process via https://github.com/plotly/dash-html-components/pull/170 - now it's just routine linter errors 😏

src/fragments/Graph.react.js
alexcjohnson: Yes, this looks good for applying the overrides. Then for updates, I guess the logic would go something like:
- We can see if the layout is being changed by looking at what's recorded on the plot div when we're about to replot: `const layoutChanged = gd.layout !== props.layout.figure`
- If `layoutChanged`, store these attributes in state and then apply the overrides.
- Otherwise reapply the original values from state and then apply the overrides (which should give the right result regardless of whether `responsive` changed or not.

almarklein:
Updated, and tests added. Only a linting issue with `prettier`. Unfortunately, it does not show *what* is wrong, and I cannot reproduce the linting locally (and I do not see any obvious errors).

almarklein:
I also updated the code sample in the top post to do about the same thing as the test does.

alexcjohnson:
https://github.com/plotly/dash-core-components/pull/905/commits/eb82ae126248e2bf96f58dfdc2641d43a8a6409c resulted from `npm run format` (fixed the prettier issues), but also for some reason locally I needed to update the `__init__.py` before the pre-commit hooks would pass. 🤷

src/fragments/Graph.react.js
```diff
@@ -226,8 +226,27 @@ class PlotlyGraph extends Component {
...
-
-        return mergeDeepRight(layout, this.getLayoutOverride(responsive));
+        const override = this.getLayoutOverride(responsive);
+        const prev_layout = this.gd.current.layout; // === this.props.figure.layout
+        const prev_override_originals =
+            (this.state && this.state.override_originals) || {};
+        // Store the original data that we're about to override
+        const override_originals = {};
+        for (const key in override) {
+            override_originals[key] = layout[key];
```
alexcjohnson, line 236: Mostly this seems right, but you're pulling out "originals" even if this is the same layout as before - which has already had the overrides applied to it.

almarklein:
Updated again.

I also ran a little test with a layout that has `width` and `height` set to 300. The figure is originally this small size. Turning `responsive` to `true` will make the figure span the available space (and `width` and `height` are `undefined`). Turning `responsive` off again resets the figure to its original small size.

almarklein:
The failing test looks like a glitch in another test.

emmanuelle:
I restarted the CI and all looks good now.

almarklein:
@alexcjohnson also added a test to check that original values are correctly restored.

tests/integration/graph/test_graph_varia.py
```diff
@@ -658,3 +658,115 @@ def load_chart(n_clicks):
...
+        }
+        """,
+        [Output("graph", "figure"), Output("output", "children")],
+        [Input("button", "n_clicks")],
+        [State("graph", "figure")],
+    )
+
+    dash_dcc.start_server(app)
+    button = dash_dcc.wait_for_element("#button")
+    time.sleep(1)
```
alexcjohnson, line 693: Is there a way we can avoid `sleep` here? Elsewhere in graph tests we use things like
`dash_dcc.wait_for_element("#my-graph:not([data-dash-is-loading])")`
(ie the renderer thinks the graph is done) or
`dash_dcc.wait_for_element("#my-graph .main-svg")`
(ie plotly.js has created some of the structure we know the graph should have)

tests/integration/graph/test_graph_varia.py
```diff
@@ -658,3 +658,115 @@ def load_chart(n_clicks):
...
+
+    # Draw a shape
+    dash_dcc.release()
```
alexcjohnson, line 760: why does just `release` create a shape? And where is this releasing? Should this have `click_and_hold_at_coord_fractions` and `move_to_coord_fractions` like above?

almarklein: I think this is a leftover that I forgot to delete after copying from the other example :D

tests/integration/graph/test_graph_varia.py
almarklein: Managed to get rid of these!

emmanuelle:
Hi, any chance to get this PR merged before the next Dash release? Pretty please :-) ?

alexcjohnson:
Yes sorry, it's been waiting on me to play with it a bit, but we'll get it in for the next release.

alexcjohnson (approved):
OK! @almarklein I was concerned about how this would work in various other cases including both mutations and brand new figures and alternating responsive and figure edits, so I expanded on your second test a bit. All works great! 💃

almarklein:
Thanks @alexcjohnson for wrapping this one up :)

PR #907, opened 2020-12-17
use dash loosen-testing-reqs branch and fix linting

alexcjohnson:
sanity for https://github.com/plotly/dash/pull/1506

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
           command: |
             sudo pip install virtualenv --upgrade
             python -m venv venv || virtualenv venv && . venv/bin/activate
-            pip install --progress-bar off -e git+https://github.com/plotly/dash.git@dev#egg=dash[dev,testing]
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           command: |
             sudo pip install virtualenv --upgrade
             python -m venv venv || virtualenv venv && . venv/bin/activate
-            pip install --progress-bar off -e git+https://github.com/plotly/dash.git@dev#egg=dash[dev,testing]
+            pip install --progress-bar off -e git+https://github.com/plotly/dash.git@loosen-testing-reqs#egg=dash[dev,testing]
```
alexcjohnson, line 31: we can revert these branches and merge this PR even before https://github.com/plotly/dash/pull/1506 merges, because all I'm doing here is adding more linting exclusions and making tests more robust.

Marc-Andre-Rivet (approved):

PR #910, opened 2020-12-17
Updating CI Python images

Marc-Andre-Rivet:
Companion to https://github.com/plotly/dash/pull/1507

Need to revert back dash branch when merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
   lint-unit-37: &lint-unit
     working_directory: ~/project
     docker:
-      - image: circleci/python:3.7.6-stretch-node-browsers
+      - image: circleci/python:3.7.9-stretch-node-browsers
         auth:
           username: dashautomation
           password: $DASH_PAT_DOCKERHUB
@@ -50,7 +50,7 @@ jobs:
   lint-unit-36:
     <<: *lint-unit
     docker:
-      - image: circleci/python:3.6.9-stretch-node-browsers
+      - image: circleci/python:3.76.12-stretch-node-browsers
         auth:
           username: dashautomation
           password: $DASH_PAT_DOCKERHUB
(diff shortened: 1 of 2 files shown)


tests/integration/input/test_number_input.py
```diff
@@ -5,16 +5,8 @@
...
     dash_dcc.start_server(ninput_app)
-    for invalid_number in (
-        "10e10000",
-        "e+++eeeeeE-",
-        "120.2.33",
-        "12-.3",
-        "-3.3.3",
-        "0..0",
-    ):
+    for invalid_number in ("10e10000", "e+++eeeeeE-", "12-.3"):
```
Marc-Andre-Rivet, line 8: 120.2.33
-3.3.3
0..0

In Chrome 87 the above entries get processed into valid numbers as the invalid characters are simply removed

Marc-Andre-Rivet: 120.2.33 -> 120.233
-3.3.3 -> -3.33
0..0 -> 0.0

.circleci/config.yml
```diff
@@ -98,7 +98,7 @@ jobs:
             . venv/bin/activate && mkdir packages
             set -eo pipefail
             # build main dash
-            git clone --depth 1 https://github.com/plotly/dash.git dash-main
+            git clone -b browser-capabilities --depth 1 https://github.com/plotly/dash.git dash-main
```
rpkyle, line 101: @Marc-Andre-Rivet Are we planning to keep this (cloning from a branch) going forward, or just for testing here?

Marc-Andre-Rivet: This is just to validate that the changes in https://github.com/plotly/dash/pull/1507 don't break any tests. At least, not in ways we can't explain / fix.

tests/integration/input/test_number_input.py
alexcjohnson: Oh interesting - what "removed" means is those characters are not accepted into the input. You type them and they just don't appear.

Begs the question why they still allow you to type `10e10000` - if they know to refuse a second decimal point, why don't they also refuse another zero past `10e100`? There's no way to continue typing and turn that into a valid number. Likewise `12-.3` makes no sense at all. Really the only non-numbers they _need_ to allow you to type are incomplete numbers, so I'd probably just put a few of those in and take out these others. Like perhaps `-.`, `12e-`. Otherwise we'll just see this problem again when they tighten up their logic further.

alexcjohnson (approved):
💃

PR #914, opened 2021-01-17
update docstring to work with dash #1531

AnnMarieW:
This small change is needed to work with [dash #1531](https://github.com/plotly/dash/pull/1531)

One of the changes in #1531 is to sort the props.  A docstring comment that ends with a `"` in the last prop will cause the component docstring to end with 4 double quotes `""""` which causes an error in the build.  Using a single quote fixes this error.

### DIFF
--- a/src/components/Location.react.js
+++ b/src/components/Location.react.js
@@ -133,13 +133,13 @@ Location.propTypes = {
      */
     id: PropTypes.string.isRequired,
 
-    /** pathname in window.location - e.g., "/my/full/pathname" */
+    /** pathname in window.location - e.g., '/my/full/pathname' */
     pathname: PropTypes.string,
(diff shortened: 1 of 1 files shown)


alexcjohnson:
Hmm an ending `"` is going to pop up in 3rd-party components. In fact, including `"""` in the docstring isn't out of the question, if for example someone tries to include a multiline code example in a docstring. We should handle this case in the component generator, rather than requiring specific formatting for the strings. Perhaps we can replace *all* `"` in the docstrings with `\"`? Anyone looking at the source code will see the backslash but it won't be included when Python turns this into a string.

AnnMarieW:
Actually, this PR is no longer needed.   

The component generator also updates descriptions so they end with a period, so now the last character of the last prop won't ever be a double quote.

Issue #808, opened 2020-05-23
disable autocomplete in dropdown by default


chriddyp:
![image](https://user-images.githubusercontent.com/1280389/82741649-4b312580-9d09-11ea-978d-b1ffe3344c8a.png)

Here is one approach: https://github.com/JedWatson/react-select/issues/758, but not sure if those `inputProps` apply to the version that we're on.

Migggz:
I have the same issue and `inputProps` not worked for me

PCPbiscuit:
Try 
<Select isSearchable={false} />

Worked for me

chriddyp:
Fixed in https://github.com/plotly/dash-core-components/pull/923 😌

Issue #787, opened 2020-04-07
Set autoComplete off on Dropdown [dcc.Dropdown]


pfbuxton:
When a Dropdown is searchable it is very annoying to have previous text shown as the options are then hidden by the autoComplete, e.g.
![logo](https://user-images.githubusercontent.com/4923588/78696088-d9d31a00-78f6-11ea-8e23-1a4feab87fba.png)

It may be best to switch autoComplete off for Dropdown's, or allow it as an optional prop.

bfaivre:
Agreed. I know i dcc.Input has this as an option.

fheilz:
third this. @chriddyp

PR #921, opened 2021-01-20
Add autoComplete prop to Dropdown #808

AnnMarieW:
Currently, the default for   `autoComplete` is `"on"` in the Dropdown component.  However, unlike in dcc.Input,  there is no `autoComplete`  prop to control this setting. 

When  `autoComplete="on"`, the browser will autocomplete fields that have ids like "name" or "address" based on the history stored in the browser. This blocks the options defined in the dropdown.

image

This pull request addresses https://github.com/plotly/dash-core-components/issues/808, https://github.com/plotly/dash/issues/960 and https://github.com/plotly/dash-core-components/issues/787
by setting the `autoComplete` default to `"off"` and adding the `autoComplete` prop

This issue has also been mentioned numerous times on the community forum:
https://community.plotly.com/t/disable-autocomplete-for-dropdowns-input-box-in-dash/22463
https://community.plotly.com/t/disable-browser-autocomplete-on-multiselect-dropdown/37478
https://community.plotly.com/t/keep-getting-autofill-for-dropdown-not-all-browsers/30359

I updated dropdown unit test to include the autoComplete prop, but I'm not sure if other tests are needed.

Here is a small app that will duplicate the issue - at least in Chrome.  

```
import dash

app = dash.Dash(__name__)
app.layout = html.Div(
    [
        dcc.Dropdown(
            id="name",
            options=[
                {"label": "New York City", "value": "NYC"},
                {"label": "Montreal", "value": "MTL"},
                {"label": "San Francisco", "value": "SF"},
            ],
            value="NYC",
           # autoComplete="on",  # or 'off'    new prop
        ),
        html.Div(id="output-div"),
    ]
)


@app.callback(Output("output-div", "children"), Input("name", "value"))
def update(city):
    return f"You have selected {city}"


if __name__ == "__main__":
    app.run_server(debug=True)

```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -253,63 +253,37 @@ workflows:
   version: 2
   python3.7:
     jobs:
-      - lint-unit-37:
-          context:
-            - dash-docker-hub
-      - build-dash-37:
-          context:
-            - dash-docker-hub
+      - lint-unit-37
+      - build-dash-37
       - test-37:
-          context:
-            - dash-docker-hub
           requires:
             - build-dash-37
       - test-legacy-37:
-          context:
-            - dash-docker-hub
           requires:
             - build-dash-37
       - percy-finalize:
-          context:
-            - dash-docker-hub
           requires:
             - test-37
             - test-legacy-37
 
   python3.6:
     jobs:
-      - lint-unit-36:
-          context:
-            - dash-docker-hub
-      - build-dash-36:
-          context:
-            - dash-docker-hub
+      - lint-unit-36
+      - build-dash-36
       - test-36:
-          context:
-            - dash-docker-hub
           requires:
             - build-dash-36
       - test-legacy-36:
-          context:
-            - dash-docker-hub
           requires:
             - build-dash-36
 
   python2.7:
     jobs:
-      - lint-unit-27:
-          context:
-            - dash-docker-hub
-      - build-dash-27:
-          context:
-            - dash-docker-hub
+      - lint-unit-27
+      - build-dash-27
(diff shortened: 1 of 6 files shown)


chriddyp:
This looks good to me! @alexcjohnson - Any concerns or ideas of how we might test this?

alexcjohnson:
Interesting... re: tests I think unit tests are sufficient, just mount a default element and ensure the `<input>` in the DOM has `autoComplete` disabled, then mount one with it explicitly turned on and ensure that this is propagated.

But I guess I wonder whether this even needs to be an option, or if we can get away with just disabling `autoComplete` here? Is there a real use case where you want the built-in browser completion with a dropdown? Isn't it always going to get in the way of the explicit dropdown items, and if you wanted autocomplete you'd use a `dcc.Input` instead? AFAICT all the people asking about this just want to be able to disable it, does anyone like it?

AnnMarieW:
I think I saw a post where the autocomplete feature was being used with the Dropdown.  But right, they could just switch it to an dcc.Input for better results.  Do you want me to remove the prop?

alexcjohnson:
I looked through the posts you referenced more carefully and still see nobody arguing they like this behavior. So yes, let's treat this as a bug and simply lock it to `autoComplete: 'off'`. If anyone comes back and complains we can always add the prop in later, with default `'off'` as you have now.

AnnMarieW:
I think I saw it in a blog post or stackoverflow or something, but I can't find it now.    I'll take the prop out for now, and it will be easy to add back if  necessary.

PR #926, opened 2021-02-06
Download component

### LINKED ISSUE #216: Official download component

alexcjohnson:
Updated version of #863 - Thanks @emilhe!
Closes #216 

The syntax we landed on there, starting with this setup:

```py
import dash

app = dash.Dash(prevent_initial_callbacks=True)
app.layout = html.Div([html.Button("Download", id="btn"), dcc.Download(id="download")])
```

Contents as a plain string, showing the raw JSON required:
```py
@app.callback(Output("download", "data"), Input("btn", "n_clicks"))
def func(n_clicks):
    return dict(content="Hello world!", filename="hello.txt")
```

And the utility functions are now on the main `dcc` namespace.

To send a file from disk:
```py
@app.callback(Output("download", "data"), Input("btn", "n_clicks"))
def func(n_clicks):
    return dcc.send_file("./tests/integration/download/download-assets/Lenna.jpeg")
```

And data frames, supporting many of the pandas export  methods:
```py
import pandas as pd
df = pd.DataFrame({"a": [1, 2, 3, 4], "b": [2, 1, 5, 6], "c": ["x", "x", "y", "y"]})

@app.callback(Output("download", "data"), Input("btn", "n_clicks"))
def func(n_clicks):
    return dcc.send_data_frame(df.to_csv, "mydf.csv")
```


emilhe (approved):

emilhe:
Looks good! I think it makes perfect sense to include the utility functions in the `dcc` namespace, it yields a more coherent experience working with the download component.

alexcjohnson:
Still some investigation needed into why the Py2.7 and Py3.7 tests are failing repeatably even though Py3.6 works fine (after downgrading to pyarrow v2, the brand new v3 wouldn't install on Py3.6 on CI). Once we sort that out we can merge this.

PR #928, opened 2021-02-11
Add n_clicks to annotationClickData so that callback is fired when clicking on an annotation multiple times

chriddyp:
Currently `dash`' verifies that the old props != the new props before firing a callback.

Line: https://github.com/plotly/dash/blame/dev/dash-renderer/src/TreeContainer.js#L126
Commit: https://github.com/plotly/dash/commit/05908b2b9c41e3bb7ec5e5505859c7fb7cb78734

This behavior prevents multiple clicks on the same annotation from firing the callback. 

Dash's comparison may not be the right behavior, but this PR will at least get around the issue.

### DIFF
--- a/src/components/Graph.react.js
+++ b/src/components/Graph.react.js
@@ -180,6 +180,8 @@ PlotlyGraph.propTypes = {
 
     /**
      * Data from latest click annotation event. Read-only.
+     * In order for an annotation to be "clickable", set
+     *`captureevents=True` within the annotation object.
      */
     clickAnnotationData: PropTypes.object,
 
--- a/src/fragments/Graph.react.js
+++ b/src/fragments/Graph.react.js
@@ -332,6 +332,14 @@ class PlotlyGraph extends Component {
                 ['event', 'fullAnnotation'],
                 eventData
             );
+            if (this.props.clickAnnotationData) {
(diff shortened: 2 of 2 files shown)


alexcjohnson:
@chriddyp while we investigate a deeper fix, were you able to test out the proposed workaround using circular callbacks https://github.com/plotly/dash/pull/1525 to reset `annotationClickData` when it's consumed?

chriddyp:
This workaround does indeed work 👍🏼 
```python
import dash
import json
app = dash.Dash(__name__)
app.layout = html.Div([
    dcc.Graph(id='graph', figure={
        'data': [{'y': [2, 1, 5]}],
        'layout': {
            'annotations': [{
                'text': 'click me',
                'x': 0.5, 'y': 0.5, 'xref': 'paper', 'yref': 'paper',
                'captureevents': True
            }]
        }
    }),
    html.Pre(id='output')
])
@app.callback(
    Output('output', 'children'), 
    Output('graph', 'clickAnnotationData'), 
    Input('graph', 'clickAnnotationData'))
def update_output(c):
    print(c)
    return [json.dumps(c, indent=4), None]
app.run_server(debug=True)
```

click

PR #934, opened 2021-03-12
Fix typo in EnhancedTab disabled_className property

ivanna-ostrovets:
I was trying to add custom `className` for the disabled tab through `disabled_className` and faced an issue that the component received `undefined` instead of the provided string. Turned out that the property name was misspelled. So I fixed this by renaming property from `disabled_classname` to `disabled_className`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - [#930](https://github.com/plotly/dash-core-components/pull/930) Fixed a bug [#867](https://github.com/plotly/dash-core-components/issues/867) with `DatePickerRange` that would sometimes shift the allowed dates by one day.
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Great catch, thanks @ivanna-ostrovets! Would you mind doing two more things to finish this up? (1) add a changelog entry under `UNRELEASED` -> `Fixed`, (2) include `disabled_className` in a test and verify that it shows up in the DOM - Maybe just add a disabled tab here:

https://github.com/plotly/dash-core-components/blob/8f241853d71b1a0c3b840a80303aedacb37c5a2e/tests/integration/tab/test_tabs_with_graphs.py#L26

and then down below:

https://github.com/plotly/dash-core-components/blob/8f241853d71b1a0c3b840a80303aedacb37c5a2e/tests/integration/tab/test_tabs_with_graphs.py#L65

add another line like
```py
dash_dcc.wait_for_element("#tab-3.my-disabled-class")
```

alexcjohnson (approved):
Perfect, thank you! 💃

PR #938, opened 2021-03-24
Added unit for Upload's max_size documentation.

marekpiotrowski:
We're using ```dcc.Upload``` component and I just had to do a runtime check to verify what unit prefix are we talking about here, so I thought I'll just create a quick PR for that.

Cheers,
Marek


alexcjohnson (approved):
Thanks! 💃

PR #941, opened 2021-03-25
Pass a bbox through graph hover data

rreusser:
This PR very minimally passes `bbox`, if present, through Graph component hover data toward custom hover with the ToolTip component in #940.


alexcjohnson:
@rreusser this looks perfect, but maybe you want to merge this into #940 so the whole new feature can be demoed together, especially in the docs https://github.com/plotly/dash-docs/pull/1104?

PR #948, opened 2021-04-08
Add disabled_days prop to datepickers

tcbegley:
Hi,

I've seen a few people requesting on the community forum and elsewhere (e.g. [here](https://community.plotly.com/t/date-picker-single-make-certain-dates-unavailable/51787)) that they have the option to make certain days unselectable in the day picker components beyond specifying a range with `{min,max}_date_allowed`.

As it happens I previously implemented this for a colleague who had the same problem, so thought I would submit it here in case it's of interest. I'm not deeply invested in any aspect of this so if you don't like the interface / implementation / idea that's totally fine 😅 . 

I think the main problem with this implementation is that if you wanted to for example disable all weekends over the span of several years you end up passing a pretty long list which could start to affect performance. I suspect for most applications it would be pretty hard to hit this limit but I haven't really investigated this much.

Here's a simple example you can use to test it. It makes only weekdays in April selectable.

```python
from datetime import datetime, timedelta

import dash

APRIL = [datetime(2021, 4, 1) + timedelta(days=i) for i in range(30)]

app = dash.Dash()

app.layout = html.Div(
    [
        dcc.DatePickerSingle(
            disabled_days=[date for date in APRIL if date.weekday() >= 5],
            min_date_allowed=APRIL[0],
            max_date_allowed=APRIL[-1],
        )
    ],
    style={"padding": "30px"},
)

if __name__ == "__main__":
    app.run_server(debug=True)
```

which looks something like this

image

### DIFF
--- a/src/utils/convertToMoment.js
+++ b/src/utils/convertToMoment.js
@@ -19,6 +19,10 @@ export default (newProps, momentProps) => {
                         undefined
                 );
             }
+        } else if (Array.isArray(value)) {
+            dest[key] = value.map(d =>
+                d === null || d === undefined ? null : moment(d)
+            );
         } else {
             dest[key] = moment(value);
 
--- a/src/utils/convertToMoment.js
+++ b/src/utils/convertToMoment.js
@@ -20,7 +20,7 @@ export default (newProps, momentProps) => {
                 );
             }
         } else if (Array.isArray(value)) {
-            dest[key] = value.map(moment);
+            dest[key] = value.map(d => moment(d));
         } else {
             dest[key] = moment(value);
 
--- a/src/utils/convertToMoment.js
+++ b/src/utils/convertToMoment.js
@@ -20,9 +20,7 @@ export default (newProps, momentProps) => {
                 );
             }
         } else if (Array.isArray(value)) {
-            dest[key] = value.map(d =>
-                d === null || d === undefined ? null : moment(d)
-            );
+            dest[key] = value.map(moment);
         } else {
             dest[key] = moment(value);
 
--- a/tests/integration/calendar/test_date_picker_range.py
+++ b/tests/integration/calendar/test_date_picker_range.py
@@ -109,3 +109,39 @@ def test_dtpr004_max_and_min_dates_are_clickable(dash_dcc):
         '#dps-initial-month .DateInput_input.DateInput_input_1[placeholder="End Date"]',
         "01/20/2021",
     )
+
+
+def test_dtpr005_disabled_days_arent_clickable(dash_dcc):
+    app = dash.Dash(__name__)
+    app.layout = html.Div(
+        [
+            html.Label("Operating Date"),
+            dcc.DatePickerRange(
+                id="dpr",
+                min_date_allowed=datetime(2021, 1, 1),
+                max_date_allowed=datetime(2021, 1, 31),
+                initial_visible_month=datetime(2021, 1, 1),
+                disabled_days=[datetime(2021, 1, 10), datetime(2021, 1, 11)],
+            ),
+        ],
+        style={
+            "width": "10%",
+            "display": "inline-block",
+            "marginLeft": 10,
+            "marginRight": 10,
+            "marginBottom": 10,
+        },
+    )
+    dash_dcc.start_server(app)
+    date = dash_dcc.find_element("#dpr input")
+    assert not date.get_attribute("value")
+    assert not any(
+        dash_dcc.select_date_range("dpr", day_range=(10, 11))
+    ), "Disabled days should not be clickable"
+    assert all(
+        dash_dcc.select_date_range("dpr", day_range=(1, 2))
+    ), "Other days should be clickable"
+
+    # open datepicker to take snapshot
+    date.click()
+    dash_dcc.percy_snapshot("dtpr005 - disabled days")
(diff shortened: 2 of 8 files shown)


alexcjohnson:
@tcbegley this looks like a really nice feature! Before looking at the code, the only thing I'd like to see is a plan for how we would extend this to support patterns of dates to disable, so that you don't have to enumerate them all:
- January 1 every year
- The first of every month
- Every Tuesday

We don't need to implement such patterns in this PR, I just want to make sure we have a consensus on how we would do that in the future, to ensure it wouldn't be a breaking change to the feature as implemented here. Do they fit into the same array with different values? A separate prop?

tcbegley:
>  I just want to make sure we have a consensus on how we would do that in the future

That sounds like a good idea!

Here's some initial thoughts, but note that I haven't really thought about implementation yet. Since the prop accepts date strings of the form `YYYY-MM-DD`, would allowing wildcards potentially be one way forward here? It could address your first two examples by allowing the user to pass something like this:
- `disabled_days=["****-01-01"]`
- `disabled_days=["****-**-01"]`

Days of the week seems harder and would maybe need a different prop. Perhaps it's as simple as `disabled_weekdays=[5, 6]` or `disabled_weekdays=["saturday", "sunday"]`? The former is consistent with `datetime.date.weekday`, the latter is a lot more readable but I guess would require some extra parsing within the component.

Anyway, certainly interested to hear alternative suggestions.

alexcjohnson:
OK, having looked at the implementation (which is great, nice and clean! 🎉 ), I think it'll be fine to include arbitrary patterns within the same `disabled_days` array - whether they're date strings with wildcards, the name of a weekday (I'd rather that than weekday numbers, as that's less readable but also there are differing numbering conventions  - that said names are language-dependent so nothing's perfect 🤷 )

So I think all we really need is a test or two - perhaps one test for each component that a disabled date can't be chosen, plus a percy snapshot showing the open datepicker so we can see that the correct dates are visibly disabled.

src/utils/convertToMoment.js
```diff
@@ -19,6 +19,10 @@ export default (newProps, momentProps) => {
                         undefined
                 );
             }
+        } else if (Array.isArray(value)) {
+            dest[key] = value.map(d =>
+                d === null || d === undefined ? null : moment(d)
```
alexcjohnson, line 24: Might not even need the null case here since we're in an array with propType string (the single values for max and min need that test because they're optional props)

But if we do keep it, you can use https://ramdajs.com/docs/#isNil

alexcjohnson:
> also there are differing numbering conventions

Actually we already build in Sunday=0 via `first_day_of_week` - so I stand corrected, we should keep using that (if and when we get to this extension 😁 )

tcbegley:
Great! I'll add some tests in the next few days.

src/utils/convertToMoment.js
tcbegley: Thanks, I opted for ditching the check since it's nice and clean: `value.map(moment)`

tcbegley:
Added some tests. From what I can tell by looking at the logs they are failing because CircleCI is installing dash-core-components 1.16.0, but when I build from this branch it builds version 1.15.0. Not sure the best way to fix that?

tests/integration/calendar/test_date_picker_range.py
```diff
@@ -109,3 +109,39 @@ def test_dtpr004_max_and_min_dates_are_clickable(dash_dcc):
...
+    assert not any(
+        dash_dcc.select_date_range("dpr", day_range=(10, 11))
+    ), "Disabled days should not be clickable"
+    assert all(
+        dash_dcc.select_date_range("dpr", day_range=(1, 2))
+    ), "Other days should be clickable"
+
+    # open datepicker to take snapshot
+    date.click()
+    dash_dcc.percy_snapshot("dtpr005 - disabled days")
```
tcbegley, line 147: @alexcjohnson please check this definitely takes a snapshot with the datepicker open. I think it does, but wasn't quite sure how to check locally?

tests/integration/calendar/test_date_picker_single.py
```diff
@@ -140,3 +140,37 @@ def test_dtps012_initial_month(dash_dcc):
...
+    date = dash_dcc.find_element("#dps input")
+    assert not date.get_attribute("value")
+    assert not dash_dcc.select_date_single(
+        "dps", day=10
+    ), "Disabled days should not be clickable"
+    assert dash_dcc.select_date_single("dps", day=1), "Other days should be clickable"
+
+    # open datepicker to take snapshot
+    date.click()
+    dash_dcc.percy_snapshot("dtps013 - disabled days")
```
tcbegley, line 176: ditto

alexcjohnson:
> Added some tests. From what I can tell by looking at the logs they are failing because CircleCI is installing dash-core-components 1.16.0, but when I build from this branch it builds version 1.15.0. Not sure the best way to fix that?

Right, this is a quirk of our CI setup - we've released a new dash version since you created this PR, if you merge in the latest `dev` this problem will go away.

src/utils/convertToMoment.js
tcbegley: Turns out that actually causes some problems, but `value.map(d => moment(d))` is ok.

tests/integration/calendar/test_date_picker_range.py
tcbegley: All good, once I fixed the other tests and Percy ran I was able to see them myself.

tcbegley:
This is ready for review now, Percy is failing because I added new snapshots, but all seems to be working.

alexcjohnson (approved):
Excellent - nice tests! 💃

Issue #857, opened 2020-09-02
Clientside countdown component to compliment `dcc.Interval`


chriddyp:
nytimes has a nice countdown badge that tells the user when the data will update:

![Screen Shot 2020-09-01 at 5 59 22 PM](https://user-images.githubusercontent.com/1280389/91919935-6999ac00-ec7c-11ea-9d5b-aea7d7f714f2.png)

This might be a nice built-in component that could be tied to a `dcc.Interval` component. It would be documented with `dcc.Interval` and its use would be encouraged for good UX.

API ideas:
- A `dcc.Countdown` component that could "target" a `dcc.Interval` component: `dcc.Countdown(target='interval-id', children='Checking for updates in {}')
- Or, built in to the `dcc.Interval` component itself: `dcc.Interval(children='Checking for updates in {}')

alexcjohnson:
Or give `dcc.Interval` a prop like `display` to choose between default no display, text in various format (seconds, HH:MM:SS etc), or a graphical element like a ring or a progress bar. Then your use case would look like:
`html.Span(['Checking for updates in ', dcc.Interval(display='M:SS')])`

bandersen23:
@chriddyp I wrote a quick example that does this using a clientside callback. It'd be nice if the countdown was global across all users, but that would be better implemented in a different way I would assume. What do you think?

```python
import dash
from dash.dependencies import Output, Input, State
import dash_core_components as dcc
import dash_html_components as html
import datetime

app = dash.Dash()
app.layout = html.Div([
    dcc.Store(id='last-update-time'),
    dcc.Interval(id='counter', interval=20000),
    dcc.Interval(id='update-in-timer', interval=500),
    html.Span(id='status', children=None, style={'background-color': 'red', 'color':'white', 'width': '200px', 'height': '100px'})
])

@app.callback(
    Output('last-update-time', 'data'), 
    Input('main-counter', 'n_intervals')
)
def update_store(n):
    return datetime.datetime.now().timestamp()

js_counter = """
function counting_down(n, last, interval) {

    t =  Math.trunc((interval - (Date.now() - last*1000))/1000);
    
    return `Checking for updates in: ${t} seconds`
}
"""

app.clientside_callback(
    js_counter,
    Output('status', 'children'),
    Input('update-in-timer', 'n_intervals'),
    State('last-update-time', 'data'),
    State('main-counter', 'interval')
)

app.run_server()
```

PR #961, opened 2021-05-03
Add Timer component

AnnMarieW:
## dcc.Timer()

The new Timer component is a response to the discussion in [#857](https://github.com/plotly/dash-core-components/issues/857).  
  
`dcc.Timer` is based on the `dcc.Interval` component.  It has all the functionality of `dcc.Interval` plus these new features:

 - Operate the timer in either `countdown` or `stopwatch` (count up) modes.
 - Specify custom messages to display at certain times.
 - Automatically convert milliseconds into human readable times.  For example, 1337000000ms can be display as:
  '15d 11h 23m 20s'  See other available formats in the `timer_format` prop.
 - Specify certain times to trigger a callback.  This makes it easy to start or stop jobs at a specified elapse time.
 - Improve load and performance times because updates can happen clientside.  This makes it unnecessary to fire a callback
   frequently (ie every interval) just to update a countdown/stopwatch message.

 

#### Component Properties

|Prop name|Type & Default value|Description|Example values|
|----|----|----|----|
| id| string; optional|id of component used to identify dash components in callbacks| |
|interval| number; default 1000| This component will increment the counter `n_intervals` every `interval` milliseconds| |
|disabled |boolean; optional| If True, the n_interval counter  and the timer no longer updates.  This pauses the timer.| |
|n_intervals| number; default 0| Number of times the interval has passed (read-only)| |
|max_intervals| number; default -1| Number of times the interval will be fired. If -1, then the interval has no limit and if 0 then the interval stops running.||
|timer| number; default 0| When in countdown mode, the timer will count down to zero from the starting `duration` and will  show the number of milliseconds remaining.  When in stopwatch mode, the timer will count up from zero and show the number of milliseconds elapsed. (read only) | |
|mode| 'stopwatch' or 'countdown'; default 'countdown'| The timer will count down to zero in `countdown` mode and count up from zero in `stopwatch` mode| |
|duration| number; default -1|  Sets the number of milliseconds the timer will run.  If -1 the timer will not be limited by the duration and if 0 then the timer stops running and may be reset.||
|reset| boolean; default True| This will start the timer at the beginning with the given prop settings.| |
|fire_times| list; optional| A list of the time(s) in milliseconds at which to fire a callback. This can be used to start a task at a given time rather than using the timer.  Since the timer is typically set at a small interval like one second, using `fire_times` can reduce the number of times a callback is fired and can increase app performance.  The time(s) must be a multiple of the interval.| |
|at_fire_time| number| This is updated when the timer reaches one of the times in the `fire_times` list. Using `at_fire_time` in a  callback will trigger the callback at the time(s) in `fire_times` (Read only)| |
|rerun|boolean; default False| When True, the  timer repeats once the timer has run for the number of milliseconds set in the `duration`.| |
|messages|dict; optional| Timer messages to be displayed by the component rather than showing the timer. It is a dictionary in the form of: { integer: string} where integer is the time in milliseconds of when the `string` message is to be displayed.  Note:  `timer_format` will override `messages`.| {10000 : "updating in 10 seconds"} will display the message "updating in 10 seconds" once the timer equals 10000.|
|timer_format|string; optional.  One of:| If a timer is displayed, it will override timer `messages`.  This formats the timer (milliseconds) into human readable formats.| |
| | 'none'; default| No timer will be displayed.  Timer `messages` will be shown.
| | 'display'| Display timer in default format.  For example, 1337000000 milliseconds will display as: '15d 11h 23m 20s'|'15d 11h 23m 20s'|
| | 'compact'| Shows a compact timer display. It will only show the first unit| 1h 10m → 1h|
| | 'verbose'| Verbose will display full-length units.|5h 1m 45s → 5 hours 1 minute 45 seconds|
| | 'colonNotation'|  Use this when you want to display time without the time units, similar to a digital watch. It will always show at least minutes: 1s → 0:01|5h 1m 45s → 5:01:45|

### Demo

Run `timer_demo_py` to see eight examples using `dcc.Timer`. Some  may be good candidates for the docs.   Here is a preview:

#### Countdown and Stopwatch timers with messages updated clientside - no callbacks required!
Peek 2021-05-03 06-31

#### Space shuttle app:
This app uses the `dcc.Timer` component to launch the space shuttle.  :rocket:
- The `messages` property is used to define the messages that are displayed at certain times during the launch.  
- The `fire_times` property specifies the time to trigger a callback to start the launch.  

The timer runs for 51 seconds, but it only fires the callback once (at liftoff) All the other messages are handled 
clientside by the component.

Peek 2021-05-03 06-39


src/components/Timer.react.js
```diff
@@ -0,0 +1,322 @@
...
+
+    handleMessages(props, new_timer) {
+        const {messages, timer_format} = this.props;
+
+        const messagesObj = Object.assign({}, messages);
+        if (new_timer in messagesObj) {
+            this.renderMessage = `${messagesObj[new_timer]}`;
+        }
+
+        const prettyMilliseconds = require('pretty-ms');
```
alexcjohnson: Put all imports at the top - and you should be able to convert this to the more modern `import prettyMilliseconds from 'pretty-ms';`

src/components/Timer.react.js
```diff
@@ -0,0 +1,322 @@
...
+        if (timer_format !== 'none') {
+            const formatObj = getFormat(timer_format);
+            this.renderMessage = `${prettyMilliseconds(new_timer, formatObj)}`;
```
alexcjohnson:
```suggestion
            this.renderMessage = prettyMilliseconds(new_timer, formatObj);
```

src/components/Timer.react.js
```diff
@@ -0,0 +1,322 @@
...
+     */
+    max_intervals: PropTypes.number,
+
+    /**
+     * When in countdown mode, the timer will count down to zero from the starting `duration` and will show the number
+     *  of milliseconds remaining.
+     *  When in stopwatch mode, the timer will count up from zero and show the number of milliseconds elapsed.
+     *  (read-only)
+     */
+    timer: PropTypes.number,
```
alexcjohnson: Maybe just call this "time"?

src/components/Timer.react.js
```diff
@@ -0,0 +1,322 @@
...
+
+        const new_n_intervals = n_intervals + 1;
+        const updateProps = {n_intervals: new_n_intervals};
+
+        this.countdown = duration - interval * new_n_intervals;
+        let new_timer;
+        if (mode === 'countdown' && duration !== -1) {
+            new_timer = this.countdown;
+        } else {
+            // stopwatch.  Note - stopwatch mode only if duration === -1
```
alexcjohnson: This logic seems confusing to me - also fragile, there's nothing stopping a user from setting some other negative duration. I think it would be cleaner if `countdown` mode treated any negative duration the same as 0, ie it's still in countdown mode but never starts.

src/components/Timer.react.js
```diff
@@ -0,0 +1,322 @@
...
+    id: PropTypes.string,
+
+    /**
+     * This component will increment the counter `n_intervals` every
+     * `interval` milliseconds
+     */
+    interval: PropTypes.number,
+
+    /**
+     * If True, the n_interval counter  and the timer no longer updates.  This pauses the timer.
```
alexcjohnson:
```suggestion
     * If True, the n_interval counter and the timer no longer updates. This pauses the timer.
```

src/components/Timer.react.js
```diff
@@ -0,0 +1,322 @@
...
+    /**
+     * This will start the timer at the beginning with the given prop settings.
+     */
+    reset: PropTypes.bool,
+
+    /**
+     * A list of the time(s) in milliseconds at which to fire a callback. This can be used to start a task at a given
+     * time rather than using the timer. Since the timer is typically set at a small interval like one second, using
+     * fire_times can reduce the number of times a callback is fired and can increase app performance. The time(s) must be a
+     * multiple of the interval.
```
alexcjohnson:
```suggestion
     * fire_times can reduce the number of times a callback is fired and can increase app performance. Each time must be a
     * multiple of the interval.
```
I wonder though if we can't relax that constraint - and update `at_fire_time` as soon as we've reached or passed the specified point. I can see this being useful for example if you're fiddling with `interval` to see how it affects the displayed values, it could be frustrating to need to adjust `fire_times` every time you do that - especially since a matching adjustment may need to happen inside your callback as well.

src/components/Timer.react.js
```diff
@@ -0,0 +1,322 @@
...
+     */
+    rerun: PropTypes.bool,
+
+    /**
+     * Timer messages to be displayed by the component rather than showing the timer. It is a dictionary in the form of:
+     * {integer: string} where integer is the time in milliseconds of when the string message is to be displayed.
+     * Note: timer_format will override messages. For example, {10000 : "updating in 10 seconds"} will display the message
+     * "updating in 10 seconds" once the timer equals 10000.
+     */
+    messages: PropTypes.object,
```
alexcjohnson: This can be `PropTypes.objectOf(PropTypes.string)` as in [Slider](https://github.com/plotly/dash-core-components/blob/50cea817f7da59809b5f2e41f2d0dadbbc5e54e7/src/components/Slider.react.js#L36)

Do these times also need to be multiples of `interval`? And if so, can we relax that too, so the message is displayed when we reach or pass the specified time?

src/components/Timer.react.js
```diff
@@ -0,0 +1,322 @@
...
+     *  `'display'`:  example - 1337000000 milliseconds will display as: '15d 11h 23m 20s';
+     *  `'compact'`: will show only the first unit: 1h 10m → 1h ;
+     *  `'verbose'`: will show full-length units. Example --  5 hours 1 minute 45 seconds
+     *  `'colonNotation'`: Useful when you want to show time without the time units, similar to
+     *                   a digital watch. Will always shows time in at least minutes: 1s → 0:01.
+     *                   Example - 5h 1m 45s → 5:01:45.
+     */
+    timer_format: PropTypes.oneOf([
+        'none',
+        'display',
```
alexcjohnson: in `getFormat` it looks like this is `'default'`?

src/components/Timer.react.js
```diff
@@ -0,0 +1,322 @@
...
+        'compact',
+        'verbose',
+        'colonNotation',
```
alexcjohnson: I know this is just a passthrough to `pretty-ms` but it's not important for us to follow their convention, I'm tempted to simplify this to a single word like the rest of them, perhaps `'colons'`?

src/components/Timer.react.js
AnnMarieW: OK, I fixed this

src/components/Timer.react.js
AnnMarieW: and  fixed this too.  Thanks

src/components/Timer.react.js
AnnMarieW: Good idea

src/components/Timer.react.js
AnnMarieW: Right, that would work better.  I was trying to make it consistent with the `max_intervals`, which when set to -1 has no limit.  But it doesn't make as much sense for the `duration`.  Maybe if the `duration` is `None`, then the timer could work the same way as  `dcc.Interval`  (ie a count up timer with no limit on the duration)

src/components/Timer.react.js
AnnMarieW: I can see the advantages of doing it that way.  The main problem is with the countdown timer, which you'd expect to stop at zero.  How do you make it stop at zero if the `duration` is not a multiple of the `interval`?

Or, if you don't make it stop at zero, do you stop it short or run the timer to a negative number?   Either would look like a bug.

src/components/Timer.react.js
AnnMarieW: For now it needs to be a multiples of the `interval` else the timer will never be equal to the time in the `messages` dict.  Unlike with `dcc.Interval` it's possible to have a lot of functionality clientside with this component so there is no   performance issue with making the `interval` a small number.  Then it's easy to make the times in the `message`, `fire_times` and the `duration` a multiple of the `interval`.

If we leave it as-is, should there be an warning message if the `duration` or `fire_times` or `message` times are not multiples of the `interval`?

src/components/Timer.react.js
AnnMarieW: # :facepalm: 
How do you find stuff like this? (fixed now - thanks)

src/components/Timer.react.js
AnnMarieW: Right - that's better and more consistent with the python naming conventions.

AnnMarieW:
I updated the component based on  your helpful feedback -Thanks!

I made changes to the timer logic to (hopefully) make it  less fragile and easier to understand.  I'm interested to see if you think it's better.

The only other item,  other than testing,  is regarding relaxing the constraint that the `interval` must be a multiple of `duration`.  Those questions are inline above.

src/components/Timer.react.js
```diff
@@ -0,0 +1,317 @@
...
+        // it hasn't started yet (& it should start)
+        this.intervalId = window.setInterval(this.reportInterval, interval);
+    } // end handle timer
+
+    handleMessages(props, new_time) {
+        const {messages, timer_format} = this.props;
+
+        const messagesObj = Object.assign({}, messages);
+        if (new_time in messagesObj) {
+            this.renderMessage = `${messagesObj[new_time]}`;
```
alexcjohnson:
```suggestion
            this.renderMessage = messagesObj[new_time];
```
Same simplification as below, I missed this one the first time around. Backtick template literals are great to avoid concatenations but they're just passthroughs here.

src/components/Timer.react.js
alexcjohnson: Right, it's a good deal more complicated if we allow `fire_times` and `duration` to not be multiples of `interval` - speaking of which we should document what happens with `duration` in this case. Looks to me like right now we just go past zero and then stop? That's fine for now as long as we document it, but later we may want to do better: either start or end with a one-off `setTimeout` of the remainder time. For `fire_times` I also think eventually we should allow them to occur between intervals, and in fact this could also be handled by a `setTimeout` created during the last full interval before the fire time. But again, not necessary for now, we can add that later.

src/components/Timer.react.js
alexcjohnson: OK, let's keep the requirement that `messages` keys be multiples of `interval` for now as well.

We could make a console warning when any of these is not a multiple of interval. That's not very prominent, in fact we normally try to avoid giving anyone a reason to look in the console. I'm tempted to say we should actually throw an error. Seems a bit heavy-handed, but things will start to fail in hard-to-debug ways when that happens so maybe it's the best option. For example: if `duration` is not a multiple, the timer will still run for about the right amount of time but `fire_times` and `messages`, even if they ARE multiples, will not respond correctly because the actual `time` values will be `duration - n_intervals * interval` so also not multiples of `interval`.

AnnMarieW:
Moved to [dash PR#2352](https://github.com/plotly/dash/pull/2352)

PR #963, opened 2021-05-06
fixes tooltips on RangeSlider

### LINKED ISSUE #763: Slider tooltip not updating in certain scenarios
as reported in https://community.plot.ly/t/tooltip-and-slider-not-refreshing-the-position/35556

### LINKED ISSUE #885: Slider tooltip remains in fixed position on scroll
Tooltips remain in a fixed position on scrollable pages where the outer style is `{"height": "100vh", "overflow": "auto"}`. Behaviour occurs regardless of the browser being used.

It's possible other style settings cause this behaviour as well, but this is the style setting which caused the error for me. In this example, removing the styling does not change the look of the app, but it is possible that a `height`/`overflow` setting is important for the look of an app and the slider is nested inside of other components.

### Minimal example
```python
import dash
import dash_bootstrap_components as dbc

app = dash.Dash(__name__, external_stylesheets=[dbc.themes.BOOTSTRAP])
server = app.server

app.layout = html.Div(
    children=[
        dbc.Card("This is a card", style={"padding": "100px"}),
        dcc.RangeSlider(
            min=0,
            max=1,
            value=[0, 1],
            step=0.01,
            tooltip=dict(always_visible=True, placement="bottom"),
        ),
        # Note that these cards exist only to make the page scrollable
        dbc.Col(
            [
                dbc.Row(
                    [dbc.Card("This is another card", style={"padding": "100px"})] * 3,
                )
            ]
            * 10
        ),
    ],
    # These styles seem to cause the problem, when these styles are removed the error does not occur
    style={
        "height": "100vh",
        "overflow": "auto",
    },
)


if __name__ == "__main__":
    app.run_server(debug=True)
```

### Requirements
```
dash==1.17.0
dash-bootstrap-components==0.10.7
dash-core-components==1.13.0
dash-html-components==1.1.1
```

### Screen recording
Screen-Recording-2020-11-26-at-1

### LINKED ISSUE #919: dcc.RangeSlider tooltip broken between 0 and 10
The tooltip in the `dcc.RangeSlider` breaks between 0 and 10, but appears to work properly in the `dcc.Slider`. 

This issue appeared in the past for the `dcc.Slider` as a result of a bug with the library that `dcc.Slider` is wrapped around: https://github.com/plotly/dash-core-components/issues/763

If this is the case, and it appears to be fixed, than perhaps the `dcc.RangeSlider` library needs to be updated as well?

The top slider in the `.gif` below is a `dcc.RangeSlider` and the bottom is a `dcc.Slider`.

slider-issue

Sample code:

```
from dash import Dash

# Build App
app = Dash(__name__)
app.layout = html.Div([
    html.Div(
        [
            dcc.RangeSlider(
                min=-50,
                max=50,
                step=0.5,
                value=[5, 15],
                # allowCross=False,
                tooltip={"always_visible": True}
            ),
        ]
    ),
    html.Div(
        [
            dcc.Slider(
                min=0,
                max=20,
                step=0.5,
                value=5,
                tooltip={"always_visible": True}
            )
        ], style={"marginTop": "25%"})
]
)

app.run_server(port=8060, debug=True)
```

Dependencies:
```
dash==1.19.0
dash-core-components==1.15.0
dash-html-components==1.1.2
dash-renderer==1.9.0
dash-table==4.11.2
```

CC: @kully

AnnMarieW:
Closes #885
Closes  #919  
Closes #763

This applies the fix from #878 to the RangeSlider.  It not only fixes the bug where the tooltips were visible when slider was not, but it also reduces the lag in the tooltips when the slider handles are moved.


alexcjohnson (approved):
Thanks @AnnMarieW - this looks great! 💃 

Let's see if the tests pass with the merge of the latest `dev`, ie was CircleCI just having a bad day or is there really something wrong.

tests/integration/sliders/test_sliders.py
```diff
@@ -162,6 +162,71 @@ def test_slsl005_slider_tooltip(dash_dcc):
...
+                ],
+                style=dict(maxHeight=300, overflowX="scroll"),
+            )
+        ]
+    )
+
+    dash_dcc.start_server(app)
+    dash_dcc.wait_for_element("#test-slider")
+    dash_dcc.percy_snapshot(
+        "slider-make sure tooltips are only visible if parent slider is visible"
```
alexcjohnson, line 226: Oh right - this line is failing because percy snapshot names need to be unique, this matches https://github.com/plotly/dash-core-components/pull/878

Normally it's nice to keep these snapshot names shorter too, though of course I should have provided that feedback in https://github.com/plotly/dash-core-components/pull/878 😉

PR #970, opened 2021-05-11
Accessibility and w3 compliance

coleweusthem:
Included some fixes for accessibility and w3 standard compliance.


alexcjohnson:
Hi @coleweusthem - thanks for the PR, from the commit descriptions it seems like some useful additions! It's based on the `master` branch though, which means it contains all the release build artifacts, which we don't want to commit to `dev`. Would you be able to pull out just the changes to `src/` and whatever of the `package.json`, webpack config, etc changes you originated and commit these to the tip of the `dev` branch? Probably best to make that a new PR, so it doesn't show all this history from `master`.

PR #975, opened 2021-06-01
Change Clipboard prop name from `text` to `content`

AnnMarieW:
This PR changes the prop name from `text` to `content` based on the feedback on the new dash-docs for dcc.Clipboard [here](https://github.com/plotly/dash-docs/pull/1127#discussion_r617720542)


alexcjohnson (approved):
💃  Excellent :)

PR #987, opened 2021-09-03
add express module to stub

alexcjohnson:
in case someone does eg:
`from dash_core_components.express import send_bytes`

### DIFF
--- a/dash_core_components/express.py
+++ b/dash_core_components/express.py
@@ -1,0 +1,1 @@
+from dash.dcc.express import *  # noqa: F401, F403, E402


jonmmease (approved):
I didn't know there was a `dcc.express` :grinning:  But looks good!  The top-level module warning should take care of this import as well.

:woman_dancing:

alexcjohnson:
> I didn't know there was a `dcc.express`

Right, it's not documented but it should be!

> The top-level module warning should take care of this import as well.

Yes, I've tested this.

PR #991, opened 2021-10-05
Empty Julia Package

waralex:
Empty Julia Package to avoid a name conflict in the Dash version, which already includes all core components


src/DashCoreComponents.jl
```diff
@@ -0,0 +1,5 @@
+module DashCoreComponents
+    function __init__()
+        @warn "`using DashCoreComponents` is deprecated, all DashCoreComponents functions have been moved to Dash package. Use `using Dash` instead"
```
alexcjohnson:
```suggestion
        @warn "The `DashCoreComponents` package is deprecated. All DashCoreComponents functions have been moved to the Dash package. Please switch to `using Dash` instead."
```

alexcjohnson (approved):
💃

